PR #19, opened 2018-01-24
Get rid of CGI forever

Difegue:
When LANraragi was initially written in 2014, I based it on what I knew of Perl at the time - CGI scripting with an Apache2 backend. 

Almost 4 years of excruciatingly slow development later, times have changed, [CGI has been removed from the Perl core](https://perl5.git.perl.org/perl.git/commitdiff/e9fa5a80), containing webapps in Docker containers is way easier than bundling them all under a singular Apache server, and there are [tons of alternatives.](https://metacpan.org/pod/CGI::Alternatives)

This PR merges the Mojolicious port of LRR I've been developing the past month, as **v.0.4.5**.

Everything is in working state, except for the current tagging system which I didn't bother porting, as I'd just rip it out later to insert the new plugin system.

Quick changelog:

- CGI's gone, and with it so is the Apache dependency - LRR now runs under the Mojolicious-provided Web Server. 
- The Redis configuration is now in **lrr.conf**, instead of tucked away in a source file. 
- NPM integration has progressed, now featuring a full-blown installer script. The script handles both Perl and JavaScript dependencies, making manual Linux installs a breeze. Upgrading and remote deployment aren't quite there yet, but that'll come.
- Dropped Vagrant support: It's been untested for years, and I don't think anyone ever used it. Docker's way more convenient nowadays.
- Speaking about the whale, the Dockerfile has been rewritten, with specific support towards storing everything in the user-provided content directory: Archives, Database, and Thumbnails. This makes upgrading LRR essentially painless, as all the user needs is to switch containers while keeping his content directory.
- The temporary folder (where unzipped archives are stored) is no longer automatically emptied when visiting the main page. It can be manually emptied from the configuration page, where you can see its current size. Support for maximum folder size will come later.
- Updated FontAwesome to the brand new [v5](https://fontawesome.com/).

No proper releases planned until I add in the plugin system and test/document the whole system to hell and back - You can still use Docker nightlies, though.


Difegue:
For historical purposes, the old CGI-based code will be left as-is in the [oldcgi](https://github.com/Difegue/LANraragi/tree/oldcgi) branch, alongside the current documentation.

PR #76, opened 2019-01-11
Add file Hdoujin.pm

pratello:


lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+sub plugin_info {
+
+    return (
+        #Standard metadata
+        name      => "Hdoujin",
+        namespace => "Hdoujinplugin",
+        author    => "Pao",
+        version   => "0.1",
+        description =>
+"Collects metadata embedded into your archives by the Hdoujin .json files",
```
Difegue: It'd be good to specify in the description that this plugin does **not** support other HDoujinDL export formats, such as the .txts or ComicInfo .xml files.

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+
+        #Parse it
+        $tags = tags_from_Hdoujin_json($hashjson);
+
+        #Delete it
+        unlink $filepath;
+
+        #Return tags
+        $logger->info("Sending the following tags to LRR: $tags");
+        return ( tags => $tags );
```
Difegue: You can add the title here now that plugins support modifying the title of the archive. 
It'll fail gracefully on older versions of LRR (aka do nothing). 
```suggestion
        return ( tags => $tags,
                 title => $hashjson->{"manga_info"}->{"title"} );
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+}
+
+#tags_from_Hdoujin_json(decodedjson)
+#Goes through the JSON hash obtained from an info.json file and return the contained tags.
+sub tags_from_Hdoujin_json {
+
+    my $hash   = $_[0];
+    my $return = "";
+
+
```
Difegue: I'd add a quick comment about the HDoujin JSON structure here:
```
#HDoujin jsons are composed of a main manga_info object, containing fields for every metadata.
#Those fields can contain either a single tag or an array of tags.
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+#Goes through the JSON hash obtained from an info.json file and return the contained tags.
+sub tags_from_Hdoujin_json {
+
+    my $hash   = $_[0];
+    my $return = "";
+
+    
+    my $tags = $hash->{"manga_info"};
+
+    foreach my $namespace ( keys(%$tags) ) {
```
Difegue: You can do this slightly better by pre-filtering the keys of the hash to remove "title" (and potentially others? not sure about this) and therefore get rid of the ```unless``` statement in the loop:

```suggestion
    #Take every key in the manga_info hash, except for title which we're already processing
    my @filtered_keys = grep { $_ ne "title" } keys(%$tags);
    foreach my $namespace ( @filtered_keys ) {
         my $members = $tags->{$namespace};
    
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+
+
+
+       my $members = $tags->{$namespace} unless $namespace eq "title";
+	   
+	   if(ref($members) eq 'ARRAY'){
```
Difegue: This shortcut is almost a good call, but it will **not** work if the user has enabled _Respect tag namespaces_ in his HDoujin downloader JSONs.   
Without the option, you get the following structure: 
```
    "status": "Completed",
    "chapters": 56,
    "pages": "",
    "tags": [
      "Comedy",
      "Drama",
      "Ecchi",
      "School"
    ],
    "type": "Manga",
    "language": [
      "English"
    ],
    "released": "10/08/2008",
    "reading_direction": "Left to Right",
```

which will give us something like ```status:Completed, tags:comedy, tags:drama, tags:ecchi```, etc. (you might also want to remove namespacing when it's for the "tags" object, it's pretty useless imo)  

With the option, you get this:
```
    "status": "Completed",
    "chapters": 56,
    "pages": "",
    "tags": {
      "Misc": [
        "Comedy",
        "Drama",
        "Ecchi",
        "School"
      ]
    },
    "type": "Manga",
    "language": [
      "English"
    ],
    "released": "10/08/2008",
    "reading_direction": "Left to Right",
```
where the current code simply won't process the "tags" sub-objects at all.  
Having a json with dynamic structuring like that is super annoying, but we'll have to deal with it.  

I recommand filtering out "tags" alongside title in the code I wrote above, and creating another subfunction dedicated to handling the "tags" namespace alone. (with similar ```ref eq 'array'``` magic I guess.)

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+    
+
+			foreach my $tag (@$members){
+
+				$return .= ", " unless $return eq "";
+				$return .= $namespace . ":" . $tag unless $members eq "";
```
Difegue: Beware here: you're adding a comma every time, whereas you don't always add a tag (only if ```$members``` isn't an empty string)  
It's not really harmful in the end since we filter this stuff out, but cleaner output is always appreciated.

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,114 @@
...
+			
+			} 
+			
+		}else{
+		
+				$return .= ", " unless $return eq "";
+				$return .= $namespace . ":" . $members unless $members eq "";
```
Difegue: Same issue here.

lib/LANraragi/Plugin/Hdoujin.pm
pratello: ok, changed

lib/LANraragi/Plugin/Hdoujin.pm
pratello: yeah, in the original, I wanted to filter the "original title" and the "title" with `unless` but after some failed tries.. I settled with only the "title" tag removed... well :)

pratello:
now, it's seems that the code is working when the "respect tags" is activate or not, but I think that I have changed so many things, so.. you prefer an update on all the code here or a new pull request? other things are that I haven't changed the comma things for a cleaner output and the namespace things for the tag.

Difegue:
Update this PR instead of making a new one, it's cleaner that way. I don't mind re-reading everything 👍

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+
+    	my $hash = $_[0];
+	my $hashcontrol = $_[0];
+    	my $return = "";
+	my $tagwRespect = $_[0];
+	my $tagnoRespect = $_[0];
+	my $choosehash = "0";
+	
+	
+	my $tagscontrol = $hashcontrol->{"manga_info"}->{"tags"};
```
Difegue: Remove the $choosehash variable, and make this check further in the code when you're about to call the respect/noRespect methods.

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+		else {
+		
+				$return .= ", " unless $return eq "";
+				$return .= $namespace . ":" . $members unless $members eq "";
+				
+		}
+	
+	}
+    
+		if( $choosehash == "1"){
```
Difegue: Matching the suggestion above:

```suggestion
                #Depending on the structure of the HDoujin .json file, the tags object can either contain an array or an object.
                #We check the type of the object and call the matching sub to get our tags.
		my $tagsobj = $hash->{"manga_info"}->{"tags"};

 		if (ref($tagsobj) eq 'HASH'){
		
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+	
+				return $return . "," . tags_from_wRespect($tagwRespect);
```
Difegue:
```suggestion
				return $return . "," . tags_from_wRespect($hash);
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+		
+		}
+		else {
+		
+				return $return . "," . tags_from_noRespect($tagnoRespect);
```
Difegue:
```suggestion
				return $return . "," . tags_from_noRespect($hash);
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+    my $hash   = $_[0];
+    my $return = "";
+    my $tags = $hash->{"manga_info"}->{"tags"};
+
+    foreach my $namespace ( keys(%$tags) ) {
+
+        my $members = $tags->{$namespace};
+        foreach my $tag (@$members) {
+
+	    $namespace = "Tags";
```
Difegue: Unnecessary namespace imo: It's cleaner at this point to just add the raw tags.

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+            $return .= ", " unless $return eq "";
+            $return .= $namespace . ":" . $tag;
```
Difegue:
```suggestion
            $return .= $tag;
```

lib/LANraragi/Plugin/Hdoujin.pm
```diff
@@ -0,0 +1,192 @@
...
+        return ( error => "No Hdoujin info.json file found in this archive!" );
+    }
+
+}
+
+#tags_from_Hdoujin_json(decodedjson)
+#Goes through the JSON hash obtained from an info.json file and return the contained tags.
+sub tags_from_Hdoujin_json {
+
+    	my $hash = $_[0];
```
Difegue: You don't need to copy $_[0] into 4 different variables since you're never modifying them directly -- just reuse $hash every time.

lib/LANraragi/Plugin/Hdoujin.pm
pratello: I have added this because, example with nhentai every tags goes on the "misc" field (misc:something) and it was visually not good imo, and so with the namespace tags every name become tags:something, like the json file without the option respect on.
now, I thinked and tried with e-hentai, that save the tag like female:something and male:something, so with that namespace renaming I will lost that information, so I think that line is an error.

lib/LANraragi/Plugin/Hdoujin.pm
pratello: like I said before, if I do this I will lost the female or male tag, and every tags goes into "other"

lib/LANraragi/Plugin/Hdoujin.pm
pratello: didn't thinked to reusing $hash, good!

pratello:
now, I think is all corrected expect the tag namespace

lib/LANraragi/Plugin/Hdoujin.pm
Difegue: I personally prefer having unnamespaced tags when there's no precise namespace rather than adding a generic one, but it doesn't really matter I guess.

Difegue:
Good to go!

PR #97, opened 2019-03-15
DateAdded plugin

Utazukin:
This is a simple plugin that adds a tag with a unix timestamp in the "date_added" namespace.  There is one option to choose between the current system time and the date modified time of the archive file.  This is mostly useful for clients to use to implement things like sorting by date, which I've already added to Ichaival.

### DIFF
--- a/lib/LANraragi/Plugin/DateAdded.pm
+++ b/lib/LANraragi/Plugin/DateAdded.pm
@@ -18,9 +18,9 @@ sub plugin_info {
         namespace => "DateAddedPlugin",
         author => "Utazukin",
         version  => "0.1",
-        description => "Adds the date the archive was added as a tag.",
-        oneshot_arg => "Use file modified time.",
-		global_args => ["Use file modified time"]
+        description => "Adds the unix time stamp of the date the archive was added as a tag under the \"date_added\" namespace.",
+        oneshot_arg => "Use file modified time (yes/true), or use current time (no/false). Leaving blank uses the global setting",
+        global_args => ["Use file modified time.\n\"Yes\" or \"True\" means use the modified time.  Anything else means use the current time"]
     );
 
 }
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Plugin/DateAdded.pm
```diff
@@ -0,0 +1,53 @@
...
+sub plugin_info {
+
+    return (
+        #Standard metadata
+        name  => "Date Added",
+        namespace => "DateAddedPlugin",
+        author => "Utazukin",
+        version  => "0.1",
+        description => "Adds the date the archive was added as a tag.",
+        oneshot_arg => "Use file modified time.",
```
Difegue: I'd like a better description for end-users -- Stuff like under which namespace the date ends up, and how should the argument be written for it to be picked up by the plugin ("yes"? "true"? It's both here if I read correctly)

I admit the current system for boolean arguments sucks, but I'm not sure how to make it work best with the current format. Suggestions welcome!

lib/LANraragi/Plugin/DateAdded.pm
```diff
@@ -0,0 +1,53 @@
...
+    my ($title, $tags, $thumbhash, $file, $oneshotarg, @args) = @_;
+
+    #Use the logger to output status - they'll be passed to a specialized logfile and written to STDOUT.
+    my $logger = LANraragi::Utils::Generic::get_logger("Date Added Plugin","plugins");
+
+    #Work your magic here - You can create subroutines below to organize the code better
+
+    $logger->debug("Processing file: " . $file);
+    my $newtags = "";
+    if ($oneshotarg ne "no" && $oneshotarg ne "false" && ($args[0] eq "yes" || $args[0] eq "true" || $oneshotarg eq "yes" || $oneshotarg eq "true")) {
```
Difegue: woah_shit.jpg

this condition is hardly readable, could you break it down a bit?  
You're checking both oneshot and global args for "yes" or "true", which could be shortened by a regex match:
https://stackoverflow.com/questions/6863518/regex-match-one-of-two-words  

Storing the globalarg in a specific variable instead of using $args[0] would also be more readable.

Utazukin:
Thanks for the suggestions.  I haven't really used Perl that much, so I mostly just tried the first thing that worked.  The description should be a little more clear now, and the arguments specify what needs to entered.

PR #128, opened 2019-05-14
Improve tag black-list

Asinin3:
Adds the following to the black-list to help stop LRR's file-tagger from adding invalid tags.
`incomplete`, `ongoing`, `complete`, `various`, `digital`, `decensored`
`decensored` should probably be automatically changed to `uncensored` but I'm not sure how to do that.

### DIFF
--- a/lib/LANraragi/Model/Config.pm
+++ b/lib/LANraragi/Model/Config.pm
@@ -101,7 +101,7 @@ sub get_password {
 
 sub get_tagblacklist {
     return &get_redis_conf( "blacklist",
-"already uploaded, forbidden content, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
+"already uploaded, forbidden content, incomplete, ongoing, complete, various, digital, decensored, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
     );
 }
 
--- a/lib/LANraragi/Model/Config.pm
+++ b/lib/LANraragi/Model/Config.pm
@@ -101,7 +101,7 @@ sub get_password {
 
 sub get_tagblacklist {
     return &get_redis_conf( "blacklist",
-"already uploaded, forbidden content, incomplete, ongoing, complete, various, digital, decensored, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
+"already uploaded, forbidden content, incomplete, ongoing, complete, various, digital, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
     );
 }
 


lib/LANraragi/Model/Config.pm
```diff
@@ -101,7 +101,7 @@ sub get_password {
 
 sub get_tagblacklist {
     return &get_redis_conf( "blacklist",
-"already uploaded, forbidden content, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
+"already uploaded, forbidden content, incomplete, ongoing, complete, various, digital, decensored, translated, russian, chinese, portuguese, french, spanish, italian, vietnamese, german, indonesian"
```
Difegue: I don't want to add too much stuff in the blacklist defaults as it shouldn't withhold potentially wanted info from the user.  
That's subjective of course, but I think `decensored` should stay, even if that means having both it and its uncensored variant.

Difegue:
```suggestion
"already uploaded, forbidden content, incomplete, ongoing, complete, various, digital, translated, russian, chinese, portuguese, english, french, spanish, italian, vietnamese, german, indonesian"
```

Asinin3: Okay, but why add English to the black list? 🤔 Although it LRR's file-tagger does quite often add it to the wrong namespaceS due to unproperly named archives it's weird to block it all-together. It'd be better to add it to block it from everything other than the language namespace. Eng and other short-hand should also probably be automatically changed to English and what-not too. For now I'll just remove Decensored.

Difegue: Felt like it made sense since we're filtering other languages already, but fine. I'll merge it as-is. 👍

Asinin3: Well feel free to change it if that's what you want, I'm not trying to stop you from making changes to your own project of course. I just thought it might've been unintentional since you didn't mention why you were making the change / wanted to know the reasoning behind it. Personally I think it's a bit weird to completely black-list every language since there are likely a lot of people who have some untranslated galleries on-top of their translated galleries. Or a couple translated in a different language simply because no one translated it in their native language. And also, searching for these languages can probably help users find galleries that were tagged after fetching meta-data from the wrong gallery so they can identify the problem there and fix it.

The languages should definitely be limited to he `language:` namespace though, as currently I see the file-tagger incorrectly putting it in the series namespace and what-not since many archives doesn't follow the proper naming schemes.

CirnoT:
Do note that 'digital' has triggered few times for 'group' namespace on partial match so it may be a good idea to update blacklist to perform full tag search instead of partial search. 'ongoing' and 'incomplete' may be useful when fetching from EH so that users can easily identify ones that potentially need to be updated (they're also used for regular translations of chapters, etc.). 'decensored' and 'uncecored' are technically separate tags, one indicates that work was decensored by someone (re-drawn) while other indicates that work was never censored to begin with by artist.

PR #149, opened 2019-07-29
Detect SSL and use WSS instead of WS for WebSocket

CirnoT:
Browsers block insecure WebSocket requests when connection was established over SSL.

Documentation should also be updated for Nginx example configuration to reflect that /batch/socket needs additional Upgrade header when doing proxy from WSS to insecure WS on backend (when SSL is not enabled in LRR): https://www.nginx.com/blog/websocket-nginx/

### DIFF
--- a/public/js/batch.js
+++ b/public/js/batch.js
@@ -66,7 +66,9 @@ function startBatch() {
 
     console.log(command);
 
-    var batchSocket = new WebSocket("ws://" + window.location.host + "/batch/socket");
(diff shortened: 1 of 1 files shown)


Difegue:
Looks good to me -- I've updated the documentation with commit 1b8326714ca04987bcf98ed37dd7e7bfcd8f1b0b.  
Feel free to correct me if I got it wrong since I don't use nginx.

PR #150, opened 2019-07-29
Update E-Hentai plugin

CirnoT:
Updates E-Hentai plugin now that ExHentai frontend was closed down.

- Added new 'star' cookie value, required to view fjorded content from E-Hentai frontend if user has appropriate permissions to do so.
- Add option to disable fetching using thumbnails
- Update thumbnail fetcher to also search for `language` namespace if one specified - this solves issue where thumbnail fetch is executed first and selects first result which is not in user desired language (for example Spanish).
- Have title search ignore Language and User blacklists by default (language can be overridden when necessary via `language` namespace).
- Update domains to use SSL (avoid unnecessary redirection that could possibly be seen as request hit from E-Hentai backend)


Difegue:
Damn, I kinda wanted to be the one deleting the ExHentai.pm file after all the pain it's put me through figuring out the panda magic, but you beat me to it. 🙏 

Thanks for both this and the WSS thing; I'll try to look at them ASAP.

CirnoT:
- Add additional sleep period to avoid getting banned for excessive requests by EH (detected by warning message "You are opening pages too fast, thus placing a heavy load on the server. Back down, or your IP address will be automatically banned." - failure to abide by it will result in automatically having account flagged and every IP accessing it marked and banned for period of time depending on count of bans in specific period of time. They can be as short as an hour or as long as whole day.

Difegue:
A warning about auto-plugin and how it can get you banned for rapid-requesting isn't a bad idea too -- Gonna edit the settings page to reflect that.

PR #159, opened 2019-08-02
Add back support for using ExHentai

CirnoT:
Looks like ExHentai will be coming back so add support for it back to E-Hentai plugin while maintaining previous changes.


Difegue:
Hmm, the previous version tried to be smart and dynamically use ex if possible, but now that boolean arguments in plugins are a thing I guess it can be a toggle.  

I'll look at this in detail later.

CirnoT:
The reason why it is not using Ex by default is to be future-proof in case it goes down again - that way it doesn't require instant update to fix it for users.

PR #160, opened 2019-08-02
Update EH plugin to fetch category and get gallery ID from source tag

CirnoT:
- Add ability to fetch gallery ID from `source` tag and use it instead of searching (oneshot arguments still override that behavior)
- Have plugin fetch category from EH API, lowercase it and add it under `category` namespace as tag.


CirnoT:
This will close https://github.com/Difegue/LANraragi/pull/160 if merged

Difegue:
Looks good to me. Thanks again!

Issue #114, opened 2019-04-30
[Suggestion] Usability Improvements


Asinin3:
LRR has a very nice web-interface and I've been impressed overall, however there are some QoL changes I'd like to see to make it more usable.

- [ ] Option to scale images to browser Width.
   - [ ] Options to Limit how much an image gets resized e.g don't scale above native image resolution, or limit it to only scaling up by a specified percentage e.g 200%, 300%, 400%
- [x] Retain search terms and page number when returning to the library view after editing a gallery's meta-data or viewing a gallery. (I'd also like to see this working with the browser's back button, I guess the only way to do that would be to change the URL based on the page count and search terms?)
- [x] Delete archive button (with confirmation) in the library view
- [ ] Improve how LRR finds cover images, by e.g trying to find portrait oriented images first, or having certain (specified) filenames be preferred over others e.g "!cover".
- ~~Remove "hover to show tags" and just make it show tags whenever hovering over the gallery. I can see how the tool-tip might accidentally get in the way, so perhaps a 0.5sec hover delay would fix that?~~
- [x] Add view random gallery button to the bottom of page navigation.
- ~~Change the "Disable Automatic Bookmarking" setting to have a (configurable) value of pages you can limit it to. Allowing users to disable automatic bookmarking on galleries with e.g less than 40 images. Change it to 0 to disable it completely, or uncheck the setting to have it be enabled on everything.~~
- [ ] Ability to hide "Series" column from list view
- [ ] Add support for shorthand namespaces in the search field e.g `a:`=`artist:`, `c:`=`character`, `m:`=`male:`, `f:`=`female:`  just like how it's done on E-H.
- [x] Remove name-spaces from the tag column in list-view and color-code them instead (for more space).
  -  Alternatively, just switch to using the shorthand namespaces in the list-view e.g `a:` = `artist:`
- [x] Add a short delay (roughly 0.2sec?) to updating search results. I found making searches in LRR a bit hard as LRR constantly refines it's search results as you type resulting in typo's due to the delay before keys register.
- [ ] In the batch tagger make gallery names link to their meta-data page or add a button.
- [ ] Implement a way to hide galleries which have tags or are not already ticked from the batch-tagger.
When I had a few galleries in my large library missing tags, it was quite hard to find them.
- [ ] Add a way to rename and if possible move gallery archives/folders inside the meta-data page, if moving is too hard simply changing the path the database points to would be enough.
[...]

PR #175, opened 2019-08-16
Disabled image select in reader

AnyByte:


Difegue:
Fixes the comment in #114.  
It's a small fix but I kept forgetting to do it, so thanks a lot!  

The test suite is also finally working for PRs thanks to Actions v2 so that's good.

PR #187, opened 2019-09-05
Update CSS for themes with styled awesomplete

CirnoT:
This tries to follow already established colors for each theme.


Difegue:
I'll probably be refining the result a bit, but this does like 90% of the work so I'll be merging it right now. Thanks a lot!

PR #191, opened 2019-09-09
Properly handle stacking favorites filter and search bar term

CirnoT:
Autocomplete adds and requires tags to be comma separated however in case where haystack is "female:sole female, already uploaded" and needle (search term) is "already uploaded, " it will fail (the same haystack will work fine with needle "female:sole female,"). As such, add simple ```.replace(",","")``` to search bar event listener to silently ignore comma when passing needle to datatables.

---

Properly handle stacking favorites filter and search bar term. This is done by moving search logic completely to ```favTagSearch``` and using it to stack two search terms in datatables by first applying regex search on favorites OR on tags column, and then applying smart search filter on table itself. This allows us to take benefit of the fact that .search() is subtractive in datatables.

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -59,10 +59,13 @@ function favTagSearch() {
 
 	//Perform search in datatables field with our own regexes enabled and smart search off
 	if (searchQuery !== ")") {
-		arcTable.search(searchQuery, true, false).draw();
+		arcTable.column('.tags.itd').search(searchQuery, true, false);
+		arcTable.search($('#srch').val().replace(",", ""), false, true);
+		arcTable.draw();
 	} else {
-		//clear
-		arcTable.search("", false, true).draw();
+		// no fav filters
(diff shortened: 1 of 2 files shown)


public/js/index.js
```diff
@@ -187,7 +190,7 @@ function loadTagSuggestions() {
 
 			// Perform a search when a tag is selected
 			Awesomplete.$('#srch').addEventListener("awesomplete-selectcomplete", function() {
-				arcTable.search($('#srch').val()).draw();
+				favTagSearch();
```
Difegue: Since this is used for all forms of search, calling the function `favTagSearch()` is now misleading; The function should be renamed to something like `performSearch()` and the comment on line 43/44 updated accordingly.

CirnoT: Updated

Difegue:
👍

PR #193, opened 2019-09-09
Bring back parody as alternative to series in createNamespaceColumn

CirnoT:


public/js/index_datatables.js
```diff
@@ -106,6 +106,7 @@ function createNamespaceColumn(namespace, type, data) {
 		if (data === "")
 			return "";
 
+		if(namespace == "series") namespace = "(?:series|parody)";
```
Difegue: The namespace switcheroo should be done around line 44 instead imo.

CirnoT: I disagree. My implementation will still work when you'll implement ability to move/change columns.

Difegue: True; I was thinking that when custom columns are implemented, users'd just put a column for either series or parody, but that'd just be more annoying that doing the check by default. Merging I guess

PR #200, opened 2019-09-27
Update docker.md

dudeitssm:
Added a bit of clarification about the --bind and --volume mounting options. 
Users on older docker versions below 17.06 have to use the --volume option, dood.

On Fedora 30 Server Edition, the version in the repos is 1.13.1.


Difegue:
Looks good at a glance. Thanks!

PR #213, opened 2019-10-06
Perform search on button click, not on keyup

CirnoT:
Adds 'Apply Filter' button, similar to EH that performs search. This stops search from being executed on every key press in search bar and also disables automatic search when new value is selected from Awesomplete (for now at least, since it doesn't properly handle "" for tags with whitespaces yet). Additionally this decreases min-width on .stdbtn to 100px as 150px buttons are too big with search bar.

### DIFF
--- a/public/js/index_datatables.js
+++ b/public/js/index_datatables.js
@@ -63,6 +63,14 @@ function initIndex(pagesize) {
 	$('#subsrch').click(function () {
 		performSearch();
 	});
+	$('#subsrch').keyup(function (e) {
+		if(e.defaultPrevented) {
+			return;
+		} else if(e.key == "Enter") {
+			performSearch();
+		}
+		e.preventDefault();
+	});
 
 	$('#clrsrch').click(function () {
 		//clear all favtags
--- a/public/js/index_datatables.js
+++ b/public/js/index_datatables.js
@@ -60,7 +60,7 @@ function initIndex(pagesize) {
 	});
 
 	//add datatable search event to the local searchbox and clear search to the clear filter button
-	$('#srch').keyup(function () {
+	$('#subsrch').click(function () {
 		performSearch();
 	});
 
--- a/public/themes/ex.css
+++ b/public/themes/ex.css
@@ -110,7 +110,7 @@ tr.gtr0{background:#4f535b}
 tr.gtr1{background:#363940}
 
 /* input */
-.stdbtn{font-size:8pt;color:#f1f1f1;background:#34353b;border:2px solid #000000;height:21px;margin:4px 1px 0px 1px;padding:0px 4px 1px 4px; min-width: 100px; cursor: pointer;}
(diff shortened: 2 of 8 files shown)


public/themes/g.css
```diff
@@ -163,7 +163,7 @@ tr.gtr1 {
     height: 21px;
     margin: 4px 1px 0;
     padding: 0 4px 1px;
-    min-width: 150px;
+    min-width: 100px;
```
Difegue: A lot of the UI in the app uses .stdbtn and therefore changing the width across all of it to 100px might break a bunch of layouts.  
I'd prefer a dedicated .searchbtn complement class that overrides the min-width here.

public/themes/ex.css
```diff
@@ -110,7 +110,7 @@ tr.gtr0{background:#4f535b}
 tr.gtr1{background:#363940}
 
 /* input */
-.stdbtn{font-size:8pt;color:#f1f1f1;background:#34353b;border:2px solid #000000;height:21px;margin:4px 1px 0px 1px;padding:0px 4px 1px 4px; min-width: 150px; cursor: pointer;}
+.stdbtn{font-size:8pt;color:#f1f1f1;background:#34353b;border:2px solid #000000;height:21px;margin:4px 1px 0px 1px;padding:0px 4px 1px 4px; min-width: 100px; cursor: pointer;}
```
Difegue: See comment on g.css

public/themes/modern.css
```diff
@@ -173,7 +173,7 @@ tr.gtr1{background:#363940}
     margin: 1px;
     outline: 0 none;
     padding: 0 4px 1px;
-    min-width: 150px;
+    min-width: 100px;
```
Difegue: ditto

public/themes/modern_clear.css
```diff
@@ -199,7 +199,7 @@ tr.gtr1{background:#34495E}
     margin: 1px;
     outline: 0 none;
     padding: 0 4px 1px;
-    min-width: 150px;
+    min-width: 100px;
```
Difegue: ditto

public/themes/modern_red.css
```diff
@@ -215,7 +215,7 @@ tr.gtr1 {
     font-size: 9pt;
     height: 28px;
     margin: 1px;
-    min-width: 150px;
+    min-width: 100px;
```
Difegue: ditto

public/js/index_datatables.js
```diff
@@ -58,7 +58,7 @@ function initIndex(pagesize) {
 	});
 
 	//add datatable search event to the local searchbox and clear search to the clear filter button
-	$('#srch').keyup(function () {
+	$('#subsrch').click(function () {
```
Difegue: I'd like for this to trigger on the Enter button as well: Converting the entire thing to `<form>` is a solution, but you can just add a keyup on the search folder that only triggers `performSearch` if the key is Enter.

Issue #151, opened 2019-07-30
A new plugin metadata suggestions


elakwmq1:
Hi i using "koromo copy"(hitomi manga downloader) program.
this program can save gallery metadata for made json file.
but LRR not support this json file.
I have so many files that it is hard to query with ehenti metadata plugin
can you add plugin with "koromo copy"s json file?
i can give json files if you want. thx

koromo copy github: https://github.com/dc-koromo/koromo-copy/issues
i attach info.txt(metadata file)
[Info.txt](https://github.com/Difegue/LANraragi/files/3445113/Info.txt)

Difegue:
I recommend you look at the [Eze](https://github.com/Difegue/LANraragi/blob/dev/lib/LANraragi/Plugin/Eze.pm) plugin, who also imports json files, and try to modify it to suit your needs.  

Sadly, I can't really implement the billion different metadata json formats that exist out there.

Difegue:
Closed by #214. Great stuff!

PR #214, opened 2019-10-09
Add metadata plugin for Koromo (#151)

CirnoT:


lib/LANraragi/Plugin/Koromo.pm
```diff
@@ -0,0 +1,143 @@
...
+        unlink $filepath;
+
+        #Return tags
+        $logger->info("Sending the following tags to LRR: $tags");
+        return ( tags => $tags );
+
+    }
+    else {
+
+        return ( error => "No eze info.json file found in this archive!" );
```
Difegue:
```suggestion
        return ( error => "No koromo info.json file found in this archive!" );
```

lib/LANraragi/Plugin/Koromo.pm
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    foreach my $tag ( @$characters ) {
+
+        $return .= ", " unless $return eq "";
+        $return .= "character:" . $tag;
+        
+    }
+
+    foreach my $tag ( @$series ) {
```
Difegue: $series is parsed twice, here and on line 114

PR #215, opened 2019-10-19
Have archive title in batch tagging be link to metadata edit

CirnoT:
Based on list from https://github.com/Difegue/LANraragi/issues/114

### DIFF
--- a/lib/LANraragi/Controller/Batch.pm
+++ b/lib/LANraragi/Controller/Batch.pm
@@ -33,7 +33,7 @@ sub index {
         if (-e $zipfile) {
             $arclist .=
                 "<li><input type='checkbox' name='archive' id='$id' class='archive' >"
-                . "<label for='$id'> $title</label></li>";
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Controller/Batch.pm
```diff
@@ -33,7 +33,7 @@ sub index {
         if (-e $zipfile) {
             $arclist .=
                 "<li><input type='checkbox' name='archive' id='$id' class='archive' >"
-                . "<label for='$id'> $title</label></li>";
+                . "<label for='$id'> <a href='./edit?id=$id' style='text-decoration: none;'>$title</a></label></li>";
```
Difegue: From a UX perspective it's super weird here that the checkbox and label don't perform the same action when clicking on them, I think a button on the side would be more appropriate.  

Also, putting <a> tags in the checklist breaks colors on both the Yotsugi and Nadeko themes.

CirnoT:
I'll revisit it later when I have some actual time.

PR #217, opened 2019-10-19
Make clicking on a gallery's tag fill the search box with namespace:tag

CirnoT:
Make clicking on a gallery's tag fill the search box with namespace:tag format instead of just the tagname.

Based on https://github.com/Difegue/LANraragi/issues/114

### DIFF
--- a/public/js/index_datatables.js
+++ b/public/js/index_datatables.js
@@ -104,12 +104,13 @@ function createNamespaceColumn(namespace, type, data) {
 		if (data === "")
 			return "";
 
-		if(namespace == "series") namespace = "(?:series|parody)";
-		regex = new RegExp(".*"+namespace+":\\s?([^,]*),*.*","gi"); // Catch last namespace:xxx value in tags
+    var namespaceRegEx = namespace;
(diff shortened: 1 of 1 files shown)


public/js/index_datatables.js
```diff
@@ -109,7 +109,7 @@ function createNamespaceColumn(namespace, type, data) {
 		match = regex.exec(data);
 
 		if (match != null) {
-			return '<a style="cursor:pointer" onclick="$(\'#srch\').val($(this).html()); arcTable.search($(this).html()).draw();">' +
+			return '<a style="cursor:pointer" arc-namespace="' + key + '" onclick="$(\'#srch\').val($(this).attr(\'arc-namespace\') + \':\' + $(this).html()); arcTable.search($(this).attr(\'arc-namespace\') + \':\' + $(this).html()).draw();">' +
```
Difegue: there's no `key` variable here, you'll need to use `namespace`. Be careful tho that namespace can be modified to the `(?:series|parody)` regex on line 107, so you'll need to copy that to another variable.

PR #220, opened 2019-10-22
fix macOS installation problems (v2)

Bl4Cc4t:
and also fixed some punctuation errors.

### DIFF
--- a/tools/install.pl
+++ b/tools/install.pl
@@ -126,17 +126,14 @@ if ( $back || $full ) {
 
     # libarchive is not provided by default on macOS, so we set the correct env vars
     # to successfully compile Archive::Extract::Libarchive and Archive::Peek::Libarchive
+    my $pre = "";
     if ($Config{"osname"} eq "darwin") {
         say("Setting Environmental Flags for macOS");
-        if (system("export CFLAGS=\"-I/usr/local/opt/libarchive/include\" && \\
-          export PKG_CONFIG_PATH=\"/usr/local/opt/libarchive/lib/pkgconfig\" && \\
-          cpanm --installdeps ./tools/. --notest") != 0 ) {
-            die "Something went wrong while installing Perl modules - Bailing out.";
-        }
-    } else {
-        if ( system("cpanm --installdeps ./tools/. --notest") != 0 ) {
-            die "Something went wrong while installing Perl modules - Bailing out.";
-        }
+        $pre = "export CFLAGS=\"-I/usr/local/opt/libarchive/include\" && \\
+          export PKG_CONFIG_PATH=\"/usr/local/opt/libarchive/lib/pkgconfig\" && ";
+    }
+    if ( system($pre . "cpanm --installdeps ./tools/. --notest") != 0 ) {
+        die "Something went wrong while installing Perl modules - Bailing out.";
     }
 }
 
--- a/tools/install.pl
+++ b/tools/install.pl
@@ -5,6 +5,7 @@ use warnings;
 use utf8;
 use open ':std', ':encoding(UTF-8)';
 use Cwd;
+use Config;
 
 use feature qw(say);
 use File::Path qw(make_path);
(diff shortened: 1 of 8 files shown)


tools/install.pl
```diff
@@ -123,8 +124,19 @@
     );
     say("\r\nInstalling Perl modules... This might take a while.\r\n");
 
-    if ( system("cpanm --installdeps ./tools/. --notest") != 0 ) {
-        die "Something went wrong while installing Perl modules - Bailing out.";
+    # libarchive is not provided by default on macOS, so we set the correct env vars
```
Difegue: I suppose you installed it using homebrew? Would be good to add that detail either here or in the documentation.

tools/install.pl
```diff
@@ -123,8 +124,19 @@
...
+    # to successfully compile Archive::Extract::Libarchive and Archive::Peek::Libarchive
+    if ($Config{"osname"} eq "darwin") {
```
Difegue: I think the export calls should be done outside of the if statement -- That way you'll only have one `cpanm` call at the end here, something like that:

```
        if ($Config{"osname"} eq "darwin") {
             `export CFLAGS=\"-I/usr/local/opt/libarchive/include\"`
             `export PKG_CONFIG_PATH=\"/usr/local/opt/libarchive/lib/pkgconfig\"`
        }

        # common code between darwin and linux
        if ( system("cpanm --installdeps ./tools/. --notest") != 0 ) {
            die "Something went wrong while installing Perl modules - Bailing out.";
        }
```

tools/install.pl
Bl4Cc4t: Yes, I did. I’m actually working on a formula for LANraragi since the beginning :)
As for manual installation, using Homebrew is obviously also the best/fastest way of doing it.
I had to install the following: `cpanminus libarchive libjpeg libpng node openssl perl redis`

tools/install.pl
Bl4Cc4t: That does not work, since `system()` creates a subshell with no/the env vars provided by your profile. I tried that approach before, I don’t like having doubles either but else it does not work sadly

Bl4Cc4t: I created a fix for this problem, pretty stupid of me to not think of that (not knowing this Perl is no excuse :D)
https://github.com/Bl4Cc4t/LANraragi/commit/e0a306a6205dd09feb5167d6c867da598e75ae27

Difegue: The PR is automatically updated as you push commits, so no need. 👍

Difegue:
Thanks for your help! I started a review on the earlier PR but copied most of it here.  
I'll be able to merge this PR easier since it's not going into `dev` now. 

I was unsatisfied with `File::ChangeNotify` since adding a subdirectory was terribly broken on my end (WSL), are you having better success on macOS? I'll admit I didn't get to the point of firing up a VM to try.  

If I were to merge this branch into `dev` for global release I'd likely have to go dig around in their Linux implementation to see what goes wrong. Might be interesting to add a dedicated Mac implementation using [Mac::FSEvent](https://metacpan.org/pod/Mac::FSEvents), but that'll likely come much later. 😁

tools/install.pl
Bl4Cc4t: oh, I did not knew that. So closing my first not necessary D:

Bl4Cc4t:
Sure thing! Sorry for the hassle with this 2nd PR :V

I just tried it, it works like it should. No errors on my side.

Oh, `Mac::FSEvent`… I remember having a very bad time fiddling around with that.
What errors do you get with `File::ChangeNotify`?

Difegue:
Well with the current code, dropping a subfolder on WSL basically triggers no events.  
Therefore, any files that are in the subfolder aren't added to the filemap/database.  

It might work properly on real Linux but break on WSL since that has a different inotify implementation; I'll do some more advanced testing to figure this out.  

In the meantime, I can certainly merge the work you've done here and use that as a base if it works fine on mac. 🎊

Bl4Cc4t:
Ah yeah, that could be the case. I never used WSL, but I guess it’s pretty much like `wine` for macOS (which also has a huge amount of errors).

PR #228, opened 2019-11-05
Homebrew with stable release

Bl4Cc4t:
Edited docs/the [formula](https://github.com/Bl4Cc4t/homebrew-other/commit/9a82cfa890f307d59afbb6ea6561e89fe85ad6d2) to adjust to the stable release we now have :)
Also added some instructions for migrating. I placed them at the beginning, I think that makes the most sense, does it?

Should I submit a PR to [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) at this point?

### DIFF
--- a/tools/Documentation/installing-lanraragi/macos.md
+++ b/tools/Documentation/installing-lanraragi/macos.md
@@ -1,5 +1,28 @@
 # Homebrew \(macOS/Linux\)
 
+## Migration
+
+To use all your existing files within a brewed LRR, you can issue the following commands:
+
+```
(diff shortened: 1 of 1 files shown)


Difegue:
Looks good to me! If the formula is good enough for core's standards, it's worth trying a PR now.  

( Just for info, I started refactoring logging today to allow for overriding the log directory, so for 0.6.7 there can be support for Console.app too.)

Bl4Cc4t:
Just created a PR: https://github.com/Homebrew/homebrew-core/pull/46343

PR #241, opened 2019-12-19
readd mkdirs in homebrew script

Bl4Cc4t:


Difegue:
Right, I brazenly removed all mkdirs since the software creates them on the fly, but I forgot about Redis which starts before us. meh.

Normally only the log folder needs to be created, but might as well make sure and create all 3.

PR #248, opened 2020-01-22
Release 0.6.8

Difegue:


public/js/index_datatables.js
```diff
@@ -53,11 +53,16 @@ function initIndex(pagesize) {
 				'name': 'tags',
 				'orderable': false,
 				'render': tagsColumnDisplay
-			},{
+			},{ // The columns below are invisible and only meant to add extra parameters to a search.
 				className: 'isnew itd',
 				visible: false,
-				'data': 'isnew',
+				'data': null,
```
Difegue: whoa shit this is wrong, the `null` data should go in the untagged column

PR #249, opened 2020-01-22
redis-config update, LRR_DATABASE_DIRECTORY additions

Bl4Cc4t:
I had trouble starting redis, so I first thought the config wasn’t working anymore. Apparently, v5.0.0 introduced some name changes (`slave -> replica`), but they made sure it’s backwards compatible (why wouldn’t they).

As it turned out, redis could not start because the database directory was not found. It has no error message when you specify a non existent directory in the config, which is kinda weird...

So, along with the added `export` and `mkdir` for `$LRR_DATABASE_DIRECTORY` in the start script, I added the “newer” config file which I created as a test.
The only change I made except from copying the already set option was setting the `daemonize` option to `yes`.


Difegue:
The redis.conf file I bundle is getting a bit old indeed -- Guess I'll update the docker version one of these days as well. 

Looks good otherwise, mergin' !

PR #252, opened 2020-01-25
Do not use sudo when installing with perlbrew.

hakonhagland:
A new PR from [pullrequest.club](https://pullrequest.club): Mention in the docs that when installing with `perlbrew` from source code, you should not run the `lanraragi-installer` with `sudo` as `sudo` will  modify the environment variables such that the installer will switch to the vendor `perl` (system `perl`) instead.


Difegue:
Ah, that is correct! I don't dabble too much in vendor perl/perlbrew differences now that I mostly use Docker. Thanks for noticing!

Issue #274, opened 2020-04-21
Mayriad's LRR Bugs and Suggestions


Mayriad:
Firstly, thanks a lot for making this fantastic library manager. I think I have done a good amount of testing and found enough issues, so I will open a big thread here. I cannot come up with a clever title so I just used my standard format.

I will edit this if I find new problems or suggestions to add. A lot of suggestions can be made, but I will focus on bugs for now.

## 1. Index

Bugs:

- [ ] 4.1. Plugins.pm missing the level after "::Plugin" in the namespace
- [ ] 4.2. Plugin update not effective until LANraragi is restarted
- [ ] 4.3. Plugin upload not working after failed upload
- [ ] 4.4. Admin settings not saving
- [ ] 4.5. Cleaning and resetting database currently broken
- [ ] 4.6. Shinobu permanently omitting files after it stops half way
- [ ] 4.7. Shinobu not responding to new archives if they are added in folders
- [ ] 4.8. Shinobu adding new archives twice if they are not added in folders
- [ ] 4.9. Batch tagging still throwing code 1006
- [ ] 4.10. Checkboxes too tiny on the batch tagging page
- [ ] 4.11. Undefined value error from the E-Hentai metadata plugin

Suggestions:

- [ ] 5.1. Namespace shorthand and naming
- [ ] 5.2. More options for the E-Hentai plugin
- [ ] 5.3. Better gallery lookup routine for the E-Hentai plugin
- [ ] 5.4. Different treatment for the dates from the DateAdded plugin
- [ ] 5.5. Delay between requests for auto-plugin
- [ ] 5.6. Empty tag blacklist
- [ ] 5.7. LANraragi thread in EH forums

## 2. Setup

LANraragi for Windows 0.6.9 (MSI installer) on Microsoft Windows 10 version 10.0.18363.720. The LANraragi content folder is on the Windows PC, but the archives are stored on a NAS and accessed via a WSL mount point in the content folder. The PC and the NAS are connected via WIFI 5. My NAS cannot run docker, and Homebrew is either impossible or too difficult for me, so I installed LANraragi on Windows. Maybe some of the bugs below are caused by my setup, but I cannot know.

## 3. Myself

[...]

PR #283, opened 2020-04-26
(#274-4.1) Add code to handle plugin types

Mayriad:
Now the "upload plugin" functionality should work again. The coding style in `Plugins.pm` is inconsistent but I tried to stay consistent.

### DIFF
--- a/lib/LANraragi/Controller/Plugins.pm
+++ b/lib/LANraragi/Controller/Plugins.pm
@@ -155,16 +155,15 @@ sub process_upload {
         if ($filetext =~ /package LANraragi::Plugin::(Login|Metadata|Scripts)::/) {
             $plugintype = $1;
         } else {
-            $logger->error("Could not find a valid plugin package type in the plugin "
-              . "\"$filename\"!");
+            my $errormess= "Could not find a valid plugin package type in the plugin \"$filename\"!";
+            $logger->error($errormess);
 
             $self->render(
                 json => {
                     operation => "upload_plugin",
                     name      => $file->filename,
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Controller/Plugins.pm
```diff
@@ -148,7 +148,30 @@ sub process_upload {
...
+
+            $self->render(
+                json => {
+                    operation => "upload_plugin",
+                    name      => $file->filename,
+                    success   => 0,
+                    error     => "Could not find a valid plugin package type in the "
+                      . "plugin \"$filename\"!"
+                }
+            );
```
Difegue, line 169: Put the error message in a single variable to avoid retyping it:  
```suggestion
            my $errormess= "Could not find a valid plugin package type in the plugin \"$filename\"!";
            $logger->error($errormess);

            $self->render(
                json => {
                    operation => "upload_plugin",
                    name      => $file->filename,
                    success   => 0,
                    error     => $errormess
                }
            );
```

Mayriad: I know. This is just how the existing code does it and, as I mentioned, I tried to stay consistent.

Difegue: Well, the only other place in that file where we spit a success 0 json with an error code has slightly different text between the log and output versions. (Since the output has html tags in it as a quick 'n dirty way to make it fit in the notification) 

Here tho they're exactly the same, so I'll take legibility over consistency. 😛 Thanks for the change.

PR #284, opened 2020-04-26
(#274-4.10) Fix tiny batch tagging checkboxes

Mayriad:
I set the `width` of the checkboxes below "override plugin global arguments" to `auto` to use the default size and stay consistent with every other checkbox. This will be override the 21px by 21px big checkboxes set by some LANraragi themes, which only apply to the E-Hentai plugin on this page.

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -307,6 +307,10 @@ a:hover > .page-number {
   width: 60%;
 }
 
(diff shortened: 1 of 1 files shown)


Difegue:
This is exclusive to the `tag-options` table in Batch, so it's fine by me. Mergin', thanks!

PR #285, opened 2020-04-26
(#274-5.2) Add more functionalities to E-Hentai metadata plugin

Mayriad:
I added three functionalities and corresponding options:
- Save the original Japanese title when available instead of the English or romanised title.
- Fetch additional timestamp (time posted) and uploader metadata.
- Search expunged galleries as well.

The first two are from my own plugin `MEMS.pm`, which will be submitted in due time. The coding style in `EHentai.pm` is inconsistent but I modified my original code to stay consistent.

I also corrected the description of the "fetch using thumbnail first" option to "fetch using cover thumbnail instead of archive title". The existing description is misleading because "using thumbnail first" sounds like there is a second fallback step but there is not. It also does not specify that it uses the cover to match covers on EH.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/EHentai.pm
+++ b/lib/LANraragi/Plugin/Metadata/EHentai.pm
@@ -31,7 +31,11 @@ sub plugin_info {
         parameters => [
             { type => "string", desc => "Forced language to use in searches" },
             { type => "bool",   desc => "Save archive title" },
-            { type => "bool",   desc => "Fetch using thumbnail first" },
+            { type => "bool",   desc => "Save the original Japanese title when available instead of the English or "
+              . "romanised title" },
+            { type => "bool",   desc => "Fetch additional timestamp (time posted) and uploader metadata" },
+            { type => "bool",   desc => "Fetch using cover thumbnail instead of archive title" },
+            { type => "bool",   desc => "Search expunged galleries as well" },
             { type => "bool",   desc => "Use ExHentai (enable to search for fjorded content without star cookie)" }
         ],
         oneshot_arg => "E-H Gallery URL (Will attach tags matching this exact gallery to your archive)"
@@ -44,7 +48,7 @@ sub get_tags {
 
     shift;
     my $lrr_info = shift;    # Global info hash
-    my ( $lang, $savetitle, $usethumbs, $enablepanda ) = @_;    # Plugin parameters
+    my ( $lang, $savetitle, $jpntitle, $additionaltags, $usethumbs, $expunged, $enablepanda ) = @_;    # Plugin parameters
 
     # Use the logger to output status - they'll be passed to a specialized logfile and written to STDOUT.
     my $logger = get_logger( "E-Hentai", "plugins" );
@@ -70,8 +74,8 @@ sub get_tags {
             $lrr_info->{archive_title},
             $lrr_info->{existing_tags},
             $lrr_info->{thumbnail_hash},
-            $lang, $lrr_info->{user_agent},
-            $domain, $usethumbs
+            $lrr_info->{user_agent},
+            $domain, $lang, $usethumbs, $expunged
         );
     }
 
@@ -91,7 +95,7 @@ sub get_tags {
         $logger->debug("EH API Tokens are $gID / $gToken");
     }
 
-    my ( $ehtags, $ehtitle ) = &get_tags_from_EH( $gID, $gToken );
+    my ( $ehtags, $ehtitle ) = &get_tags_from_EH( $gID, $gToken, $jpntitle, $additionaltags );
     my %hashdata = ( tags => $ehtags );
 
     # Add source URL and title if possible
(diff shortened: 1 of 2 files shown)


lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -106,24 +110,30 @@ sub get_tags {
 
 sub lookup_gallery {
 
-    my ( $title, $tags, $thumbhash, $defaultlanguage, $ua, $domain, $usethumbs) = @_;
+    my ( $title, $tags, $thumbhash, $ua, $domain, $defaultlanguage, $jpntitle, $additionaltags, $usethumbs,
```
Difegue, line 113: You don't actually need `$jpntitle` and `$additionaltags` in this sub -- It'd be better to not include them since the signature is getting super long already 😛

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -140,6 +150,11 @@ sub lookup_gallery {
       . "?advsearch=1&f_sname=on&f_stags=on&f_sdt2=on&f_spf=&f_spt=&f_sfu=on&f_sft=on&f_sfl=on"
       . "&f_search=" . uri_escape_utf8(qw(").$title.qw("));
 
+    #Include expunged galleries in the search if the option is enabled.
+    if ( $expunged ) {
+        $URL = $URL . "&f_sh=on";
```
Difegue, line 155: My e-h search-fu is a bit rusty, expunged really can use both `f_sh=on` and `fs_exp=1` as query parameters?

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -65,8 +69,8 @@ sub get_tags {
     } else {
         # Craft URL for Text Search on EH if there's no user argument
         ( $gID, $gToken ) =
-          &lookup_gallery( $lrr_info->{archive_title}, $lrr_info->{existing_tags}, 
-                           $lrr_info->{thumbnail_hash}, $lang, $lrr_info->{user_agent}, $domain, $usethumbs);
+          &lookup_gallery( $lrr_info->{archive_title}, $lrr_info->{existing_tags}, $lrr_info->{thumbnail_hash},
+                           $lrr_info->{user_agent}, $domain, $lang, $jpntitle, $additionaltags, $usethumbs, $expunged);
```
Difegue, line 73: This change seems to be what breaks the tests.  
`$ua` aka `$lrr_info->{user_agent}` is for some reason not an object reference anymore, haven't looked further.

lib/LANraragi/Plugin/Metadata/EHentai.pm
Mayriad: I think I included them to pass them to the function that calls the EH API.

lib/LANraragi/Plugin/Metadata/EHentai.pm
Mayriad: Yeah, image search and title search use different parameters.

lib/LANraragi/Plugin/Metadata/EHentai.pm
Mayriad: I changed the order of these arguments because the order was weird, but perhaps that order was required?

lib/LANraragi/Plugin/Metadata/EHentai.pm
Difegue: dangit tenboro

lib/LANraragi/Plugin/Metadata/EHentai.pm
Difegue: Aye, it's needed for `get_tags_from_EH`, which isn't called from `lookup_gallery` but from the base `get_tags` sub.

Mayriad: You are right. The function call in line 73 should be changed, too.

lib/LANraragi/Plugin/Metadata/EHentai.pm
Mayriad: I just checked the code and I have no idea what is wrong here. I will have to leave this one to you.

Difegue: Thought about it a bit more and it's dead simple: The tests in plugin.t call this function directly, so you'll need to change the call in the tests too.

Mayriad: Sorry, somehow I thought I did not have this file. The function call has been updated.

Difegue:
There's a merge conflict, it'd be better if you rebased your branch on top of the current dev so I don't have to manually resolve conflicts. 🙇‍♂️

Difegue:
Perfect! Thank you.

CirnoT:
> I also corrected the description of the "fetch using thumbnail first" option to "fetch using cover thumbnail instead of archive title". The existing description is misleading because "using thumbnail first" sounds like there is a second fallback step but there is not. It also does not specify that it uses the cover to match covers on EH.

The function only returns if `gId` and `gToken` from `ehentai_parse` are not "", otherwise it continues as usual with title search.

Difegue:
That is correct after further re-reading -- Blame's on me since I didn't review the PR properly, so I pushed a fix out.

Mayriad:
No no it is 100% my fault, sorry. I guess I was too tired and did not spend enough time with that plugin.

Issue #279, opened 2020-04-21
[Suggestion] Namespace shorthand and naming


Mayriad:
*This issue comes from #274 my big list of LRR bugs and suggestion. Please refer to that to see details of my setup and usage.*

Currently LANraragi does not support namespace shorthand in search terms, which means terms like `l:english` will not work and you have to type `language:english`. Allowing shorthand like `l:` can make it much easier to type these search terms. I understand the difficulty here since LANraragi allows custom namespaces to be created freely and also handles metadata from different websites that may use different namespaces, but this problem can be solved by abbreviating popular namespaces only e.g., `p:` will only mean `parody:` and not a custom namespace like `posted:`.

The EH metadata plugin I wrote for my script retrieves additional timestamp and uploader metadata and I have been assuming that eventually LANraragi can use the abbreviations below:
```
a - artist
c - character
f - female
g - group
m - male
l - language
r - reclass
p - parody

u - uploader
t - timestamp ("posted" in JSON returned by EH API)
i - imported (replacing "date_added" from the DateAdded metadata plugin)
d - division (replacing the default "category" namespace)
```
This way probably all namespaces can be abbreviated, but of course this is just my wishful thinking.

Difegue:
I never thought about this before but since I'm going through old issues right now: What if a user actually creates custom namespaces that match those shorthands, like `a:thing, p:other_thing`? 

As it is it wouldn't be too hard to add specific checks for those shorthands in the Search engine and replace them, but I feel the solution would be kinda limited and not very useful?

Moving to a discussion. (I'm loving this feature already)

PR #303, opened 2020-05-13
Add new metadata plugin for Mayriad's EH Master Script

Mayriad:
I am submitting the LANraragi metadata plugin I made for my own userscript. Let me know if you think I should add a test for this plugin to `plugins.t` as well. I am quoting the feature description from my repository to let you know what it does:

> This plugin uses the gallery ID and token appended by the relevant option under the automated gallery downloads feature, and generally only applies to archives downloaded using this feature with this required option enabled. Essentially it works with archives that have identifiers in the format of `[GID * GT *]` in their filenames. Note that this required option cannot work on some browsers, because they do not respect the filename specified in userscripts. It should at least work on Firefox, while Chromium browsers will not support it.
>
> This MEMS (acronym for Mayriad's EH Master Script) plugin generally only works with archives downloaded by this script, but is otherwise superior to the E-Hentai LANraragi metadata plugin because the latter relies on title and reverse image search methods, which are not always accurate and cannot find metadata for deleted galleries.

At the moment, this plugin mostly follows the same standardjs coding convention as the userscript it works with. I can try to restyle it to match the other LANraragi scripts after it is proven to be stable. Of course, you can restyle it yourself the next time you run PerlTidy.

Lastly, my apologises for the E-Hentai plugin last time. I looks like my pull request was problematic after all.


Asinin3:
Slightly off topic, but isn't there too many meta-data formats for galleries already? It seems silly to have so many formats for saving meta-data, when down the line people will have messy galleries with metadata in many different formats and then will require systems that support them all.

Why not make your script use a tried and tested format like the jsons from dnsev's /x/ or eze scripts that has been around for nearly a decade? Or one of the many other formats?

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,161 @@
...
+  # Send the API request.
+  my $response = $user_agent->post(
+    $api => json => {
+      method => 'gdata',
+      gidlist => [[$gallery_id, $gallery_token]],
+      namespace => 1
+    }
+  )->result;
+
+  # Check for temporary ban and warning messages.
```
Difegue, line 116: There's no rate limits on the API that I know of unlike the rest of the site, so this check/warning is probably unnecessary here.

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,161 @@
...
+      . 'EQZm/xEGZv8RBmb/////AP///wARBmb/EQZm/xEGZv8RBmb/////ABEGZv8RBmb/////ABEGZv8RBmb/EQZm/xEGZv8RBmb/EQZm/////wD///'
+      . '8AEQZm/xEGZv////8A////AP///wD///8A////AP///wARBmb/EQZm/////wD///8AEQZm/xEGZv////8A////ABEGZv8RBmb/////AP///wD/'
+      . '//8A////AP///wD///8AEQZm/xEGZv////8A////ABEGZv8RBmb/////AP///wARBmb/EQZm/////wD///8A////AP///wD///8A////ABEGZv'
+      . '8RBmb/////AP///wARBmb/EQZm/////wD///8AEQZm/xEGZv8RBmb/EQZm/xEGZv////8A////AP///wARBmb/EQZm/////wD///8AEQZm/xEG'
+      . 'Zv////8A////ABEGZv8RBmb/EQZm/xEGZv8RBmb/////AP///wD///8AEQZm/xEGZv////8A////ABEGZv8RBmb/////AP///wD///8A////AP'
+      . '///wD///8A////AP///wD///8A////AP///wD///8A////AP///wD///8A////AP///wD///8A////AP///wD///8A////AP///wD///8A////'
+      . 'AP///wD///8A////AP///wD///8A////AP///wD///8A//8AAP//AACDmQAAg5kAAJ+ZAACfmQAAn5kAAISBAACEgQAAn5kAAJ+ZAACfmQAAg5'
+      . 'kAAIOZAAD//wAA//8AAA==',
+    # Custom arguments:
+    oneshot_arg => 'Enter a valid EH gallery URL to copy metadata from this EH gallery to this LANraragi archvie',
```
Difegue, line 37: This duplicates functionality already in the stock E-H plugin (see review comment)

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,161 @@
...
+  my $logger = get_logger('Mayriad\'s EH Master Script', 'plugins');
+  my $gallery_id = '';
+  my $gallery_token = '';
+
+  # Use the URL from oneshot parameters or source tag first when applicable.
+  if ($lrr_info->{oneshot_param} =~ /e(?:x|-)hentai\.org\/g\/(\d+)\/([0-9a-z]+)/i) {
+    $gallery_id = $1;
+    $gallery_token = $2;
+    $logger->debug("Directly using gallery ID $gallery_id and token $gallery_token from oneshot parameters.");
+  } elsif ($lrr_info->{existing_tags} =~ /source:e(?:x|-)hentai\.org\/g\/(\d+)\/([0-9a-z]+)/i) {
```
Difegue, line 61: Ditto

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,161 @@
...
+use LANraragi::Utils::Logging qw(get_logger);
+
+# Meta-information about the plugin.
+sub plugin_info {
+  return (
+    # Standard metadata:
+    name => 'Mayriad\'s EH Master Script',
+    type => 'metadata',
+    namespace => 'memsplugin',
+    login_from => 'ehlogin',
```
Difegue, line 16: I was pretty happy to see another plugin depending on ehlogin at first, but it doesn't look like you actually need a useragent authentified to E-H at all since you're only consuming the API.

Difegue: Of course, if you rework the plugin to call the stock EH one, you'll need this so you can forward a `$lrr_info` object with the proper information

Mayriad:
> Slightly off topic, but isn't there too many meta-data formats for galleries already? It seems silly to have so many formats for saving meta-data, when down the line people will have messy galleries with metadata in many different formats and then will require systems that support them all.
> 
> Why not make your script use a tried and tested format like the jsons from dnsev's /x/ or eze scripts that has been around for nearly a decade? Or one of the many other formats?

I am not sure what you are talking about. My userscript only inserts `[GID * GT *]` into the filename.
- If you are talking about this kind of filename identifier, please give me examples of other existing formats so that I can consider and probably use them.
- If you are talking about JSON, then there could be a misunderstanding because my userscript does not download JSON.
- If you are talking about the metadata this plugin adds in LANraragi, the only thing this plugin does differently from others is that it says `timestamp:` instead of `posted:` (used by EH API) or `date_uploaded:` (used by dnsev-h), etc. I do not mind changing this at all; I just thought using `timestamp:` miiiiight give it its own namespace shorthand in the distant future (#279). By the way, I am also responsible for making the E-Hentai plugin use `timestamp:`, since I basically copied the code from this MEMS plugin to the E-Hentai plugin. We can rename it in both plugins if needed.

> I get that this plugin adds a few extra tags and fits your workflow better (`timestamp:` and `uploader:`), but otherwise it's too similar to add as a stock plugin in its current state.

It does not add extra stuff anymore after I copied them to the stock plugin, which now makes it too similar to the stock plugin.

> The rest of the features(one-shot url override, source tag recognition) are essentially duplicated from the stock E-H plugin, which I think would confuse users when confronted with both plugins out of the box.
> ...
> If you want this plugin merged I'd like it if its capabilities were reduced so that it's a companion to the original stock plugin and not a copy/replacement.

Err, I did want to keep this plugin fully standalone. I found it cleaner and slightly more convenient to delete plugins I do not use, so dependencies need to be avoided. For example, I only use this plugin so I removed everything else and now I never need to select plugin again. Secondly, I think it is clear enough that this plugin is for "Mayriad's EH Master Script", so I did not think it would confuse users.

If you are still against my plugin in its current form, then I will take option two, which I also think is the cleanest solution:
> * Call the stock EH plugin directly within your plugin with the gID/gToken you have, and forward the tags it gives you to your final return - I think this is the cleanest solution.

I can then maintain a standalone version in my userscript repository for myself and any interested users of my userscript.

lib/LANraragi/Plugin/Metadata/MEMS.pm
Mayriad: The EH wiki says the API has a load limit ("25 entries per request, 4-5 sequential requests usually okay before having to wait for ~5 seconds") and an EH mod mentioned API related bans to me once, and the stock plugin also has these warnings, so I added them here.

However, I did not get banned while testing this plugin with 0 delay, so maybe the way LANraragi sends requests one at a time is safe from bans, or maybe there is no real limit? I do not mind removing these, and we can remove them from the stock plugin as well.

lib/LANraragi/Plugin/Metadata/MEMS.pm
Mayriad: Oh right, the login plugin is no longer needed. Back then I was going to add an option to check if a gallery has been removed and tag it accordingly, and that would require the login.

Difegue:
> Err, I did want to keep this plugin fully standalone. I found it cleaner and slightly more convenient to delete plugins I do not use, so dependencies need to be avoided. For example, I only use this plugin so I removed everything else and now I never need to select plugin again. Secondly, I think it is clear enough that this plugin is for "Mayriad's EH Master Script", so I did not think it would confuse users.  

Hm, I advise against doing this since the stock plugins are reinstalled on each update of the app really, although I suppose this would be an argument in favor of a proper plugin repository with auto-updates and the like. Far too much work considering how little plugins there are yet.

Confusing users is only one part of it - If a plugin ends up in mainline I'm usually the one maintaining it, so code duplication should be highly avoided whenever possible. 
Whenever tenboro fucks up e*hentai access again I'm gonna have to update both the stock plugin and this one and that's just extra work ! 😛  

I'd forgotten we put `timestamp` and the like in the stock plugin already too, which makes my above point all the more important. 😅 I still recommend you go with option 2 if you want this merged. There's of course no problems with maintaining a standalone version on your side for users to sideload. 👍

lib/LANraragi/Plugin/Metadata/MEMS.pm
Difegue: Didn't think we had those in place for the API calls as well in the stock plugin -- I guess they can stay but if you're going for option 2 this point is moot.

Asinin3:
> I am not sure what you are talking about. My userscript only inserts [GID * GT *] into the filename.

If that's the case then ignore what I said, it sounded like you had a whole new meta-data format.

Mayriad:
> Confusing users is only one part of it - If a plugin ends up in mainline I'm usually the one maintaining it, so code/feature duplication should be highly avoided whenever possible.

Oh, we had opposite ideas here: I thought I would be the one to maintain it since I am the author, but you thought you would probably end up doing that, so calling the stock plugin makes it more difficult for me to control the plugin but easier for you to maintain.

I have done option 2 and also restyled the code. Let me know what else needs to be done.

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,99 @@
+package LANraragi::Plugin::Metadata::MEMS;
```
Difegue, line 1: Add this namespace to the test in `modules.t` so it can be checked during CI.  
I don't think this plugin needs other specific testing otherwise.

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,99 @@
...
+        # successfully retrieved; otherwise $metadata{tags} may only contain this source tag and truly untagged
+        # galleries may be incorrectly hidden.
+        my $host         = ( $use_exhentai ? 'exhentai.org' : 'e-hentai.org' );
+        $metadata{tags} .= ", source:$host/g/$gallery_id/$gallery_token";
+        # Title is always updated to hide the identifiers and also to reflect title changes due to rename petitions.
+        $metadata{title} = $eh_title;
+    }
+
+    # Return a hash containing the new metadata to be added to LRR.
+    return %metadata;
```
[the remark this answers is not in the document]
Mayriad: It is good. The plugin used to check `if ( $eh_all_tags ne "" )` and give an error message in another function before it is replaced by the stock plugin call. I did not think that check was very necessary though, because EH API would have returned an error first, which would have already given an error in LANraragi.

By the way, do you not also need a `$logger->info("No matching EH Gallery Found!");` here?

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,99 @@
+package LANraragi::Plugin::Metadata::MEMS;
+
+use strict;
+use warnings;
+use Mojo::UserAgent;
+use Mojo::JSON qw(decode_json encode_json);
```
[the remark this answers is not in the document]
Mayriad: Yes, thanks for spotting.

lib/LANraragi/Plugin/Metadata/MEMS.pm
```diff
@@ -0,0 +1,98 @@
...
+        my %metadata = ( tags => $eh_all_tags, title => $eh_title );
+        # Add the source tag outside get_tags_from_EH(), so that this tag is only added when metadata has been
+        # successfully retrieved; otherwise $metadata{tags} may only contain this source tag and truly untagged
+        # galleries may be incorrectly hidden.
+        my $host         = ( $use_exhentai ? 'exhentai.org' : 'e-hentai.org' );
+        $metadata{tags} .= ", source:$host/g/$gallery_id/$gallery_token";
+        # Return a hash containing the new metadata to be added to LRR.
+        return %metadata;
+    } else {
+        return ( error => "No matching EH Gallery Found!" );
```
Mayriad, line 93: Do you not also need a `$logger->info("No matching EH Gallery Found!");` here?

Difegue: Indeed, that'd be better.

tests/modules.t
```diff
@@ -46,8 +46,9 @@ my @modules = (
     "LANraragi::Plugin::Metadata::Chaika",    "LANraragi::Plugin::Metadata::CopyTags",
     "LANraragi::Plugin::Metadata::DateAdded", "LANraragi::Plugin::Metadata::EHentai",
     "LANraragi::Plugin::Metadata::Eze",       "LANraragi::Plugin::Metadata::Hdoujin",
-    "LANraragi::Plugin::Metadata::Koromo",    "LANraragi::Plugin::Metadata::nHentai",
-    "LANraragi::Plugin::Login::EHentai",      "LANraragi::Plugin::Scripts::SourceFinder"
+    "LANraragi::Plugin::Metadata::Koromo",    "LANraragi::Plugin::Metadata::MEMS",
```
Difegue, line 49: Needs to update the `use Test::More tests => 44;`, each module is a test so that brings us to 45.

Difegue:
Alright, I think this is good. Congrats!

Mayriad:
Welp, thanks for looking after me so far. I feel I will need a much better understanding of the internal workings of LANraragi before the next time I make a contribution. I will always maintain this plugin, though, because I consider it part of my userscript.

lib/LANraragi/Plugin/Metadata/MEMS.pm
Mayriad: It seems the API does have a load limit. See [this forum post](https://forums.e-hentai.org/index.php?s=&showtopic=238939&view=findpost&p=5694807).

PR #310, opened 2020-05-19
fix s6 uid/gid dropping

koma-cute:
We're using `#!/bin/sh` now because we need shell features like `export` and command substitution. For some reason, s6-envuidgid can take a colon separated user/group pair, but [s6-setuidgid can't](https://skarnet.org/software/s6/s6-setuidgid.html) (though it can take colon separated uid/gids, or just the username, for some reason). An alternative might be to use the LRR_UID/GID envvars directly, but I think it's better to continue using the koyomi mapping after it's been set once in `cont-init.d`.

### DIFF
--- a/tools/build/docker/s6/services.d/lanraragi/run
+++ b/tools/build/docker/s6/services.d/lanraragi/run
@@ -1,4 +1,4 @@
 #!/bin/sh
 cd /home/koyomi/lanraragi/
 export HOME=/home/koyomi
-s6-setuidgid $(id -u koyomi):$(id -g koyomi) perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
+s6-setuidgid koyomi perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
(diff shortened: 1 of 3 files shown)


tools/build/docker/s6/services.d/lanraragi/run
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/execlineb -P
-s6-envuidgid -B koyomi:koyomi
+#!/bin/sh
 cd /home/koyomi/lanraragi/
-perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
+export HOME=/home/koyomi
+s6-setuidgid $(id -u koyomi):$(id -g koyomi) perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
```
Difegue, line 4: According to the doc link you gave, wouldn't it be better to directly use `s6-setuidgid koyomi` instead of grabbing stuff with `id` ?

koma-cute: Reading the docs I assumed it would only update the UID, not the GID, but testing it out that seems not to be the case. Updated.

Difegue:
Perfect, thanks!

PR #362, opened 2020-11-12
Fix category filtering

pastalian:
Fix a bug that prevented filtering when there is only one archive in a category.

I haven't even looked at perl before, so it may not be the right fix. Feel free to close if it's not.


Difegue:
Indeed, that seems to fix it. Didn't even know that was an existing bug 👀💦
Thanks a lot!

PR #364, opened 2020-11-16
Only change file permissions for the owner in Docker init

koma-cute:
LRR keeps removing perms from my group on its bind mount and it's p'in me the f o. We only chown a user, so don't try and force permissions for stuff we're not interested in.


Difegue:
Looks good to me!

PR #370, opened 2020-12-09
Allow non-unique id to LRR_UID/GID

pastalian:
This is necessary when you want to
- run as root (LRR_UID/GID=0)
- run as the same uid as the host when using a rootless container


Difegue:
LGTM - I never had issues running with UID/GID set to 0 without the `-o` flag, but maybe it's needed by some linux/container flavors?

pastalian:
Oh really? I'm using Gentoo Linux with rootless podman. Rootless containers still have a lot of problems, which may be the reason.
Anyway, thanks for merging :)

PR #372, opened 2020-12-11
Hide overflow tags

pastalian:
Long tags in the tooltip are not displayed nicely.
This PR adds `text-overflow` to the tag tooltip so it looks nicer (I think)


Difegue:
Looks good - I took a more indepth look and have a few more fixes queued up after this PR as a result, so better get mergin'.

PR #383, opened 2020-12-28
Delete the thumbnail after the archive file is deleted

pastalian:
Currently, even if you delete the archive, the thumbnail remains. This should also be deleted.


Difegue:
Looks fine. Thanks!

PR #451, opened 2021-04-16
Fix file permissions

nonamethanks:
A number of static files had executable permissions. This commit fixes that.


Difegue:
The CI seems to be failing only on one test in the Docker environment only, so I'll be abusing my authority for the time being and validate PRs regardless of it. 😇  
This is just perm fixes so, good to go for me!

Issue #299, opened 2020-05-08
[Suggestion] Add a next/previous gallery button


Asinin3:
Reading a series with LRR which is in many seperate galleries/chapters could be made a lot easier if we had a button to simply go to the next alphabetically sorted gallery.

Preferably this would be using the search tags/terms you were using on the index page, so it would go to the next/previous gallery you saw when you were on the index page. But that isn't **too** important if it can't be done. The main reason for this will be for galleries with the same name but are numbered, so those would work regardless.

Salvora:
adding a clickable link to the "reader view" for easily returning the library would be appreciated.

Difegue:
The downward arrow at the bottom of the page does that 👍  
![image](https://user-images.githubusercontent.com/8237712/134817669-cddbf74f-27cd-468f-8c0c-64a01201672a.png)

Salvora:
Oh, I missed that since I tried the infinite scrolling. It is kinda hard to navigate there. Is it possible to put it next to this place.
![image](https://user-images.githubusercontent.com/6360098/134818201-d5f5ed28-3d9b-4a21-9452-8e92cb77e663.png)

psilabs-dev:
PR #1214 (read previous/next) aims to tackle the third one. It has more generalized functionality than alphabetical sorting, since it uses the underlying datatables result to form a reading list.

groutoutlook:
> > [ ]  Add a button/keybind to go to the next/previous alphabetically sorted gallery. Useful when reading series of chapters.
> 
> PR [#1214](https://github.com/Difegue/LANraragi/pull/1214) (read previous/next) aims to tackle the third one.


hi... there is one thing bugging me a bit on this PR..

this doesnt help if I edit the metadata of series name and chapter index (as in the tag) no?

psilabs-dev:
> this doesnt help if I edit the metadata of series name and chapter index (as in the tag) no?

it doesn't cover navigation *within* an archive.

If you have a series with multiple archives and each archive has a chapter, you can search by "series:xxx" and then sort by chapter in ascending order, then navigation is supported.

groutoutlook:
> If you have a series with multiple archives and each archive has a chapter, you can search by "series:xxx" and then sort by chapter in ascending order, then navigation is supported.

I sort by "Title" since that's one of default working right on the Home UI...and your PR work fine.

bit off-topic but I did add something like "chapter" tag though, and that couldnt be sorted somehow. 

[...]

PR #452, opened 2021-04-17
Reader: add some keybinds

nonamethanks:
As discussed on discord
* add a/d for single hand navigation
* add backspace to go back to index
* change overlay key from CTRL to q to avoid interfering with browser and system shortcuts

Some notes:
* event.keyCode is supposed to be deprecated, but AFAIK some old browsers like internet explorer and very old versions of firefox don't have support for the much more readable alternative ([event.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), which outputs simple strings like "RightArrow" or "d"). For now it's better to go for the easiest option and just build on the existing system, you never know someone is browsing lanraragi through a nintendo DS browser
* there are several bugs with the existing implementation, one notable example is the fact that spacebar with right-to-left browsing navigates backwards pagewise. There are other minor issues such as the pages still being scrollable while the overlay is open
* a lot of these functions are referenced in the template code - this makes it harder for example to rename or shuffle around functions, since there's multiple points that need to be modified every time a change needs to be made. 

If it's ok with you I plan to fix some of the LANraragi javascript modules to bring them up to date with modern JS techniques. Namely, a single (exported) object per file with various methods in it, and a single initializing function that binds each of these methods to applicable DOM elements, rather than having the function names directly in the html.
Think something like `Reader.update_tags()`, `Reader.remove_category()`, `Reader.next_page()`, `Archive.open_source_in_new_tab()` (for the index thumbnails) etc.
This would make it much easier then to implement neat things like dynamically changing single-double page mode without refreshing/changing pages, or adding an infinite scroll reader, or provide utility functions to userscripts such as to edit tags or use custom keybinds without having to deal with the API directly.
It would also force every action of the same kind to be called from a single point, making it much easier to debug issues or expand functionality.

(I don't want to do too much at the same time though, don't worry. I'll do small, easily reviewable bits at a time).

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -2,24 +2,27 @@
 //also handles the thumbnail archive explorer.
 
 function moveSomething(e) {
-
+    if (e.target.tagName === "INPUT") {
+        return;
+    }
 	switch (e.keyCode) {
-		case 37:
-			// left key pressed
+		case 8:   // backspace
+			returnToIndex();
+			break;
+		case 37:  // left arrow
+		case 65:  // a
 			advancePage(-1);
 			break;
-		case 32:
-			// spacebar pressed
+		case 39:  // right arrow
+		case 68:  // d
+			advancePage(1);
+			break;
+		case 32:  // spacebar
 			if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
 				advancePage(1);
 			}
 			break;
-		case 39:
-			// right key pressed
-			advancePage(1);
-			break;
-		case 17:
-			// Ctrl key pressed
+		case 81:  // q
 			openOverlay();
 			break;
 	}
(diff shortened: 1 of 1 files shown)


nonamethanks:
While I was testing this further I discovered another existing bug: even when focusing on input text (such as in the "Container Width" option), shortcuts are still passed along and activated. 

To prevent this from happening I added a hack in the form of 
```   js
    if (e.target.tagName === "INPUT") {
        return;
    }
```
That prevents shortcuts from firing in case a text field is focused.

public/js/reader.js
```diff
@@ -2,24 +2,27 @@
...
 			break;
-		case 32:
-			// spacebar pressed
+		case 39:  // right arrow
+		case 68:  // d
+			advancePage(1);
+			break;
+		case 32:  // spacebar
 			if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
 				advancePage(1);
```
Difegue, line 22: Might as well fix the spacebar issue while we're at it. 👍 
It's a bit cheap to just check readorder and negate the value a second time, but I don't think it's worth refactoring advancePage over.
```suggestion
(localStorage.readorder === 'true') ? advancePage(-1) : advancePage(1);			
```

Difegue:
Looks good, thanks a lot ! 🎉 The keybind code is very old™, I can confirm LRR did work on windows mobile 6.x back in 2015.   

Using a plugin for the hotkeys indeed doesn't seem worth it considering how little of them are there -- It might be worthwhile if it came with support for rebinding so I could cross that off the #299 list.  
(I think you could check for the modal being opened by looking at the css classes? That's likely a bit too specific to write and not worth the trouble.)

I very much welcome incremental rewrites/improvements to the JS portion if it can make further contributions easier -- It'd also force me to understand how modern JS works so that's not necessarily a bad thing. 😁  
Having a global object/namespace per page would also make autocompletion way easier when debugging 🙏

Difegue (approved):
woops the tabulation was wrong on my suggestion - this can be fixed later tho, no biggie.

PR #453, opened 2021-04-17
Add eslint configuration

nonamethanks:
As discussed on discord, adds a basic eslintrc following [AirBNB](https://github.com/airbnb/javascript), with double quoted strings and four spaces indentation.

`func-names": ["error", "never"]` [is needed](https://eslint.org/docs/rules/func-names#never) so that it's possible to use syntax of the type:
```
Reader.toggle_overlay = function () {
}
```

which is the standard for exporting namespaces. (This does not affect normal function definitions, just functions launched on creation and functions bound to objects such as the above example)

The ["no-unused-expressions"](https://eslint.org/docs/rules/no-unused-expressions) options are there to allow ternary operations with functions without having to assign the result.

Also added the "allowForLoopAfterthoughts" option since I noticed you use var++ in for loops.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,27 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "jquery": true
+    },
+    "extends": [
+        "airbnb-base"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 12,
+        "sourceType": "module"
+    },
+    "rules": {
+        "indent": ["error", 4],
+        "quotes": ["error", "double"],
+        "func-names": ["error", "never"],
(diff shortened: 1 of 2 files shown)


Difegue:
Gave this a spin and it looks fine, I'll be pushing changes to the vscode default dev config in .devcontainer to add the eslint extension as an auto formatter. 🎉  

Adding eslint into devDependencies means it'll get downloaded/installed by `npm run lanraragi-installer`, which I think is fine since most users just use the Docker containers instead of source installs, and those yeet node_modules away.  

If someone complains I'll just edit the install script to do `npm --only=production install` or something. ¯\\\_(ツ)_/¯

PR #458, opened 2021-04-28
Reader: add preload option

nonamethanks:
Makes preload size customizable (only works in normal browsing, not infinite scrolling) and properly caches preloaded images.

There's a weird behavior with Chrome that makes it ignore preloading in certain cases and only in single page mode: I could only reproduce it on page reload when going backwards once, then fowards once. Only those two images are reloaded. I'm thinking it's something to do with how chrome handles cache, because I can't reproduce it with firefox, but I could only discover that it was tied to the HEAD requests for the filesize. Caching those stopped it from happening 90% of the time, and given that it requires specific actions to happen, I decided to open this anyway.

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -4,7 +4,8 @@
 const Reader = {};
 Reader.previousPage = -1;
 Reader.showingSinglePage = true;
-Reader.imagesLoaded = 0;
+Reader.preloadedImg = {};
+Reader.preloadedSizes = {};
 
 Reader.initializeAll = function () {
     Reader.maxPage = Reader.pages.length - 1;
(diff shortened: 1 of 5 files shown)


public/js/reader.js
```diff
@@ -323,30 +337,22 @@ Reader.goToPage = function (page, fromHistory = false) {
...
+        const loadHandler = () => { (imagesLoaded += 1) === 2 && Reader.drawCanvas(img1, img2); };
+        $([img1, img2]).each((_i, img) => {
+            img.onload = loadHandler;
+            // If the image is preloaded it does not trigger onload, so we have to call it manually
+            if (img.complete) { loadHandler(); }
+        });
     } else {
-        // In single view, just use the source URLs as is
-        $("#img").attr("src", Reader.pages[Reader.currentPage]);
+        const preloaded = Reader.loadImage(Reader.currentPage);
```
Difegue, line 350: We don't really need to know the image is preloaded in this portion of the code so I'd just rename the const to `img` here

Difegue (changes requested):

Difegue (approved):

PR #460, opened 2021-04-28
Reader: fix hiding header not making image reach 98vh

nonamethanks:
Reported on discord. Fixes an issue caused by a typo (Reader.hideHeader was actually meant to be localStorage.hideHeader! Embarassing.)

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -408,7 +408,8 @@ Reader.applyContainerWidth = function () {
         // Fit to height forces the image to 90% of visible screen height.
         // If the header is hidden, or if we're in infinite scrolling, then the image
         // can take up to 98% of visible screen height because there's more free space
-        $(".reader-image").attr("style", `max-height: ${Reader.hideHeader || Reader.infiniteScroll ? 98 : 90}vh;`);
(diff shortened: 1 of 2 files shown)


public/js/reader.js
```diff
@@ -447,6 +448,7 @@ Reader.toggleHeader = function () {
     localStorage.hideHeader = $("#i2").is(":visible");
     $("#toggle-header input").toggleClass("toggled");
     $("#i2").toggle();
+    Reader.applyContainerWidth();
```
Difegue, line 451: This seems to break double-page view: The loading spinner never hides even though the pages are fully loaded.

Difegue (changes requested):

public/js/reader.js
nonamethanks: Yeah sorry for the mess, that was fixed in pull #458 lol I discovered it afterwards but it would've caused a conflict with the preload pull, so I just added the fix there.

Difegue: Ah, that's on me for taking them out of order. I'll go check #458 and come back to this later I guess

Difegue: Saw the fix in the other PR so good to go here 👍

PR #465, opened 2021-05-04
Fix eh downloader autostart behavior

koma-cute:
Prior to this, any of the "X Select, Auto Start" options in the uconfig page would cause the download to fail with a mysterious "Unsupported File Extension" error. The issue stems from the ?autostart=1 query param not being replaced with ?start=1. No query leads to a webpage with the URL, ?autostart=1 is the same thing but with a js trigger to redirect to ?start=1, which in turn is the real file.

### DIFF
--- a/lib/LANraragi/Plugin/Download/EHentai.pm
+++ b/lib/LANraragi/Plugin/Download/EHentai.pm
@@ -96,11 +96,11 @@ sub provide_url {
     $content = $response->body;
     $logger->debug("/archiver.php result: $content");
 
-    my $finalURL = "";
+    my $finalURL = URI->new();
     eval {
         # Parse that to get the final URL
         if ( $content =~ /.*document.location = "(.*)".*/gim ) {
-            $finalURL = $1;
+            $finalURL = URI->new($1);
             $logger->info("Final URL obtained: $finalURL");
         }
     };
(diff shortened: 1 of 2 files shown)


lib/LANraragi/Plugin/Download/EHentai.pm
```diff
@@ -109,10 +110,8 @@ sub provide_url {
         return ( error => "Couldn't proceed with an original size download: <pre>$content</pre>" );
     }
 
-    # If the URL doesn't already have autostart, append start=1 to get an URL that automatically triggers the download.
-    unless ( $finalURL =~ /.*autostart=1.*/gi ) {
-        $finalURL .= "?start=1";
-    }
+    # Append start=1 to get an URL that automatically triggers the download.
```
Difegue, line 113: It's not really an append anymore since we're just replacing the query parameters directly now, would be good to edit the comment to reflect that.
```suggestion
    # Set URL query parameters to ?start=1 to automatically trigger the download.
```

lib/LANraragi/Plugin/Download/EHentai.pm
```diff
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 no warnings 'uninitialized';
 
+use URI;
```
Difegue, line 7: We don't really install the `URI` module as a dependency in the cpanfile, it's brought in by our dependency on `URI::Escape`.
It seems `URI::Escape` was basically replaced/included in the full URI pod if https://metacpan.org/pod/URI is any indication, so the matching line in the cpanfile should be modified: https://github.com/Difegue/LANraragi/blob/8e2114bc77e3adff0d4cf1cc4adf30b8198ac1b5/tools/cpanfile#L19 
`requires 'URI', 5.09;`

Difegue (changes requested):
Replacing the query parameter outright does seem to be a better approach. I have a few nitpicks on this one but nothing too major 👍

lib/LANraragi/Plugin/Download/EHentai.pm
koma-cute: honestly thought it was part of the stdlib and thought nothing of it

PR #467, opened 2021-05-04
retry download_url attempts up to 5 times

koma-cute:
perhaps make this user configurable later
this function can fail easily with timeouts if queuing many downloads at once


Difegue:
I haven't encountered the issue on my own usage of the downloaders, but I suppose this can't hurt considering the cost.  
The `attempts` parameters is implemented in Minion::Backend::Redis so it should all be peachy!

PR #470, opened 2021-05-08
Ignore namespace if its value is "other"

### LINKED ISSUE #446: Clicking to search a tag with no namespace searches in namespace "other:"
Running 0.7.8 using docker on a raspberry pi
Using the newly added feature to click on a tag to search for it seems to default to the namespace "other:" if it isn't associated with a namespace. This returns no results, as the tag has no namespace!

firefox_Kgvc249WD8

Additionally, while this already has an issue open, searching a tag in the "male:" namespace also returns results from the "female:" namespace.

pastalian:
Fixes #446

### DIFF
--- a/public/js/common.js
+++ b/public/js/common.js
@@ -106,9 +106,11 @@ function buildTagsDiv(tags) {
         line += `<tr><td class='caption-namespace ${encodedK}-tag'>${ucKey}:</td><td>`;
 
         tagsByNamespace[key].forEach(function (tag) {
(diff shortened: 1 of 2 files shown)


Difegue:
`createNamespaceColumn` handles the creation of the <a> links for columns that show a dedicated namespace in compact mode, like so:  
image  

I believe it's fine to leave the namespace in the search field for those, since most users load their columns with namespaced tags anyway.

Difegue (approved):
Looks and works fine, thank you! 🚀

PR #485, opened 2021-07-07
looks for language inside HTML

### LINKED ISSUE #422: Japanese language "tag"
Sadpanda doesn't allow tagging japanese galleries as such with the exception of a few cases. It does however have a, I'll just call it _secondary language tag_.

japanese

I think it would be helpful if metadata plugins would recognize if no _normal language tag_ exists and then look for this _secondary language tag_ instead.

IceBreeze:
closes #422

Also changed the test system a bit.
Not sure if you like it so I contained myself to work only on the target.


tests/plugins.t
```diff
@@ -53,13 +53,10 @@ my ( $test_eH_gID, $test_eH_gToken ) =
...
-my $eH_tags =
-  "parody:touhou project, character:hong meiling, character:marisa kirisame, character:reimu hakurei, character:sanae kochiya, character:youmu konpaku, group:handful happiness, artist:nanahara fuyuki, artbook, full color, category:non-h";
-
-my ( $test_eH_tags, $test_eH_title ) = LANraragi::Plugin::Metadata::EHentai::get_tags_from_EH( $ua, $eH_gID, $eH_gToken );
-
-is( split( ", ", $test_eH_tags ), split( ", ", $eH_tags ), 'eHentai API Tag retrieval test' );
-is( $test_eH_title,
+my $test_eH_json = LANraragi::Plugin::Metadata::EHentai::get_tags_from_EH( $ua, $eH_gID, $eH_gToken );
+ok( exists $test_eH_json->{gmetadata}, 'gmetadata exists' );
+isa_ok( $test_eH_json->{gmetadata}, 'ARRAY', 'type of gmetadata' );
```
Difegue, line 58: I'm wondering if it'd be worth it to also check `gmetadata[0]{tags}` here with a `cmp_bag`: Checking the title is probably already a pretty good way of making sure the live API still works, so it's up to you if you want to do it. 😀

tests/LANraragi/Plugin/Metadata/EHentai.t
```diff
@@ -0,0 +1,82 @@
...
+use Data::Dumper;
+
+use Cwd qw( abs_path );
+use Mojo::JSON qw(decode_json encode_json);
+use Mojo::File;
+
+use Test::More;
+use Test::Deep;
+
+my ($ROOT) = abs_path(__FILE__) =~ m/(.*)\/tests.+/;
```
Difegue, line 14: Not seeing the point of doing a regex on `__FILE__` here but I might be missing something 😛  
Wouldn't 
```
my $cwd = getcwd;
my $SAMPLES = "$cwd/tests/samples";
```
achieve a similar result?

Difegue (changes requested):
This is great, thank you very much! I have a few comments but it's all minor stuff.  

I don't mind extending the tests with sample data at all so if you fancy doing more of that, no problem on my end. 👍  
I didn't know about the `note` method in Test::More either, so that's something I'll try to use from now on.

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -101,28 +102,44 @@ sub get_tags {
...
     }
 
-    my ( $ehtags, $ehtitle ) = &get_tags_from_EH( $lrr_info->{user_agent}, $gID, $gToken, $jpntitle, $additionaltags );
-    my %hashdata = ( tags => $ehtags );
+    my $json = &get_tags_from_EH( $lrr_info->{user_agent}, $gID, $gToken );
+    my %hashdata = ( tags => "" );
+
+    if ( $json ) {
+        my $tags = &get_tags_from_json( $json, $additionaltags );
+        # only check language on e-hentai
```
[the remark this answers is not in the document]
IceBreeze: I don't know well how Ex works, but I found out that it needs at least to pass a cookie.
But the "oneshot_arg" is taken. How can I pass it?
Should I add a new config param to the plugin? It didn't seem a good place at the moment... and can always be extended.

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -101,28 +102,44 @@ sub get_tags {
...
+        $logger->info("Sending the following tags to LRR: " . $hashdata{tags});
     }
 
     #Return a hash containing the new metadata - it will be integrated in LRR.
     return %hashdata;
 }
 
+sub get_local_logger {
+    my %pi = plugin_info();
+    return get_logger( $pi{name}, "plugins" );
```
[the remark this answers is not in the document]
IceBreeze: I'm too lazy to do copy&paste :)

Anyway I think it's preferable to create a "`Const::Fast`" hash to pass with `plugin_info()`, so it's easier to use inside the module.

But as I said, I tried to contain myself because I don't know what's behind the plugins.

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -248,34 +265,103 @@ sub get_tags_from_EH {
...
+
+    #Execute the request
+    my $rep = $ua->get($uri)->result;
+    if ( $rep->is_error ) {
+        $logger->error( "E-H GET message: " . $rep->message );
+        return;
     }
+
+    return $rep->dom;
+
```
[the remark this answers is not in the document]
IceBreeze: I thought it was symmetric with the space before the arguments :smile:

tests/plugins.t
IceBreeze: I could, but I think this test is better done inside the `EHentai.t` specific module.

`plugin.t` is very similar to an integration test and it calls many modules.
 Adding here a "cmp_bag" test for any plugin used, will result in cluttering the code.
Besides I should have removed also the title because, like tags, it can change any moment breaking the test. Its better to test them using the "more stable" samples. 

Normally in this type of tests I check the response codes or that the needed fields are present. So I can be sure the API haven't changed.

tests/LANraragi/Plugin/Metadata/EHentai.t
IceBreeze: It's an habit of mine: like this it doesn't matter where I am when I call the script.

lib/LANraragi/Plugin/Metadata/EHentai.pm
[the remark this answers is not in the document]
Difegue: Ex cookies are handled by the login plugin system beforehand, so there's nothing to do! 👍 

 The `Mojo::UserAgent` object you use to make requests automatically has an exhentai cookie if the user specified proper credentials in Plugin Configuration.

lib/LANraragi/Plugin/Metadata/EHentai.pm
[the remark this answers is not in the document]
Difegue: I was thinking of eventually making a generic Util method that plugins could call to get a logger by passing their own instance (so we could dynamically call `plugin_info` and get the name there), but the const method could also work! Although it'd bring an extra module dependency. 😁

lib/LANraragi/Plugin/Metadata/EHentai.pm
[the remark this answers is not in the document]
Difegue: I like a space after arguments but not before the closing bracket 😅 I'm aware it's weird

tests/plugins.t
Difegue: Np, fair enough.

tests/LANraragi/Plugin/Metadata/EHentai.t
Difegue: Oh alright I see: I was asking since the tests for `Eze.pm` use `getcwd` and I'd rather keep things consistent. 😌

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -227,13 +244,13 @@ sub ehentai_parse() {
 }
 
 # get_tags_from_EH(userAgent, gID, gToken, jpntitle, additionaltags)
-# Executes an e-hentai API request with the given JSON and returns tags and title.
+# Executes an e-hentai API request with the given gallery info and returns the result JSON.
 sub get_tags_from_EH {
```
Difegue, line 248: The sub name also needs to be changed here (and in `plugins.t` iirc)

Difegue (changes requested):

CirnoT:
Please note that this is technically incorrect. API will NOT return tags below specific threshold of vote even if they are present. All newly uploaded galleries would get wrongly tagged as Japanese by your code EVEN if they contained proper `language` namespaced tag which was not voted enough yet.

Restricting `language:japanese` tag is a very specific edge case on EH and it does not apply to any other language. Hence for simplicity case, you can ALWAYS use `-"language:*"` on LRR search to find Japanese non-translated galleries and `"language:english" -"language:translated"` to find English original languages (replace `english` with your desired language to extend this further).

As such, I believe that this feature is redundant as LRR does not support explicit languages and only relies on tag data which is never consistent across sources.

Of important note is also that this will very likely break sooner or later and LRR has no proper CI set up to perform tests OUTSIDE of actual commit being pushed, so this would not be visible immediately even if explicitly tested for but once commit is made or PR is opened.

IceBreeze:
Thank you @CirnoT for the explanation about the language tag.
So the "language" visible on the left on EH is like a placeholder and will change when the language tag will be upvoted?

Anyhow now I think that this option should be putted behind a parameter and disabled by default to be enabled if someone needs it.

I haven't checked how the search works so far ( is something like `-language:* OR language:japanese` legit? ), but I think that the absence of language doesn't mean it's necessarily "japanese" unless it's true that every gallery is correctly tagged and it's true that every non-tagged gallery is japanese. 
I'm trying to find the needs behind the request here and maybe the user is fine with what they can see on the EH web page.

CirnoT:
> So the "language" visible on the left on EH is like a placeholder and will change when the language tag will be upvoted?

Yes, once the tag is properly voted it will change classification of language. The same applies to some special tags such as `reclass` namespace.

>I haven't checked how the search works so far ( is something like -language:* OR language:japanese legit? )

LRR does not support OR queries, everything is implicitly AND. Additionally, on EH presence of `language:japanese` is allowed if the original work has been translated into Japanese or if the work is dual-language.

> but I think that the absence of language doesn't mean it's necessarily "japanese" unless it's true that every gallery is correctly tagged and it's true that every non-tagged gallery is japanese.

Every gallery defaults to Japanese implicitly and uploader can not change it in any other way than tagging the gallery with `language` namespace tag and having it voted by others. Uploader can only provide initial main title, japanese script title, comment and category. Everything else is done post-upload via tagging in appropriate namespaces. In addition, all imagesets galleries are implicitly Japanese even when there is no text proper, the exception to this is when 50% or more of images in the set contains text and language namespaced tag is added.

> I'm trying to find the needs behind the request here and maybe the user is fine with what they can see on the EH web page.

Not every request deserves to be integrated into product. Sometimes user should be instead told that they can work around their specific case with appropriate tag search.

IceBreeze:
I agree with you @CirnoT , maybe it's not worth doing this.

Difegue:
I wasn't aware that the "Japanese" field on the left was a default and potentially incorrect until a replacement `language:` tag is voted to 100.  

As such, it's probably better to scrap the whole DOM parsing part of this PR. 😥 Sorry to have your work thrown away like that, I should've properly reviewed #422 before marking it as a volunteer issue.  

However, I still think the small refactoring to the metadata plugin with `get_json_from_API`, as well as the new tests with sample data are worth keeping and merging.

IceBreeze:
It's not a problem. I stopped by because I saw it's a project in Perl and Mojolicious that I find somewhat funny to use.
So I picked up something that seemed small and localized to give it a try because I saw we have different style in programming.

But if you think that refactoring the tests could be useful, with some guideline ( like sticking to `getcwd` or not ), it could be done more extensively than that "snippet".

Difegue:
From what I've seen, I don't think the style is that far apart. (although I'd rather stick to `getcwd` 😁) 
I'll happily welcome extra Perl/Mojo contributions are those are hard to get a hand on!  

Refactoring the tests to add some structure to them (sorting per module, putting samples in a dedicated folder, etc) is a welcome addition in my book. 👀  

We can either keep going here to get the small EH fixes and their tests merged, or do a fresh new PR -- Your call 👍

IceBreeze:
It's better a new dedicated PR.

PR #490, opened 2021-07-26
adds utility sub "get_plugin_logger"

IceBreeze:
This is what I meant in the previous PR. It's something that I rarely use, but I don't think it's so strange in Perl :smile:


Difegue (changes requested):
Awesome! I wasn't thinking of using `caller` as well but I think it works quite well.  

Could you just update the docs at https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/plugin-docs/code-examples.md to show the new way of getting a logger object? Aside from that I see nothing wrong here 👍

IceBreeze:
I will and, at this point, maybe I should update also the other plugins before closing this PR.

Difegue (approved):

PR #496, opened 2021-08-28
Add cookie support for fakku

### LINKED ISSUE #475: FAKKU plugin fails to retrieve metadata from pages that contain "Controversial Content" tags
**LRR Version and OS**
Version 0.7.8, "Looking for Lester" (Windows)

**Bug Details**
As the title says, the plugins fails to retrieve any information if the content is behind the "Show Controversial Content" filter.

**Matching Logs**
`Error: Can't call method "child_nodes" on an undefined value at /home/koyomi/lanraragi/script/../lib/LANraragi/Plugin/Metadata/Fakku.pm line 163.`

**Screenshots**
image

Nodja:
This fixes #475.

I never used perl before so hopefully I got everything right.


lib/LANraragi/Plugin/Login/Fakku.pm
```diff
@@ -0,0 +1,60 @@
+package LANraragi::Plugin::Login::Fakku;
```
Difegue, line 1: Please add this package to the list of tested modules in https://github.com/Difegue/LANraragi/blob/dev/tests/modules.t

lib/LANraragi/Plugin/Login/Fakku.pm
```diff
@@ -0,0 +1,60 @@
...
+        ]
+    );
+
+}
+
+sub do_login {
+
+    shift;
+    my ( $fakku_sid ) = @_;
+    return get_user_agent( $fakku_sid );
```
Difegue, line 31: Not sure we really need a spare function here since there's not much going on.

lib/LANraragi/Plugin/Login/Fakku.pm
```diff
@@ -0,0 +1,60 @@
...
+sub plugin_info {
+
+    return (
+        name      => "Fakku",
+        type      => "login",
+        namespace => "fakkulogin",
+        author    => "Nodja",
+        version   => "0.1",
+        description =>
+          "Handles login to fakku. The cookie is only valid for 7 days so don't forget to update it.",
```
Difegue, line 19: Having to update the cookie every 7 days kinda sucks, but I'm guessing the login procedure is too annoying to automate.   
Is `fakku_sid` the only cookie that's needed? I don't actually have a Fakku account so I haven't tested this.

Difegue (changes requested):
Looks fine, I only have a few minor nitpicks. Thanks for tackling this!    
I renamed some stuff in the fakku plugin so you might have to rebase the PR, sorry 🙏

lib/LANraragi/Plugin/Login/Fakku.pm
Nodja: Yeah only fakku_sid is needed, the login form uses the a hidden google recaptcha field so I didn't even bother trying.

Nodja:
I added the module to the tests but had to reshuffle everything to keep the 2 column structure aligned.

The old variable names were funny reference but I agree it's probably not a good idea to keep that kind of naming.

Hopefully I rebased everything properly.

lib/LANraragi/Plugin/Metadata/Fakku.pm
```diff
@@ -24,7 +24,7 @@ sub plugin_info {
         type        => "metadata",
         namespace   => "fakkumetadata",
         author      => "Difegue",
```
Difegue, line 26: Looks like the `loginfrom` field disappeared during the rebase here.

lib/LANraragi/Plugin/Metadata/Fakku.pm
```diff
@@ -24,7 +24,7 @@ sub plugin_info {
...
-        version     => "0.5.1",
+        version     => "0.6",
         description => "Searches FAKKU for tags matching your archive.",
```
Difegue, line 28: Since we now have cookie support, it might be worth it to edit the description to say you'll only get controversial content metadata if you login.  

 ```suggestion
        description => "Searches FAKKU for tags matching your archive. If you have an account, don't forget to enter the matching cookie in the login plugin to be able to access controversial content.",
```

**Review by Difegue — CHANGES_REQUESTED** [review]

File: tests/modules.t [hunk 6]
```diff
@@ -26,34 +26,34 @@ require $cwd . "/tests/mocks.pl";
...
+    "LANraragi::Model::Upload",                             "LANraragi::Plugin::Metadata::Chaika",
+    "LANraragi::Plugin::Metadata::CopyTags",                "LANraragi::Plugin::Metadata::DateAdded",
+    "LANraragi::Plugin::Metadata::EHentai",                 "LANraragi::Plugin::Metadata::Eze",
+    "LANraragi::Plugin::Metadata::Hdoujin",                 "LANraragi::Plugin::Metadata::Koromo",
+    "LANraragi::Plugin::Metadata::MEMS",                    "LANraragi::Plugin::Metadata::nHentai",
+    "LANraragi::Plugin::Metadata::RegexParse",              "LANraragi::Plugin::Metadata::Fakku",
+    "LANraragi::Plugin::Login::EHentai",                    "LANraragi::Plugin::Login::Fakku",
+    "LANraragi::Plugin::Scripts::SourceFinder",             "LANraragi::Plugin::Scripts::FolderToCat",
+    "LANraragi::Plugin::Download::EHentai",                 "LANraragi::Plugin::Download::Chaika",
+    "LANraragi::Plugin::Scripts::nHentaiSourceConverter",   "LANraragi::Plugin::Scripts::Normalizer"
```
Difegue, line 56: The testcount on line 7 needs to be bumped up to 56 😅 I always forget about this too.

lib/LANraragi/Plugin/Metadata/Fakku.pm
Nodja: yep, I probably deleted it by mistake when resolving the conflicts

Difegue:
All clear, I'll be merging this. Thanks again! 🙌

PR #497, opened 2021-08-31
adds tags rules support

IceBreeze:
This adds support for tags rules (disabled by default).

The rules are expanded and stored in Redis when the configuration is saved, so they can be used directly when needed.
I would have preferred to store the expanded array itself, but I have never used Redis so far so I chose to convert it in JSON.

At the moment I have extended the section "Tags and Thumbnails" to allow the configuration of the rules. Blacklist and tag rules can be enabled at the same time and the rules are applied before the blacklist (for no reason: just put them there to see the result).
Anyhow the tag rules can replace the blacklist, but they're separated by "new line" characters instead of commas. It can be changed, but comma separated rules are not easily readable.

Do you think it can be useful?
How would you prefer to manage blacklist and rules? :)

--- 

Mh... I forgot to tell you that the rules are case sensitive at the moment


Difegue:
Thanks a lot for this, it looks awesome! I'll try to give you a code review ASAP. 💯  

I'm all for getting rid of the old blacklist mechanism since tag rules seem to basically be a more powerful superset, but we'll see. 👀

templates/templates_config/config_tags.html.tt2
```diff
@@ -47,4 +47,22 @@
...
+<tr>
+    <td class="option-td">
+        <h2 class="ih"> Tag Rules </h2>
+    </td>
+    <td class="config-td">
+        <input id="tagsruleson" name="tagsruleson" class="fa" type="checkbox" [% IF tagsruleson %]checked[% END %]>
+        <br>
+        <textarea class="stdinput" size="20" style='width:100%; height:75px' name="tagsrules"
+            maxlength='5000'>[% tagsrules %]</textarea>
+        <br>
```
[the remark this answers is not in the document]
Difegue: It might also be good to add a page to the documentation for this feature later:  
https://github.com/Difegue/LANraragi/tree/dev/tools/Documentation/advanced-usage  
(Blacklisting didn't have a page, but then again it was a much simpler feature 😅)

lib/LANraragi/Controller/Config.pm
```diff
@@ -72,7 +77,12 @@ sub save_config {
         devmode       => ( scalar $self->req->param('devmode')       ? '1' : '0' ),
         enableresize  => ( scalar $self->req->param('enableresize')  ? '1' : '0' ),
         blackliston   => ( scalar $self->req->param('blackliston')   ? '1' : '0' ),
-        nofunmode     => ( scalar $self->req->param('nofunmode')     ? '1' : '0' )
+        tagsruleson   => ( scalar $self->req->param('tagsruleson')   ? '1' : '0' ),
+        nofunmode     => ( scalar $self->req->param('nofunmode')     ? '1' : '0' ),
+
+        # For backend only
+        # Expands rules on save so they can be used directly
+        exptagsrules  => encode_json( [ tags_rules_to_array($self->req->param('tagsrules')) ] )
```
Difegue, line 85: Right, I see what you mean now by preferring to store the array in Redis 😁 Storing stuff in ``LRR_CONFIG`` that doesn't directly mirror the configuration entered by the user should be avoided.  

It'd be better to store the computed rule array as a different Redis key directly.  
Redis has rather simple data structures sadly so we can't transpose an array-of-arrays as-is, but I think it's workable with a simple list.  

Here's how this could look like imo(hastily written, probably doesn't run as-is):  

````
# Save
my @array = ("replace_ns","artist","creator","remove","incomplete",null,"strip_ns","test",null);
$redis->del("LRR_TAGRULES"); # Clear previous rules
$redis->lpush("LRR_TAGRULES", reverse @array); # LPUSH always pushes at the head of the list, hence the reverse

# Restore
my @values = $redis->lrange("LRR_TAGRULES", 0,-1);
my @restored = ();

for (my $i=0; $i <= $#values; $i += 3) {
    push @restored, ($values[$i], $values[$i+1], $values[$i+2]);
}
# @restored = ( ("replace_ns","artist","creator"), ("remove","incomplete",null), ("strip_ns","test",null))
````

File: lib/LANraragi/Utils/Tags.pm [hunk 3]
```diff
@@ -0,0 +1,96 @@
...
+}
+
+sub apply_rules {
+    my ( $tag, $rules ) = @_;
+
+    foreach my $rule ( @$rules ) {
+        my $match = $rule->[1];
+        my $value = $rule->[2];
+        given($rule->[0]) {
+            when ('remove')     { return if ( $tag eq $match ); }
```
Difegue, line 85: You can use `uc($tag)` and ` uc($match)`  here to fix the case-sensitivity issue, and for the other cases that rely on regexes the `/i` flag should be enough.

Difegue (changes requested):
Yep, this is great stuff and should make power users quite happy. 🥳 I think this can completely replace the blacklist code in due time. 
My only major gripe is with the way the computed rules are saved to the config hash, the other remarks are minor stuff.  

Feel free to ask questions about Redis if you're having trouble using it 👍

lib/LANraragi/Utils/Tags.pm
```diff
@@ -0,0 +1,96 @@
...
+
+use strict;
+use warnings;
+use utf8;
+use feature "switch";
+no warnings 'experimental';
+
+use LANraragi::Utils::Generic qw(remove_spaces remove_newlines);
+
+# Generic Utility Functions.
```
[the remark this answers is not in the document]
IceBreeze: I'll go with a more generic "Tag system": it also include the function to split tags and can be extended with other tag related stuff.

lib/LANraragi/Utils/Generic.pm
```diff
@@ -281,4 +281,16 @@ sub css_default_names {
     }
 }
 
+sub replace_CRLF {
```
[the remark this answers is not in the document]
IceBreeze: OK, though I don't think it's a good place either :thinking:

In `Utils/Generic` there are many subs related to string manipulation, that's the only reason why I put them there. Maybe they are enough to create a `Utils/String` package.
Anyway I'll move those two to `Utils/Tags`.

templates/templates_config/config_tags.html.tt2
```diff
@@ -47,4 +47,22 @@
...
+</tr>
+
+<tr>
+    <td class="option-td">
+        <h2 class="ih"> Tag Rules </h2>
+    </td>
+    <td class="config-td">
+        <input id="tagsruleson" name="tagsruleson" class="fa" type="checkbox" [% IF tagsruleson %]checked[% END %]>
+        <br>
+        <textarea class="stdinput" size="20" style='width:100%; height:75px' name="tagsrules"
```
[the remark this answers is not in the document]
IceBreeze: Good for me... I prefer to leave the graphic details to you :grin:

lib/LANraragi/Utils/Tags.pm
IceBreeze: I was wondering how to express choice using the rules, but I'm fine to go with the "case-insesitivity" :grin:

IceBreeze: I'll go with the `lc` function so I can store the computed array in lowercase. I think it's more readable if printed out.

lib/LANraragi/Model/Config.pm
```diff
@@ -143,6 +154,7 @@ sub enable_cors          { return &get_redis_conf( "enablecors",    "0" ) }
 sub get_apikey           { return &get_redis_conf( "apikey",        "" ) }
 sub enable_localprogress { return &get_redis_conf( "localprogress", "0" ) }
 sub enable_blacklist     { return &get_redis_conf( "blackliston",   "1" ) }
+sub enable_tagsrules     { return &get_redis_conf( "tagsruleson",   "0" ) }
```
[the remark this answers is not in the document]
IceBreeze: This requires more actions than just the switch of the flag.
Surely, I need to initialize `exptagrules`, but the real problem is that if you switch from Blacklist to Tag rules with one release, users who have edited the blacklist will lose their changes.

We should either allow those users to move their list manually by keeping the two lists for one release or implement a "temporary" function that does that for them.

lib/LANraragi/Controller/Config.pm
IceBreeze: I'll play a bit with this :smile:

lib/LANraragi/Model/Config.pm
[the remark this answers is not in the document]
Difegue: Oh yeah, I planned on writing a migration script-type plugin (ala https://github.com/Difegue/LANraragi/blob/cb485f7dcd29fc8e05ec31b5317bcabd60744a33/lib/LANraragi/Plugin/Scripts/FavtagMigration.pm) to handle existing blacklists, of course.

lib/LANraragi/Controller/Config.pm
IceBreeze: I added the two subs `save_computed_tagrules` and `get_computed_tagrules` to `Utils/Database`.
You should check them because I think they're not what you expected, but the alternative would have been to make `save_config` and `exec_metadata_plugin` bigger than what they already are...

I also added `flat` from `List::Flatten` without importing the library because it seems that you want to keep the project light.
I had no problem adding test libraries because they can be excluded from the installation, but in this case I prefer to ask before doing that ( you are even using `eval` where I would have used `Try::Tiny` :smile: )

lib/LANraragi/Utils/Database.pm
```diff
@@ -303,4 +304,20 @@ sub invalidate_isnew_cache {
...
+    my $redis = LANraragi::Model::Config->get_redis;
+    $redis->del("LRR_TAGRULES");
+    $redis->lpush("LRR_TAGRULES", reverse flat(@$tagrules)) if (@$tagrules);
+    $redis->quit();
+    return;
+}
+
+sub get_computed_tagrules {
+    my $redis = LANraragi::Model::Config->get_redis;
+    my @flattened_rules = $redis->lrange("LRR_TAGRULES", 0,-1);
```
Difegue, line 318: As you said earlier, this won't work properly if the user hasn't changed the settings, since the tag rules won't be calculated yet.  

This sub should probably include a check like `if ($redis->exists("LRR_TAGRULES"))`, and calculate/save the rules in case the list doesn't exist in Redis yet.

lib/LANraragi/Utils/Tags.pm
```diff
@@ -0,0 +1,122 @@
+package LANraragi::Utils::Tags;
```
Difegue, line 1: This new module should be added to the list in `tests/modules.t`.   
(It's already quite well tested by the unit tests you've written, but all modules are in this list for basic perl interpreter checks on CI.)

Difegue (changes requested):
` save_computed_tagrules` and `get_computed_tagrules ` look quite close to what I was expecting/thinking of, so as far as I'm concerned it's all good on that point. 👍  

Got a few more review comments but this is almost ready imo 🤠

lib/LANraragi/Utils/Tags.pm
```diff
@@ -29,6 +29,20 @@ sub restore_CRLF {
     return $val;
 }
 
+# from "List::Flatten"
```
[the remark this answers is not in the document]
IceBreeze: I opted to keep it this time :smiley:

tools/Documentation/advanced-usage/tag-rules.md
```diff
@@ -0,0 +1,77 @@
+---
```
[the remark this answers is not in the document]
IceBreeze: Sure!

lib/LANraragi/Utils/Database.pm
IceBreeze: I can't believe I forgot it :facepalm:

lib/LANraragi/Utils/Tags.pm
IceBreeze: No problem!

Difegue (approved):
Alright, I think this is good enough for an initial merge! 🎊  

I think the display of the rules could be further improved using the jtagger library I recently integrated for metadata edition:  
image
which would also allow to get rid of `replace/restore_CLRF`, but for a first version this is already real good stuff. 🤠

IceBreeze:
> I think the display of the rules could be further improved using the jtagger library I recently integrated for metadata edition:

Yes, I've had a taste for the past few days and I liked it :smiley:

PR #504, opened 2021-09-18
Fix Fakku metadata due to site changes

Nodja:
Fakku recently changed the layout of the website and it broke the metadata scraper.

This fixes search and tag fetching.

I also added tildes (~) to the list of banned chars in search as they're quite common in fakku titles but search breaks if they're present.

I'm not happy with how the elements are being found, but the layout now seems to be generated by some frontend framework that doesn't name anything and is just div vomit. So I had to use user visible text that's unique to the page to find the right elements.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Fakku.pm
+++ b/lib/LANraragi/Plugin/Metadata/Fakku.pm
@@ -25,7 +25,7 @@ sub plugin_info {
         namespace   => "fakkumetadata",
         login_from  => "fakkulogin",
         author      => "Difegue",
-        version     => "0.6",
+        version     => "0.6a",
        description => "Searches FAKKU for tags matching your archive. If you have an account, don't forget to enter the matching cookie in the login plugin to be able to access controversial content.",
         icon =>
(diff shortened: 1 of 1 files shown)


Difegue:
Thanks for taking the time to fix the plugin!

I haven't had time to look at the code diff in detail yet, but it appears the tests are breaking 😅 
Please make sure to fix them before we can get this merged.

Nodja:
Yeah I spent an hour yesterday trying to setup a perl dev environment on my windows machine to run the tests and I don't have time to learn a whole new ecosystem. I was just changing the files and uploading them to my live lanraragi instance.

I'll close the PR but feel free to use my code as reference if you ever get the time to fix it yourself, you can consider my changes public domain. 

[gist with my final file](https://gist.github.com/Nodja/2b29b08d54de64a2dfddbb4c7d996d32)

Difegue:
I understand 🥲 
If you feel like trying again you might have some luck using codespaces: https://github.com/Difegue/LANraragi/codespaces to get a remote dev environment -- I haven't spun one up in a while but that should still work.

Difegue:
The issue was mostly just the HTML samples for the tests being out of date, I didn't have much to fix compared to your original file 👍 Thanks again!

PR #518, opened 2021-10-09
Fix nh param regex

Guerra24:


Difegue:
Looks good after a quick regex101 test, thanks!

PR #538, opened 2021-11-18
Build AVX-free libffi

Guerra24:
Should do the trick. A bit jank but should guarantee that the libffi build is ok. I'm not sure about Platypus tho but it should be ok too.

I picked `x86-64-v2` since I don't think anyone would be running this on pre-2009 systems but if needed we can go down to bare `x86-64` which should work with any x64 cpu.

### DIFF
--- a/tools/build/docker/install-everything.sh
+++ b/tools/build/docker/install-everything.sh
@@ -4,16 +4,22 @@
 apk update
 apk add perl perl-io-socket-ssl perl-dev redis libarchive-dev libbz2 openssl-dev zlib-dev
 apk add imagemagick imagemagick-perlmagick libwebp-tools libheif
-apk add g++ make pkgconf gnupg wget curl nodejs nodejs-npm
+apk add g++ make pkgconf gnupg wget curl nodejs nodejs-npm file
 apk add shadow s6 s6-portable-utils
 
 #Hey it's cpanm
 curl -L https://cpanmin.us | perl - App::cpanminus
+
+#Alpine's libffi build comes with AVX instructions enabled
+#Rebuild our own libffi with those disabled
+if [ $(uname -m) == 'x86_64' ]; then
+  cpanm --notest Alien::FFI
+fi
 
 #Install the LRR dependencies proper
 cd tools && cpanm --notest --installdeps . -M https://cpan.metacpan.org && cd ..
 npm run lanraragi-installer install-full
 
 #Cleanup to lighten the image
-apk del perl-dev g++ make gnupg wget curl nodejs nodejs-npm openssl-dev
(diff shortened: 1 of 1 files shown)


tools/build/docker/install-everything.sh
```diff
@@ -4,16 +4,26 @@
...
-apk add g++ make pkgconf gnupg wget curl nodejs nodejs-npm
+apk add g++ make pkgconf gnupg wget curl nodejs nodejs-npm file
 apk add shadow s6 s6-portable-utils
 
 #Hey it's cpanm
 curl -L https://cpanmin.us | perl - App::cpanminus
 
+#!AVX
+if [ $(uname -m) == 'x86_64' ]; then
+  cpanm --notest --installdeps Alien::FFI
```
Difegue, line 15: IIRC this just installs the required dependency packages and not the package itself, right? 

This whole block could use a bit more comments since it's kinda jank and might be forgotten otherwise. 😅

tools/build/docker/install-everything.sh
```diff
@@ -4,16 +4,26 @@
...
+  curl -L -s https://cpan.metacpan.org/authors/id/P/PL/PLICEASE/Alien-FFI-0.25.tar.gz | tar -xz
+  cd Alien-FFI-0.25
+  sed -i 's/--disable-builddir/--disable-builddir --with-gcc-arch=x86-64-v2/' alienfile
```
Difegue, line 18: I recall seeing that Alien:FFI builds libffi with AVX support by default(according to the related GH issues), is the gcc arch override really necessary?

Difegue: (that or it needs to be set in order to override alpine's built-in libffi? This is another reason more comments'd be awesome 👍)

Difegue (changes requested):
I can't properly test this until monday or so but I'm up for merging it as-is if it fixes nightlies 😅 Just have a few Qs, and more comments'd be appreciated.

tools/build/docker/install-everything.sh
Guerra24: Yes, only deps.

Sure 😅 I'll add comments.

tools/build/docker/install-everything.sh
Guerra24: That I'm not sure. Alpine's libffi build has AVX so I assumed that GCC defaults to it enabled. It's a *just in case* thing.

I'm gonna try on a host that has AVX, if simply installing Alien::FFI is enough then it could be simplified to an one-liner.

Guerra24:
Sometimes it do be that simple.... (I think)
My test with an AVX host worked fine so it's just a matter of testing github action runners.

Difegue:
Neat -- going to merge as is and we can see in 4 hours or so if the resulting containers run fine. 👍

PR #552, opened 2021-11-24
Fix .collapsible-title padding override applying outside of carousel

CirnoT:
This override has been added in v0.8.2 in order to solve visual issue on carousel, however by applying it globally it decreases size of other collapsible elements, such as settings tiles.


Difegue:
Oh woops indeed, it's only in there for the ex and g themes and I didn't test those far enough.  

Thanks a lot for this! (And the other style PRs beforehand. 👍)

PR #599, opened 2022-04-03
Double page laod speedup & Fullscreen & Mouse wheel move

djacks6278:
canvas draw is so slow and heavy when load high resolution image...
so change this code like single page view

no more need to draw in canvas

and i change imagemap to jquery
outer-left, outer-right is now work perfectly

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -284,7 +284,6 @@ p#nb {
   max-width:100%;
   height: auto;
   width: auto;
-  min-height:300px;
   user-select: none;
 }
 
@@ -690,4 +689,39 @@ body.infinite-scroll #display img {
 
 body.infinite-scroll .sb {
   margin-top: 10px;
+}
+
+/* Double Page & Fullscreen */
+
+@media screen and (orientation:portrait) {
+  /* Image size fix when device is portrait and double page mode */
+  div.sni .double-mode img {
+    width: calc(50% - 2px);
+  }
+  div.fullscreen .double-mod img {
+    width: calc(50%);
+  }
+}
+
+div.sni img[src=""] {
+  display: none;
+}
+
+div.fullscreen img {
+  height: unset !important;
+  max-height: 100vh !important;
+  margin: 0px !important;
+}
+
+#display {
+  display: flex;
+  justify-content: center;
+}
+
+.fullscreen-infinite {
+  overflow-y: scroll;
+}
+
+body.infinite-scroll .fullscreen-infinite img {
+  margin: 0 auto !important;
 }
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -698,7 +698,7 @@ body.infinite-scroll .sb {
   div.sni .double-mode img {
     width: calc(50% - 2px);
   }
-  div.fullscreen .double-mod img {
+  div.fullscreen .double-mode img {
     width: calc(50%);
   }
 }
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -111,32 +111,26 @@ Reader.loadImages = function () {
             if (Reader.infiniteScroll) {
                 Reader.initInfiniteScrollView();
             } else {
-                $(window).on('wheel', function(e) {
+                $(window).on("wheel", function(e) {
                     if(Reader.isFullscreen) {
-                        if (e.originalEvent.deltaY > 0) Reader.changePage(-1);
-                        else Reader.changePage(1);
+                        let changePage_val = 1;
+                        if (e.originalEvent.deltaY > 0) changePage_val = -1;
+                        if (!Reader.mangaMode) changePage_val *= -1;
+                        Reader.changePage(changePage_val);
                         return false;
                     }
                 });
-                $(document).on('click', function(event) {
-                    var img_y_offset = $('a#display img').offset()['top'] + $('a#display img').height();
-                    if (event.pageY < img_y_offset && !$('#overlay-shade').is(':visible') ){
-                        var whereside = event.pageX < $(window).width()/2 ? 'left' : 'right';
-                        if (!$(event.target).closest('a#display').length) {
-                            whereside = 'outer-'+whereside;
-                        }
-                        switch (whereside) {
-                        case "outer-left":
-                            Reader.changePage("first");
-                            break;
+                // when click left or right img area change page
+                $(document).on("click", function(event) {
+                    // check click Y position is in img Y area 
+                    if (event.pageY < ($("a#display img").offset()["top"] + $("a#display img").height()) && !$("#overlay-shade").is(":visible") ){
+                        // is click X position is left on screen or right
+                        switch (event.pageX < $(window).width()/2 ? "left" : "right") {
                         case "left":
                             Reader.changePage(-1);
                             break;
                         case "right":
                             Reader.changePage(1);
-                            break;
-                        case "outer-right":
-                            Reader.changePage("last");
                             break;
                         default:
                             break;
@@ -344,9 +338,9 @@ Reader.updateMetadata = function () {
     const img = $("#img")[0];
     const imageUrl = new URL(img.src);
     const filename = imageUrl.searchParams.get("path");
-    const img_pre = $("#img_pre")[0];
-    const imageUrl_pre = new URL(img_pre.src);
-    const filename_pre = imageUrl_pre.searchParams.get("path");
+    const img_doublepage = $("#img_doublepage")[0];
+    const imageUrl_doublepage = new URL(img_doublepage.src);
+    const filename_doublepage = imageUrl_doublepage.searchParams.get("path");
     if (!filename && Reader.showingSinglePage) {
         Reader.currentPageLoaded = true;
         $("#i3").removeClass("loading");
@@ -355,10 +349,10 @@ Reader.updateMetadata = function () {
 
     const width = img.naturalWidth;
     const height = img.naturalHeight;
-    const width_pre = img_pre.naturalWidth;
-    const height_pre = img_pre.naturalHeight;
-    const width_view = width + width_pre;
-    const height_view = height > height_pre ? height : height_pre;
+    const width_doublepage = img_doublepage.naturalWidth;
+    const height_doublepage = img_doublepage.naturalHeight;
+    const width_view = width + width_doublepage;
+    const height_view = height > height_doublepage ? height : height_doublepage;
 
     if (Reader.showingSinglePage) {
         // HEAD request to get filesize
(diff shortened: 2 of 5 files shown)


public/js/reader.js
```diff
@@ -234,6 +267,9 @@ Reader.handleShortcuts = function (e) {
     case 68: // d
         Reader.changePage(1);
         break;
+    case 70: // d
```
Difegue, line 270:
```suggestion
    case 70: // f
```

public/js/reader.js
```diff
@@ -234,6 +267,9 @@ Reader.handleShortcuts = function (e) {
...
+        Reader.toggleFullScreen();
+        break;
     case 72: // h
         Reader.toggleHelp();
```
Difegue, line 274: The help text needs to be updated to mention the new fullscreen option.

public/js/reader.js
```diff
@@ -331,7 +374,36 @@ Reader.updateMetadata = function () {
...
             });
         } else { $(".file-info").text(`${filename} :: ${width} x ${height} :: ${size} KB`); }
-    } else { $(".file-info").text(`Double-Page View :: ${width} x ${height}`); }
+    } else { 
+        // HEAD request to get filesize
+        let size = Reader.preloadedSizes[Reader.currentPage];
+        let size_pre = Reader.preloadedSizes[Reader.currentPage+1];
+        let size_view = size + size_pre;
+        if (!size || !size_pre) {
+            $.ajax({
```
Difegue, line 383: Since we're doing the HEAD call three times in this JS file now, it'd be better to refactor it to a specific function in `common.js` and call it here.

public/js/reader.js
```diff
@@ -331,7 +374,36 @@ Reader.updateMetadata = function () {
...
+                    Reader.preloadedSizes[Reader.currentPage] = size;
+                    $.ajax({
+                        url: Reader.pages[Reader.currentPage+1],
+                        type: "HEAD",
+                        success: (data, textStatus, request) => {
+                            size_pre = parseInt(request.getResponseHeader("Content-Length") / 1024, 10);
+                            Reader.preloadedSizes[Reader.currentPage+1] = size_pre;
+                            size_view = size + size_pre;
+                            $(".file-info").text(`${filename} - ${filename_pre} :: ${width_view} x ${height} :: ${size_view} KB`);
+                            $(".file-info").attr("title", `${filename} :: ${width} x ${height} :: ${size} KB - ${filename_pre} :: ${width_pre} x ${height_pre} :: ${size_pre} KB`);
```
Difegue, line 397: Nice touch, but the `attr` parameter isn't removed if we switch back to single page mode.  

image  

I think nobody will know this exists here either -- It might be interesting to move this to a small `<i>` element alongside the text, which is only shown when doublePage mode is enabled.  
(crude mockup below)
image

```
<span>
<div class="file-info" style="display: inline;">002.jpg - 003.jpg :: 2128 x 1544 :: 548 KB</div>
<i class="fa fa-info-circle" id="doublepage-detail" style="display: inline;margin-left: 8px;" title="002.jpg :: 1064 x 1544 :: 279 KB - 003.jpg :: 1064 x 1527 :: 269 KB"></i>
</span>
```

public/js/reader.js
```diff
@@ -109,7 +111,38 @@ Reader.loadImages = function () {
...
+                });
+                $(document).on('click', function(event) {
+                    var img_y_offset = $('a#display img').offset()['top'] + $('a#display img').height();
+                    if (event.pageY < img_y_offset && !$('#overlay-shade').is(':visible') ){
+                        var whereside = event.pageX < $(window).width()/2 ? 'left' : 'right';
+                        if (!$(event.target).closest('a#display').length) {
+                            whereside = 'outer-'+whereside;
+                        }
+                        switch (whereside) {
+                        case "outer-left":
```
Difegue, line 129: Both this and `outer-right` breaks clicking on the arrow icons to move pages;  
I personally think those cases should be removed, since:  
- They're hard to figure out and will probably just confuse users who activate them by mistake 
- If users want to get to the first or last page they can just click the matching arrow shortcuts. (there's much less use for first/last page compared to -1/+1)
- It makes the above bit of code that much harder to read 😅

public/js/reader.js
```diff
@@ -109,7 +111,38 @@ Reader.loadImages = function () {
             if (Reader.infiniteScroll) {
                 Reader.initInfiniteScrollView();
             } else {
-                $(document).on("click.imagemap_change_pages", "#Map area", Reader.handlePaginator);
+                $(window).on('wheel', function(e) {
+                    if(Reader.isFullscreen) {
```
Difegue, line 115: This doesn't seem to work properly in manga/reverse mode.

public/js/reader.js
```diff
@@ -109,7 +111,38 @@ Reader.loadImages = function () {
...
+                        if (e.originalEvent.deltaY > 0) Reader.changePage(-1);
+                        else Reader.changePage(1);
+                        return false;
+                    }
+                });
+                $(document).on('click', function(event) {
+                    var img_y_offset = $('a#display img').offset()['top'] + $('a#display img').height();
```
Difegue, line 122: This needs some extra comments.

public/js/reader.js
```diff
@@ -541,6 +621,50 @@ Reader.toggleArchiveOverlay = function () {
     return LRR.toggleOverlay("#archivePagesOverlay");
 };
 
+Reader.toggleFullScreen = function () {
+  // if already full screen; exit
+  // else go fullscreen
+  if (
+    document.fullscreenElement ||
```
Difegue, line 628: Can we get by just using `document.fullscreenElement` and `document.webkitFullscreenElement`?
I don't care about compatibility with older browsers (it's already broken since I use a bit of ES6 here and there), and the API seems to be supported enough without having to use vendor-specific properties (except for safari): https://caniuse.com/mdn-api_document_fullscreenelement

public/js/reader.js
```diff
@@ -541,6 +621,50 @@ Reader.toggleArchiveOverlay = function () {
...
+      document.mozCancelFullScreen();
+    } else if (document.webkitExitFullscreen) {
+      document.webkitExitFullscreen();
+    } else if (document.msExitFullscreen) {
+      document.msExitFullscreen();
+    }
+    Reader.isFullscreen = false;
+  } else {
+    if($('body').hasClass('infinite-scroll')) {
+        $('div#i3').addClass('fullscreen-infinite');
```
Difegue, line 650: This breaks fullscreen in infinite scrolling mode for now.

Difegue (changes requested):
👋 Thanks a lot for willing to contribute with Reader improvements! 🙏  
As you've probably seen it's a bit complex, so I'll gladly accept the work.  

Most of the changes here are good (I especially appreciate fullscreen mode!), but I tested the PR and found a few issues that have to be fixed before I can merge this.

public/themes/modern.css
```diff
@@ -351,15 +351,14 @@ td.itdc {
 div.sni {
 	text-align: center;
 	margin: 2px auto 6px;
-	padding: 0 10px 5px;
+	padding: 0 0px 5px;
```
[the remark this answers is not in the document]
djacks6278: oh.. that's a personal theme setup. i see this css don't need to commit

templates/reader.html.tt2
```diff
@@ -40,11 +40,7 @@
             </div>
 
             <a id="display">
-                <img id="img" class="reader-image" usemap="#Map" />
-                <map name="Map" id="Map">
-                    <area id="leftmap" style="cursor:pointer;" shape="rect" value="left" />
-                    <area id="rightmap" style="cursor:pointer;" shape="rect" value="right" />
-                </map>
+                <img id="img" class="reader-image"><img id="img_pre" class="reader-image">
```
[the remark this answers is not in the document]
djacks6278: that line is need to be 1 line
cuz when split 2 line system make some blank space between img div

public/js/reader.js
djacks6278: i need check that

public/js/reader.js
djacks6278: when click screen js caculate img area space and window space

and check click position on page
and do if click position Y < img div height 
then caculate left or right on screen
so now we got a same work like map

i will add comment this action

public/js/reader.js
djacks6278: some case document.fullscreenElement or webkit is need to check fullscreen
cuz some lower version chrome or firefox or edge is need that

public/themes/modern.css
```diff
@@ -374,6 +373,10 @@ div.sni img {
 	margin: 1px;
 	clear: both;
 }
+div.fullscreen img {
```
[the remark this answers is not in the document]
djacks6278: so thankful!

djacks6278:
@Difegue i don't know how close the your review
but i resolve all thing in your review

so u need check again!

public/themes/modern.css
Difegue, line 354: Changes in this file haven't been removed from the PR, it seems. 🙂  
(The help text also hasn't been updated to mention F for fullscreen 🙏)

public/css/lrr.css
```diff
@@ -690,4 +689,39 @@ body.infinite-scroll #display img {
...
+}
+
+/* Double Page & Fullscreen */
+
+@media screen and (orientation:portrait) {
+  /* Image size fix when device is portrait and double page mode */
+  div.sni .double-mode img {
+    width: calc(50% - 2px);
+  }
+  div.fullscreen .double-mod img {
```
Difegue, line 701:
```suggestion
  div.fullscreen .double-mode img {
```
?

public/js/reader.js
```diff
@@ -111,33 +111,27 @@ Reader.loadImages = function () {
...
-                            whereside = 'outer-'+whereside;
-                        }
-                        switch (whereside) {
-                        case "outer-left":
-                            Reader.changePage("first");
-                            break;
+                // when click left or right img area change page
+                $(document).on("click", function(event) {
+                    // check click Y position is in img Y area 
+                    if (event.pageY < ($("a#display img").offset()["top"] + $("a#display img").height()) && !$("#overlay-shade").is(":visible") ){
```
Difegue, line 126: This needs a check for `event.pageY > [height of header]` in case the header is visible.

Difegue: Also, clicking the arrow buttons triggers this and causes the page to switch twice. 🥲

public/js/reader.js
```diff
@@ -355,10 +349,10 @@ Reader.updateMetadata = function () {
...
-    const height_pre = img_pre.naturalHeight;
-    const width_view = width + width_pre;
-    const height_view = height > height_pre ? height : height_pre;
+    const width_doublepage = img_doublepage.naturalWidth;
+    const height_doublepage = img_doublepage.naturalHeight;
+    const width_view = width + width_doublepage;
+    const height_view = height > height_doublepage ? height : height_doublepage;
 
     if (Reader.showingSinglePage) {
         // HEAD request to get filesize
```
Difegue, line 358: Use `LRR.getImgSize(Reader.pages[Reader.currentPage]);` here as well.

public/js/reader.js
```diff
@@ -111,33 +111,27 @@ Reader.loadImages = function () {
...
                 Reader.initInfiniteScrollView();
             } else {
-                $(window).on('wheel', function(e) {
+                $(window).on("wheel", function(e) {
                     if(Reader.isFullscreen) {
-                        if (e.originalEvent.deltaY > 0) Reader.changePage(-1);
-                        else Reader.changePage(1);
+                        let changePage_val = 1;
+                        if (e.originalEvent.deltaY > 0) changePage_val = -1;
+                        if (!Reader.mangaMode) changePage_val *= -1;
```
Difegue, line 118: An extra comment'd be nice here:  
```suggestion
                        // In Manga mode, reverse the changePage variable so that we always move forward
                        if (!Reader.mangaMode) changePage_val *= -1;
```

public/js/reader.js
```diff
@@ -438,21 +424,24 @@ Reader.goToPage = function (page) {
...
             if (Reader.mangaMode) {
                 $("#img").attr("src", img2.src);
-                $("#img_pre").attr("src", img1.src);
+                $("#img_doublepage").attr("src", img1.src);
             } else {
                 $("#img").attr("src", img1.src);
-                $("#img_pre").attr("src", img2.src);
+                $("#img_doublepage").attr("src", img2.src);
             }
+            $("a#display").addClass("double-mode");
```
Difegue, line 438: There's only one `display` element, so is the `a` selector needed?
```suggestion
            $("#display").addClass("double-mode");
```

public/js/reader.js
```diff
@@ -438,21 +424,24 @@ Reader.goToPage = function (page) {
         if (img1.naturalWidth > img1.naturalHeight || img2.naturalWidth > img2.naturalHeight) {
             // Depending on whether we were going forward or backward, display img1 or img2
             $("#img").attr("src", Reader.previousPage > Reader.currentPage ? img2.src : img1.src);
-            $("#img_pre").attr("src", "");
+            $("#img_doublepage").attr("src", "");
+            $("a#display").removeClass("double-mode");
```
Difegue, line 428: It'd be better to move this block:  
```
$("#img_doublepage").attr("src", "");
$("#display").removeClass("double-mode");
``` 
before the `if` statement on line 420, so that we only have to write it once 👍 (and not twice in both branches)

public/js/reader.js
```diff
@@ -308,6 +338,9 @@ Reader.updateMetadata = function () {
     const img = $("#img")[0];
     const imageUrl = new URL(img.src);
     const filename = imageUrl.searchParams.get("path");
+    const img_doublepage = $("#img_doublepage")[0];
```
Difegue, line 341: I forgot to mention this in the previous review, but I try to stick to camelCase for variable names on the JS side of things.  

It'd be awesome if you could update the code to be as compliant as possible with the rules in https://github.com/Difegue/LANraragi/blob/dev/.eslintrc.json ! 🙏

public/js/reader.js
```diff
@@ -316,6 +349,10 @@ Reader.updateMetadata = function () {
 
     const width = img.naturalWidth;
     const height = img.naturalHeight;
+    const width_doublepage = img_doublepage.naturalWidth;
+    const height_doublepage = img_doublepage.naturalHeight;
+    const width_view = width + width_doublepage;
+    const height_view = height > height_doublepage ? height : height_doublepage;
```
Difegue, line 355: This constant doesn't seem to be used.

public/js/reader.js
```diff
@@ -541,6 +610,50 @@ Reader.toggleArchiveOverlay = function () {
...
+      document.msExitFullscreen();
+    }
+    Reader.isFullscreen = false;
+  } else {
+    if($("body").hasClass("infinite-scroll")) {
+        $("div#i3").addClass("fullscreen-infinite");
+    } else {
+        $("div#i3").addClass("fullscreen");
+    }
+    element = $("div#i3").get(0);
```
Difegue, line 643:
```suggestion
    const element = $("div#i3").get(0);
```

Difegue (changes requested):
Lookin' good and almost bug free it seems! 👀 
A few more nitpicks from me and I think we can merge this afterwards.

public/css/lrr.css
djacks6278: oh that's typo i missed

public/js/reader.js
djacks6278: maybe if we check height of header then button trigger will be fix

public/js/reader.js
djacks6278: my misstake

public/js/reader.js
djacks6278: this constant use when caculate file width and height size
but i just copy your code single page width height size code

public/js/reader.js
djacks6278: if we don't need outer click just change if check part

djacks6278:
i fixed all review well i'm too late sorry about that

Difegue (approved):

Difegue:
Yup, looks great! 🎊  
Once again, thanks a ton for taking the time to work on this.

Issue #601, opened 2022-04-10
Import meta/tags using provided URL while download new archives by built-in downloader


xiazeyu:
When want to download an archive from the Internet, 
I use the **Adding Archives to the Library** function combined with Metadata Plugins, hoping to automatically add meta.

Take E-Hentai plugin for example, It seems that if I turn on "Run Automatically" and give "URL(s) to download".

It will always match metadata using **TITLE** instead of **URL** I provided in the downloader.

This can result in mismatch when the title contains Unicode characters, or multiple gallary shares the similar title.

My current solution is to disable "Run Automatically" option and Use "Click here to edit metadata." function to manually paste the "E-H Gallery URL (Will attach tags matching this exact gallery to your archive)".

Is it possible to just let "Mayriad's EH Master Script" to retrieve the URL I provided in the downloader, and use that information to search for metadata?

xiazeyu:
I've switched to the `nightly` branch, and some test will be done later.

xiazeyu:
I looked into the Meta plugin, found 

https://github.com/Difegue/LANraragi/blob/19756a1075da4cceaa311e97e931b294ed63f537/lib/LANraragi/Plugin/Metadata/MEMS.pm#L64

https://github.com/Difegue/LANraragi/blob/19756a1075da4cceaa311e97e931b294ed63f537/lib/LANraragi/Plugin/Metadata/EHentai.pm#L72

these codes seems to extract the "source" tag from existing tags.

However, when I download a new comic using `E*Hentai Downloader v.1.0 by Difegue`, the only tag I seen is "date_added:1649757812". These's no tag like "source" added, even though I download it by "Add Archives". Is it expected? Or should the "source" tag being added after the file is downloaded(just like how the "date_added" tag is added), and before Meta plugin is called?

xiazeyu:
Another question, may be off-topic. Is there any difference between `E-Hentai v.2.5.1 by Difegue and others` and `Mayriad's EH Master Script v.1.1.0 by Mayriad`?
It seems that the MEMS just "[Mayriad's EH Master Script] [info] Source identified. Calling E-Hentai metadata plugin to retrieve metadata from EH API."

Difegue:
The MEMS plugin is only useful if you have the matching userscript on your browser: The base EH plugin does everything you'd need as a regular user. 

As for downloading through add archives, it calls on the DL system all the same so you should be getting source tags.
More investigating needed..

xiazeyu:
[...]

Issue #297, opened 2020-05-08
[Suggestion] Allow users to change LRR's plug-in settings before the initial scan


Asinin3:
If people could configure LRR to how they wanted it set-up before LRR goes through and scans all their files, It would really help stop people running into a lot of issues and using the batch-tagger after-the-fact. Alternatively, these files could still be added, but the auto-tagger wouldn't run on the new files until a user goes through and configures LRR

Then just add a big "Initial Scan" button on the index page that appears for new users perhaps with some text saying to check that the plugin settings and tag blacklist is set to their liking before they run the scan.

Currently new users that already have a bunch of galleries are installing LRR, pointing it to their library, and then have to go back and re-run the batch tagger.

Difegue:
From a design perspective this is actually pretty easy : Just don't start the background worker immediately for first-timers. Starting it after boot can be done using the existing API endpoints.

The main workload here is designing a proper, understandable out-of-the-box user experience: Adding a popup on the index to inform the user could work, I guess?

PR #604, opened 2022-04-14
remove timestamp from global system config

xiazeyu:
fixed https://github.com/Difegue/LANraragi/issues/601


xiazeyu:
Is there any way to enable such plugin by default?

Difegue:
See https://github.com/Difegue/LANraragi/issues/601#issuecomment-1100099437 .

Difegue:
There's no way to enable plugins by default at the moment either, since I'd rather make plugins more obvious to first-time users through things like https://github.com/Difegue/LANraragi/issues/297 .

PR #605, opened 2022-04-14
Support for external JSON metadata files

### LINKED ISSUE #498: [Suggestion] Support for external JSON metadata files
## Suggestion, and the problem it'd solve

**Support for external JSON meta files.** I have lots of (a couple thousand) archives where the JSON metadata isn't included inside the ZIP file but inside the same directory as the ZIP file is located (visualized example below). This is mostly because they are torrents from ExH which don't include the metadata inside the ZIP. I don't really want to move the metadata inside those ZIP files as my torrent client would not recognize them anymore + other issues like file integrity.

The name of the metadata file can be an exact match but not always. When it's not: at least the title and the author/circle could used used as rules for a fuzzy matching.

Probably the name (e.g. `プロトタイプ マドモアゼル` below) could be the minimum requirement for a match (aside from whitespace) with an option to also match the author/circle (e.g. `リンゴヤ (あるぷ)` below).

Conflicts would probably be rare as usually people only have one version of the doujinshi, but of course not impossible. In that case LANraragi could maybe list all the conflicts in text form, but this is not as important at the moment I think.

## Examples:

📂archive_dir
 ┣ 📜(成年コミック) [40010試作型] プロトタイプ マドモアゼル.zip
 ┣ 📜[40010試作型] プロトタイプ マドモアゼル [DL版].json
 ┣ 📜(COMIC1☆11) [リンゴヤ (あるぷ)] TRANCE CONTROL (ラブライブ! サンシャイン!!).zip
 ┗ 📜(COMIC1☆11) [リンゴヤ (あるぷ)] TRANCE CONTROL (ラブライブ! サンシャイン!!).json

The first two files and last two files should match together.

Those JSON files are in the standard format which is also used by the eze plugin. Example for `[40010試作型] プロトタイプ マドモアゼル [DL版].json`: [[40010試作型] プロトタイプ マドモアゼル [DL版].txt](https://github.com/Difegue/LANraragi/files/7086540/40010.DL.txt)

The extension here is txt because GitHub didn't allow uploading json files. Also, omitted links and other sensitive data.

xiazeyu:
fix #498 
Support for external eze JSON metadata files

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Eze.pm
+++ b/lib/LANraragi/Plugin/Metadata/Eze.pm
@@ -6,6 +6,7 @@ use warnings;
 #Plugins can freely use all Perl packages already installed on the system
 #Try however to restrain yourself to the ones already installed for LRR (see tools/cpanfile) to avoid extra installations by the end-user.
 use Mojo::JSON qw(from_json);
+use File::Basename;
 
 #You can also use the LRR Internal API when fitting.
 use LANraragi::Model::Plugins;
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Plugin/Metadata/Eze.pm
```diff
@@ -42,6 +43,11 @@ sub get_tags {
 
     my $logger = get_plugin_logger();
     my $path_in_archive = is_file_in_archive( $lrr_info->{file_path}, "info.json" );
+
+    my @suffixlist = qw(.zip .cbz .rar .cbr .tar.gz .cbt .lzma .7z .cb7 .xz .pdf .epub);
+    my ($name, $path, $suffix) = fileparse($lrr_info->{file_path}, @suffixlist);
```
Difegue, line 48: It's fine to use the classic regexp here, the `file_path` variable given by the info hash will always be an archive anyway. 
```suggestion
    my ($name, $path, $suffix) = fileparse($lrr_info->{file_path}, qr/\.[^.]*/);
```

lib/LANraragi/Plugin/Metadata/Eze.pm
```diff
@@ -25,7 +26,7 @@ sub plugin_info {
         author    => "Difegue",
         version   => "2.2",
         description =>
-          "Collects metadata embedded into your archives as eze-style info.json files. ({'gallery_info': {xxx} } syntax)",
+          "Collects metadata from eze-style info.json file in archives or {archive_name}.json file nearby the archive file. ({'gallery_info': {xxx} } syntax)",
```
Difegue, line 29:
```suggestion
          "Collects metadata from eze-style info.json files ({'gallery_info': {xxx} } syntax), either embedded in your archive or in the same folder with the same name. ({archive_name}.json)",
```

lib/LANraragi/Plugin/Metadata/Eze.pm
```diff
@@ -79,8 +85,43 @@ sub get_tags {
             return ( tags => $tags );
         }
 
+    } elsif (-e $path_nearby_json){
```
Difegue, line 88: Don't duplicate the entire codeblock here, just create a new sub and call it both here and in the `path_in_archive` case.  
```
elsif (-e $path_nearby_json){
 
$logger->debug("Found file $path_nearby_json");
my $hashjson = consume_json($path_nearby_json);

#Parse it
my ( $tags, $title ) = tags_from_eze_json($hashjson);
[...]
}

sub consume_json() {
     my $filepath = $_[0];

     #Open it
     my $stringjson = "";
     [...]
     $logger->debug("Found nearby and loaded the following JSON: $stringjson");
     return $hashjson;
}
```

Difegue (changes requested):
Looking good! Just got a few general code quality remarks. 🙌

lib/LANraragi/Plugin/Metadata/Eze.pm
```diff
@@ -42,86 +42,60 @@ sub get_tags {
...
-    my @suffixlist = qw(.zip .cbz .rar .cbr .tar.gz .cbt .lzma .7z .cb7 .xz .pdf .epub);
-    my ($name, $path, $suffix) = fileparse($lrr_info->{file_path}, @suffixlist);
+    my ($name, $path, $suffix) = fileparse($lrr_info->{file_path}, qr/\.[^.]*/);
     my $path_nearby_json = $path . $name . '.json';
 
-    if ($path_in_archive) {
-
-        #Extract info.json
-        my $filepath = extract_file_from_archive( $lrr_info->{file_path}, $path_in_archive );
+    my $filepath = undef;
```
Difegue, line 51: The critic CI shouldve yelled about this too (looks like its checks API integration is broken oops), but there's no need to initialize variables to `undef`:  

```
my $filepath;
my $delete_after_parse;
```

Difegue (changes requested):
One small nitpick left and I'll get this merged. 👍 Thanks again!

Difegue (approved):

Difegue:
All good here, thanks again! 👍

PR #620, opened 2022-05-01
New toast and notification dependcies

xiazeyu:
Great thanks to @EYHN for CSS styling guide.


public/js/init_toast.js
```diff
@@ -0,0 +1,25 @@
...
+        draggable: c.draggable ?? ((c.icon || c.type) === "info" || (c.icon || c.type) === "success") ?? false,
+    });
+    const toastDiv = document.createElement("div");
+    document.body.appendChild(toastDiv);
+    toastDiv.style.textAlign = "initial";
+    window.React.render(
+        window.React.createElement(window.reactToastify.ToastContainer, {
+            style: {},
+            limit: 7,
+            theme: "colored",
```
Difegue, line 23: I believe this is a good opportunity to style the toasts according to the loaded css theme (especially since toastify has colored loading bars and icons depending on the status which look quite nice)
```suggestion
            theme: "light",
```  

We should then define custom `--toastify-color-light` and `--toastify-text-color-light` variables in the theme files. An example for modern.css:  

```
:root {
  --toastify-text-color-light: #DDDDDD;
  --toastify-color-light: #4F535B;
}

.Toastify__close-button--light {
  color: #fff;
  opacity: 1;
}
```  

image
image

public/js/init_toast.js
```diff
@@ -0,0 +1,25 @@
+jQuery(() => {
+    window.toast = (c) => window.reactToastify.toast(window.React.createElement("div", { dangerouslySetInnerHTML: { __html: `<h2>${c.heading}</h2><p>${c.text}</p>` } }), {
```
Difegue: Additional css to add in lrr.css to align the toast icon at the top:  

```
.Toastify__toast-icon {
  margin-top: 12px;
  margin-bottom: auto;
}
```

public/js/index.js
```diff
@@ -484,9 +498,21 @@ Index.handleContextMenu = function (option, id) {
...
+        window.Swal.fire({
+            title: "Are you sure?",
+            text: "This is a destructive operation! Are you sure you want to delete this archive?",
+            icon: "warning",
+            showCancelButton: true,
+            focusConfirm: false,
+            confirmButtonText: "Yes, delete it!",
+            reverseButtons: true,
+            confirmButtonColor: "#d33",
+            cancelButtonColor: "#3085d6",
```
Difegue, line 510: I believe non-destructive buttons should follow the color of the theme's buttons. 
```suggestion
            cancelButtonColor: $('.stdbtn').css("backgroundColor"),
```
If there's a way to attach a class to buttons, it'd probably be easier to add the `stdbtn` class to them here.

public/js/index.js
```diff
@@ -484,9 +498,21 @@ Index.handleContextMenu = function (option, id) {
         LRR.openInNewTab(`./edit?id=${id}`);
         break;
     case "delete":
-        if (window.confirm("Are you sure you want to delete this archive?")) {
-            Server.deleteArchive(id, () => { document.location.reload(true); });
-        }
+        window.Swal.fire({
```
Difegue, line 501: Similarly, swal2 popups should be themed (example for modern.css)  

```
.swal2-popup {
  background: #43464E !important;
  color: #DDDDDD !important;
}
```

image

tools/install.pl
```diff
@@ -201,10 +203,14 @@ sub cp_node_module {
...
     $item =~ /([^\/]+$)/;
     my $newname = getcwd . $newpath . $&;
+    my $nodemapname = $nodename . ".map";
+    my $newmapname = $newname . ".map";
 
-    say("Copying $nodename \r\n to $newname \r\n");
+    say("\r\nCopying $nodename \r\n to $newname");
     copy( $nodename, $newname ) or die "The copy operation failed: $!";
 
+    my $mapresult = copy( $nodemapname, $newmapname ) and say("Copied sourcemap file.\r\n");;
```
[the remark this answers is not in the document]
xiazeyu: source maps files will not be loaded into Browser unless user entered the `Developer's Tools`.
It will be fine to leave it, unless you cares about the docker image size very very much.

public/js/server.js
```diff
@@ -254,22 +246,15 @@ Server.deleteArchive = function (arcId, callback) {
...
         .then((data) => {
             if (data.success === "0") {
-                $.toast({
-                    showHideTransition: "slide",
-                    position: "top-left",
-                    loader: false,
+                window.toast({
                     heading: "Couldn't delete archive file. <br> (Maybe it has already been deleted beforehand?)",
                     text: "Archive metadata has been deleted properly. <br> Please delete the file manually before returning to Library View.",
-                    hideAfter: false,
```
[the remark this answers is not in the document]
xiazeyu: hideAfter and autoClose are both OK due to the "compatability layer" I wrote.
It's your decision if these toastes should be auto closed.
I removed `hideAfter: false`, so that toasts can be closed in default timer. (5s for normal info and 7s for warning and errors. In both situation, lose focus or hover the mouse on the toast can pause the timer.)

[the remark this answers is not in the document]
xiazeyu: Personally I don't want these warnings stay in the page all the time.

public/js/init_toast.js
xiazeyu: Yep, I may not cover all the situations. But changing colored to light may result in significant differences. More tests should be done later.

tools/install.pl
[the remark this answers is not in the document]
Difegue: I try to keep the docker image as small as possible so yeah, I'd rather have them out sadly 😅  
If you want to keep them for source builds/debug purposes we could remove them in the cleanup phase of the Docker image build here: https://github.com/Difegue/LANraragi/blob/2cb71c2c601927d6ec79d5e2b459327770efa0d3/tools/build/docker/install-everything.sh#L30

public/themes/g.css
```diff
@@ -513,3 +513,37 @@ div.id4 {
...
 }
 
+
+/** handle the notificaiton color and the text color based on the theme **/
+/** handle the notificaiton color and the text color based on the theme **/
+.Toastify {
+    --toastify-font-family: "Open Sans", arial, sans-serif;
+  }
+  
+  .Toastify__toast-body {
```
Difegue, line 523: Most of this css seems to be common across the 5 themes, so it should probably be in lrr.css for now.

public/themes/modern_clear.css
```diff
@@ -551,3 +551,37 @@ div.id4 a:hover {
 .awesomplete li:hover mark, .awesomplete li[aria-selected="true"] mark {
     background: #ED2553;
 }
+
+
+/** handle the notificaiton color and the text color based on the theme **/
+.Toastify {
+    --toastify-font-family: "Open Sans", arial, sans-serif;
```
Difegue, line 558: Except for the fonts -- This theme uses Inter UI, not Open Sans

public/themes/modern_red.css
```diff
@@ -621,3 +621,36 @@ div.id4 a:hover {
   background: #E9A53A;
   color: #F1F1F1;
 }
+
+
+/** handle the notificaiton color and the text color based on the theme **/
+.Toastify {
+  --toastify-font-family: "Open Sans", arial, sans-serif;
```
Difegue, line 628: And this one uses Roboto.

public/themes/modern.css
```diff
@@ -597,3 +597,14 @@ div.id4 {
...
+
+/** handle the popup color and the text color based on the theme **/
+.swal2-popup {
+  background: #3c3f44;
+  color: #e1e1e1;
+}
+
+body.swal2-shown > [aria-hidden="true"] {
+  transition: 0.1s filter;
+  filter: blur(3px);
```
Difegue, line 609: `filter: blur` looks out of place compared to the other overlays in the frontend who just darken the background content (`opacity: 0.6`); I don't think this should be added.

public/js/server.js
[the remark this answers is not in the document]
Difegue: Most of those aren't warnings but full-on errors, which I sincerely believe shouldn't dismiss on their own as we must make sure the user gets to see them.  
Warnings dismissing after a longer time is OK.  

The remaining toasts that used `hideAfter: false` are the help ones since they're a bit long to read -- For this the hover-on-toast to pause the timer behavior should indeed be enough.

public/js/init_toast.js
Difegue: I personally prefer the modded-light look to the colored one, but we can adjust that later indeed.

public/js/init_toast.js
Difegue: The method with `flex` is better, nice job 👍

public/js/index.js
Difegue: Needs colors for the other themes

Difegue (changes requested):
The new styling looks good! Good job on preserving the clickable links and such.   
Toastify + sweetalert is a nice combo that sets a good base for further modernizations of the app, so pretty stoked to get this merged. 👍🔥

tools/install.pl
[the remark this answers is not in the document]
xiazeyu: Please have a look if the added line works correctly. (Test fine on my machine.)

public/js/server.js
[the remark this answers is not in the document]
xiazeyu: Yep. The toast now follows this rules:
success: close after 5s
info: close after 5s
warning: close after 10s
error: never auto close

All the toast are read by myself, and the reading time is measured using a timer, as long as possible.

We also provide a fallback if mouse hovers on the toast. (Not sure if it is intuitive enough, but it is now treated as a fallback)

Besides, it is possible to give every toast a unique ID, so they won't duplicate, and if error keeps occuring, when one toast with its unique ID disappear, a same one can be generated to make it always on screen. This needs tons of work, and its not implement yet, but possible.

public/js/index.js
xiazeyu: I prefer use the original color.
```js
confirmButtonColor: "#d33",
```
1. image
image


True that we can use 
```js
customClass: {
    cancelButton: "stdbtn",
},
confirmButtonColor: "#d33",
```
to attach stdbtn class to the button.
2. image
image

However, as you can see, the existing css only have the `button`, and don't have `cancelButton`. Also, the shape, on:hover and formating is different than the SweetAlert one's. So you we may rather write new classes for destructive buttons or using the same class for both buttons.

```js
customClass: {
    cancelButton: "stdbtn",
    confirmButton: "stdbtn",
},
confirmButtonColor: "#d33",
// with background changed to #484b51
```

3. image
image

public/js/index.js
xiazeyu: Actually .swal2-popup is themed. The color is careful picked from the theme and then modified by adding more light. It is meant to give enough transition from the background and the pop-up window.
https://github.com/Difegue/LANraragi/pull/620/files#diff-f2a0f5f9f666148bd12d3034d9ddc78ff66904cfbac1aa2a24666dfb7f37d1e7R602-R609

I tried your color and see only little difference. So I changed it to yours.

xiazeyu: For the other themes, at first I thought it is fine to just use the default read. (as other themes are light themes). But now I write some styles for them, and I feels not bad. Have a look.
Sorry that I'm not good at styling themes. So maybe you can have a chance to continuing styling it.

public/themes/g.css
xiazeyu: Sorry that I just struggled into themes and not even notices the lrr.css.

public/js/index.js
Difegue: Good to know that we can assign custom classes to the buttons 👍  
I'm going to merge things as-is for now, and I'll go through the CSS to try and harmonize it all a bit later. (I need to re-review the toasts as well but can't check the frontend atm)  
Thanks for your work!

Difegue (approved):

public/js/server.js
```diff
@@ -23,12 +23,10 @@ Server.callAPI = function (endpoint, method, successMessage, errorMessage, succe
...
             } else {
                 if (successMessage !== null) {
-                    $.toast({
-                        showHideTransition: "slide",
-                        position: "top-left",
-                        loader: false,
+                    LRR.toast({
                         heading: successMessage,
                         icon: "success",
+                        hideAfter: 70000,
```
xiazeyu, line 29: @Difegue Sorry but typo here.
It should be `hideAfter: 7000,`

PR #625, opened 2022-05-04
fix #618

### LINKED ISSUE #618: Fullscreen hooks aren't removed if Escape is used to exit fullscreen instead of F
* Press F to enter fullscreen mode
* Press Esc to exit it
* Scrolling the mouse wheel still changes pages and the max-width of the pages is constrained

@xiazeyu This might interest you 😳

xiazeyu:


public/js/reader.js
```diff
@@ -58,6 +57,81 @@ Reader.initializeAll = function () {
         }
     });
 
+    // Apply full-screen utility
+    // Taken from Fscreen - Fullscreen API, https://github.com/rafgraph/fscreen
```
Difegue, line 61: Couldn't we simply take in the npm package as a dependency here?  
I'd rather not have more code in this 600line behemoth of a JS file if I can help it.

Difegue (changes requested):

public/js/reader.js
xiazeyu: The original codes already have this code, and it is not too long, and it will not change frequently, so I hard coded it here.

Difegue: Well if it's not subject to much change, that's all the more reason to put it away by keeping it in the package I'd say 😅

xiazeyu: Also the Fscreen - Fullscreen API, https://github.com/rafgraph/fscreen don't have a umd-style in their dist folder. So there's no way to use an <script> tag to just import it.

Difegue: Didn't see there wasn't a umd build -- fair enough I suppose.

Difegue (approved):

Issue #631, opened 2022-05-24
nhentai not working


Andor233:
nhentai metafile is not working, and after doing some debugging I found that it is 503.

Is that a Cloudflare problem, or just for me?

I have a proxy(clash) on my router and it works well. No problem for ehentai or else

Pheromir:
nhentai is using cloudflare with the javascript challenge enabled, causing the problem.

Maybe a possible solution could be to create a login-plugin using the cloudflare cookies (depending on how long they're valid), so the js challenge gets skipped. I'm currently testing how long the cookies are valid, but as I just started today, it'll take a while I guess.

Andor233:
> nhentai is using cloudflare with the javascript challenge enabled, causing the problem.
> 
> Maybe a possible solution could be to create a login-plugin using the cloudflare cookies (depending on how long they're valid), so the js challenge gets skipped. I'm currently testing how long the cookies are valid, but as I just started today, it'll take a while I guess.

So with the cookies will work? I used to want to try it but I am not familiar with Perl and HTTP.....

Pheromir:
> So with the cookies will work? I used to want to try it but I am not familiar with Perl and HTTP.....

Seems so, yes. Without cookies I wasn't able to perform a test request using curl to (as an example) nhentai.net/api/gallery/177013, but with the cloudflare cookies (csrftoken and cf_clearance) as well as the matching User-Agent from your Webbrowser I could successfully access the api.
The cookies set in my browser have an expiry set for 1 year, but as the default expiry for  cloudflare is 30minutes, the question is if they're really 1 year valid, which would be nice.

Andor233:
> > So with the cookies will work? I used to want to try it but I am not familiar with Perl and HTTP.....
> 
> Seems so, yes. Without cookies I wasn't able to perform a test request using curl to (as an example) nhentai.net/api/gallery/177013, but with the cloudflare cookies (csrftoken and cf_clearance) as well as the matching User-Agent from your Webbrowser I could successfully access the api. The cookies set in my browser have an expiry set for 1 year, but as the default expiry for cloudflare is 30minutes, the question is if they're really 1 year valid, which would be nice.

Will you release the login plugin?  (By the way I want to know the reason of choosing perl)

Pheromir:
> Will you release the login plugin?
I'll try to create one this weekend and if it works (and the cookie expiry time is not too short) I could make a pull-request to get it merged
> (By the way I want to know the reason of choosing perl)
[...]

PR #633, opened 2022-05-25
Add nHentai-Plugin to bypass CF JS-Challenge

Pheromir:
Added a new plugin to pass cookies and a useragent to LRR to bypass the Cloudflare JavaScript-Challenge.
I specifially named it like that (instead of "nHentai Login" or something like that) so it doesn't get confused with an actual login to the website.

The login plugin is based on the login-example from the documentation and your ehlogin.
for the Metadata-Plugin I changed the author to "Difegue and others" as it is the same way in the E-Hentai Metadata Plugin as an example.

I hope everything's fine, but you might check twice to be sure, as it's the first time I used pearl.

Tested one archive by providing a URL (see screenshot) and one archive by uploading it with the gID in the name ("{id} name") - both worked.

firefox_2022-05-25-18-45-10

### DIFF
--- a/lib/LANraragi/Plugin/Login/nHentai.pm
+++ b/lib/LANraragi/Plugin/Login/nHentai.pm
@@ -20,9 +20,9 @@ sub plugin_info {
         description =>
           "Bypasses the Cloudflare Javascript-challenge by re-using cookies from your browser. Both CF cookies and the user-agent must originate from the same webbrowser.",
         parameters => [
-            { type => "string", desc => "user-agent" },
-			{ type => "string", desc => "csrftoken cookie of domain nhentai.net" },
-			{ type => "string", desc => "cf_clearance cookie of domain .nhentai.net" }
+              { type => "string", desc => "Browser UserAgent string (Can be found at http://useragentstring.com/ for your browser)" },
+			{ type => "string", desc => "csrftoken cookie for domain nhentai.net" },
+			{ type => "string", desc => "cf_clearance cookie for domain nhentai.net" }
         ]
     );
 
--- a/lib/LANraragi/Plugin/Login/nHentai.pm
+++ b/lib/LANraragi/Plugin/Login/nHentai.pm
@@ -1,0 +1,83 @@
+package LANraragi::Plugin::Login::nHentai;
+
+use strict;
+use warnings;
+no warnings 'uninitialized';
+
+use Mojo::UserAgent;
+use LANraragi::Utils::Logging qw(get_logger);
+
+#Meta-information about your plugin.
+sub plugin_info {
+
+    return (
+		#Standard metadata
+        name      => "nHentai CF Bypass",
+        type      => "login",
+        namespace => "nhentaicfbypass",
+        author    => "Pheromir",
+        version   => "0.1",
+        description =>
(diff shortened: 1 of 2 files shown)


Pheromir:
(this should close #631)

lib/LANraragi/Plugin/Login/nHentai.pm
```diff
@@ -0,0 +1,83 @@
...
+        type      => "login",
+        namespace => "nhentaicfbypass",
+        author    => "Pheromir",
+        version   => "0.1",
+        description =>
+          "Bypasses the Cloudflare Javascript-challenge by re-using cookies from your browser. Both CF cookies and the user-agent must originate from the same webbrowser.",
+        parameters => [
+            { type => "string", desc => "user-agent" },
+			{ type => "string", desc => "csrftoken cookie of domain nhentai.net" },
+			{ type => "string", desc => "cf_clearance cookie of domain .nhentai.net" }
```
Difegue, line 25:
```suggestion
			{ type => "string", desc => "cf_clearance cookie for domain nhentai.net" }
```

lib/LANraragi/Plugin/Login/nHentai.pm
```diff
@@ -0,0 +1,83 @@
...
+        name      => "nHentai CF Bypass",
+        type      => "login",
+        namespace => "nhentaicfbypass",
+        author    => "Pheromir",
+        version   => "0.1",
+        description =>
+          "Bypasses the Cloudflare Javascript-challenge by re-using cookies from your browser. Both CF cookies and the user-agent must originate from the same webbrowser.",
+        parameters => [
+            { type => "string", desc => "user-agent" },
+			{ type => "string", desc => "csrftoken cookie of domain nhentai.net" },
```
Difegue, line 24:
```suggestion
			{ type => "string", desc => "csrftoken cookie for domain nhentai.net" },
```

lib/LANraragi/Plugin/Login/nHentai.pm
```diff
@@ -0,0 +1,83 @@
...
+		#Standard metadata
+        name      => "nHentai CF Bypass",
+        type      => "login",
+        namespace => "nhentaicfbypass",
+        author    => "Pheromir",
+        version   => "0.1",
+        description =>
+          "Bypasses the Cloudflare Javascript-challenge by re-using cookies from your browser. Both CF cookies and the user-agent must originate from the same webbrowser.",
+        parameters => [
+            { type => "string", desc => "user-agent" },
```
Difegue, line 23:
```suggestion
              { type => "string", desc => "Browser UserAgent string (Can be found at http://useragentstring.com/ for your browser)" },
```

Difegue (changes requested):
Thank you very much for having taken the time to research this; Code-wise everything looks fine, I only have a few nitpicks on wording for the plugin's user-facing instructions. ✌️

Pheromir:
> Code-wise everything looks fine

Nice.

> I only have a few nitpicks on wording for the plugin's user-facing instructions. ✌️

I kinda thought that would happen, I wasn't sure if the description would have been too long or something like that (except for the of/for mistake). I applied your suggestions. (Writing that because I don't know if you've gotten any notification about it, as the webhook in the discord channel didn't post anything, lol)

Andor233:
Thanks! It works fine for me.

PR #636, opened 2022-06-01
Fix the problem of non-ISO8859-1 decoding garbled

uparrows:
Fix the problem of non-ISO8859-1 decoding garbled .
some characters like chinese or japanese can not be decoded correctly at shinobu logs ,try to fix it .
before:

after:

### DIFF
--- a/lib/Shinobu.pm
+++ b/lib/Shinobu.pm
@@ -304,7 +304,7 @@ sub deleted_file_callback {
 sub add_new_file {
 
     my ( $id, $file, $redis,$outputname ) = @_;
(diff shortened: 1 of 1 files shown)


uparrows:
> Thanks for going in to try and fix this! I knew there was a missing decode somewhere but hadn't gone searching for it yet. 😅
> 
> Got a few remarks to have the PR fit better with existing code.

thanks for your reply , actually i'm a rookie of perl ,😁

Difegue (approved):

Difegue:
Looks good, thanks again!

PR #649, opened 2022-06-26
New metadata plugin for Hitomi.la

doublewelp:
Hello,

I don't know if these contributions are welcome, but I wrote this to import my library into LANraragi. I mostly use Hitomi.la - it's similar to nHentai but they've never bothered to implement Cloudflare or any security checks, so it's much easier to scrape.

This was my first time writing Perl (and also my first time doing a pull request), so let me know if I messed something up. Codewise, I think there's two things that could be improved by someone that actually knows how to write Perl - I had to declare "use utf8" globally for the whole file to get tag emojis working, and the get_tags_from_taglist method looks bloated, but I don't know how to improve on either of these.

Thank you for developing this app.


Difegue:
Thank you! This is definitely welcome and appreciated. I'll try to review this ASAP for integration. 👍

lib/LANraragi/Plugin/Metadata/Hitomi.pm
```diff
@@ -0,0 +1,253 @@
...
+			
+			if($gendertag) {
+				if($male eq 1){
+					$name = $name . " ♂️"
+				}
+				
+				if($female eq 1){
+					$name = $name . " ♀️"
+				}
+			}
```
Difegue, line 168: LRR tends to use namespaced tags, so I think it'd be better to scrap gender emojis altogether and simply prepend the tags with `male:` and `female:`.  
That'd also allow you to remove `use utf8;` from the plugin, which is better!  

In case you still want the gender emojis for your personal use, I believe you should be able to add those back through LRR's Tag Rules system.

```suggestion
		if($male){
			$name = "male:$male";
		}
		
		if($female){
			$name = "female:$name";
		}
```

lib/LANraragi/Plugin/Metadata/Hitomi.pm
```diff
@@ -0,0 +1,253 @@
...
+
+			push( @tags, $name );
+		}
+	}
+	
+	if(defined $json->{"parodys"}) {
```
Difegue, line 174: I think you could indeed simplify the method a bit here: There's a fair bit of copy-pasted code that does the same thing except on a different array and for a different namespace.  

Ideally you should have a separate sub that'd look a bit like this (napkin perl, sorry if the syntax is incorrect at times!):  
```
sub extract_tags{
       my ( $sourcejson, $arrayname, $namespace) = @_;
       my @tags;

      $logger->debug("Extracting $arrayname array");
      my @list = @{ $json->{$arrayname} };
      $logger->debug("Cycling $arrayname array");

      foreach my $tag (@list) {
		my $name = $tag->{$namespace};
		push( @tags, "$namespace:$name");
      }
      return @tags;
}
```  

Which you could then call like so:  
```
push (@tags, extract_tags($json, "parodys", "parody");
push (@tags, extract_tags($json, "artists", "artist");
push (@tags, extract_tags($json, "groups", "group");
push (@tags, extract_tags($json, "characters", "character");
```

lib/LANraragi/Plugin/Metadata/Hitomi.pm
```diff
@@ -0,0 +1,253 @@
+package LANraragi::Plugin::Metadata::Hitomi;
```
Difegue, line 1: It would be nice to add tests for this plugin before integrating it! It helps a lot for monitoring status of metadata providers so we can stay on top when they change formats. (Or add cloudflare 😿)  

There are essentially three test types for plugins:  
* Dead simple test on the namespace to check if it's interpreted by Perl without any issues: This goes in https://github.com/Difegue/LANraragi/blob/dev/tests/modules.t
* Basic tests on the live website, which are basically just calling your `get_tags_from_hitomi` method with a known good gallery ID and verifying that we get the proper tags out of it: See https://github.com/Difegue/LANraragi/blob/dev/tests/plugins.t for examples.
* Offline tests on a sample of data, so that we can always test the plugin logic even if the live website is unavailable. In this case the only sample you'd have to provide would be one of those gigantic hitomi JSONs. See https://github.com/Difegue/LANraragi/tree/dev/tests/samples for the sample folder, and https://github.com/Difegue/LANraragi/tree/dev/tests/LANraragi/Plugin/Metadata for examples.  

I won't ask you to do the third type if it sounds too daunting to you, but it'd certainly make my life easier. 😅  
The first two should be pretty easy though!

Difegue (changes requested):
This is great for a first try! I know writing Perl often drives people away but you nailed it quite well for the most part.  

There are fairly easy solutions to both of your worries, for which I've written some pointers. Don't hesitate if you have any other questions! 👍 The rest is mostly regex improvements.

lib/LANraragi/Plugin/Metadata/Hitomi.pm
```diff
@@ -0,0 +1,253 @@
...
+    shift;
+    my $lrr_info = shift;    # Global info hash
+    my ($gendertag, $savetitle) = @_;    # Plugin parameters
+    my $logger = get_plugin_logger();
+
+    # Work your magic here - You can create subs below to organize the code better
+    my $galleryID = "";
+
+    # Quick regex to get the nh gallery id from the provided url or source tag.
+    if ( $lrr_info->{oneshot_param} =~ /.*\/g\/([0-9]+).*/ ) {
```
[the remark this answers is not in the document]
doublewelp: You're right, this was me getting tunnel vision from my needs. Hitomi URLs can vary along those formats, but "galleries" seems to be a noncanonical shortcut that redirects to one of the others (there's another one for "cg"). I only had to worry about the ID so I kept using the "galleries" links.

I did some tests in regex101 and yours will capture all four cases. I'll update the example link in the description as well.

lib/LANraragi/Plugin/Metadata/Hitomi.pm
doublewelp: Thanks for the tag rules tip, I do prefer emojis here and I had no idea one could postprocess tags like that. I removed the gender emojis option and switched back to namespaces.

doublewelp:
Thank you for the review - encapsulating the JSON array code into its own sub should have been obvious, I completely missed the possibility.

After testing, I had to do a couple of changes to the gallery ID lookup regex - it was picking up every case except the /galleries/ variation because that one is galleries/[id].html instead of [title]-[id].html. I also added the /cg/ one to the tag regex. Everything else should be okay, I think I managed to respond to every raised issue.

Regarding the test cases, I tried to implement all three, but I admit I couldn't actually test them - I'm unsure as to how to run these scenarios. Sorry about all the commits, right now my dev environment is just Notepad++ and creating files by hand in Github... 

Let me know if there's anything else I can improve on.

doublewelp (commented):
Changed regex to support /galleries/ and /cg/ links, every other point implemented as discussed.

Difegue (approved):
Looking good!  
The tests aren't running as-is but it was mostly simple oversights, so I fixed 'em up on my end (just to confirm the actual plugin worked properly!) and will push them after merging this.  

Once again, thanks a bunch for your work on this 🙏

PR #658, opened 2022-07-14
add Timezone support in Docker.

xiazeyu:
Solution from https://github.com/Difegue/LANraragi/issues/114#issuecomment-610612683

Now pass `TZ` env to docker container can change its timezone.

### DIFF
--- a/tools/build/docker/Dockerfile
+++ b/tools/build/docker/Dockerfile
@@ -35,6 +35,8 @@ ENV LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 LANGUAGE=en_US.UTF-8 \
     # Allow Mojo to automatically pick up the X-Forwarded-For and X-Forwarded-Proto headers
     MOJO_REVERSE_PROXY=1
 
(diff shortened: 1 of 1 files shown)


tools/build/docker/Dockerfile
```diff
@@ -35,6 +35,8 @@ ENV LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 LANGUAGE=en_US.UTF-8 \
     # Allow Mojo to automatically pick up the X-Forwarded-For and X-Forwarded-Proto headers
     MOJO_REVERSE_PROXY=1
 
+RUN apk add tzdata
```
Difegue, line 38: This will add an extra layer to the Docker image which is wholly unnecessary -- Alpine packages are installed/uninstalled in the install script here: https://github.com/Difegue/LANraragi/blob/dev/tools/build/docker/install-everything.sh , so this package should go in there instead 👍

Difegue (changes requested):

PR #659, opened 2022-07-16
add Timezone support in Docker

xiazeyu:
Solution from https://github.com/Difegue/LANraragi/issues/114#issuecomment-610612683
Now pass TZ env to docker container can change its timezone.

Closed https://github.com/Difegue/LANraragi/pull/658


Difegue (approved):
👍

PR #684, opened 2022-08-21
Rework s6-overlay config

Guerra24:
So that it works with s6-overlay 3.x

### DIFF
--- a/tools/build/docker/s6/s6-rc.d/lanraragi/finish
+++ b/tools/build/docker/s6/s6-rc.d/lanraragi/finish
@@ -1,0 +1,4 @@
+#!/bin/sh
+kill -15 `cat /home/koyomi/lanraragi/public/temp/shinobu.pid-s6`
+kill -15 `cat /home/koyomi/lanraragi/public/temp/minion.pid-s6`
+sleep 5
--- a/tools/build/docker/s6/s6-rc.d/lanraragi/run
+++ b/tools/build/docker/s6/s6-rc.d/lanraragi/run
@@ -1,0 +1,4 @@
+#!/bin/sh
+cd /home/koyomi/lanraragi/
+export HOME=/home/koyomi
+exec s6-setuidgid koyomi perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
--- a/lib/LANraragi/Utils/Generic.pm
+++ b/lib/LANraragi/Utils/Generic.pm
@@ -142,6 +142,9 @@ sub start_minion {
 
     # Freeze the process object in the PID file
     store \$proc, get_temp . "/minion.pid";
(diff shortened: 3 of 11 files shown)


tools/build/docker/s6/s6-rc.d/lanraragi/run
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+cd /home/koyomi/lanraragi/
+export HOME=/home/koyomi
+exec s6-setuidgid koyomi perl /home/koyomi/lanraragi/script/launcher.pl -f /home/koyomi/lanraragi/script/lanraragi
```
Difegue, line 4: Does this mean the older `services.d/lanraragi` and `services.d/redis` folders are not to be used anymore? I'm guessing the s6 structure changed between versions. 

I was going to say we could remove those in that case, but since I have to maintain a legacy alpine 3.12 build for the time being I can't lol 🫠  

Comments to explain that difference would be much appreciated though!

tools/build/docker/s6/s6-rc.d/lanraragi/finish
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+kill -15 `cat /home/koyomi/lanraragi/public/temp/shinobu.pid-s6`
```
Difegue, line 2: Might be interesting to add those kills to the legacy `services.d/lanraragi` in that case?  
I'm curious to know why those are needed too; seems like the old `s6-svscanctl` was doing a fine enough job by just killing the lrr and redis services. 🤔 (Or maybe it wasn't and I never noticed it!)

Difegue (changes requested):
Thanks for this; Looks like it was more of a pain than I thought it'd be 😓  

Re-adding old-fashioned text file for the PIDs feels like a bit of a step back, but I might just use those in a convenience `npm stop` command for source users since the subprocesses get stuck every now and then. 👍  

Got a few misc. comments.

tools/build/docker/s6/s6-rc.d/lanraragi/finish
Guerra24: From the docs:
> The supervision tree now resides in /run/service, and you should not attempt to stop it when you want to exit the container; more generally you should not attempt to send control commands to the supervision tree. In particular, you should not try to run the s6-svscanctl -t /var/run/s6/services command - it will not work anyway because the supervision tree has changed locations. If you need to exit a container from the inside, without your CMD dying (or without having declared a CMD), run the /run/s6/basedir/bin/halt command instead. 

Those processes would throw exceptions when redis shutdowns since s6 v3 only kills the main process not the childs so we need to kill those too before redis.

Difegue (approved):

Difegue:
Perfect, thanks!

PR #707, opened 2022-11-02
EHPlugin update for the new search system

polak14:
So while i believe the text search should still work fine i went ahead and updated it anyway.

> File searches can no longer be combined with keyword searches or other filters. This search mode will show both normal and expunged galleries. Default tag, language and uploader filters are now automatically disabled for these searches.

>Selecting "Search Expunged Galleries" will now only search expunged galleries in normal searches. (File searches, GID searches and favorite searches will always display both normal and expunged galleries.)

Additionally from what i understand the image search works best (only works?) when using the full image hash not the thumbnail hash.
~~Also we can now search by gID which can be added as an option later.~~
Added an option for this, worked nicely and will fall back to title name if no gId found or search failed.
`[2022-11-02 19:11:23] [E-Hentai] [debug] Found gID: 2365440, Using URL https://exhentai.org?f_search=gid%3A2365440 (gID from archive title)
[2022-11-02 19:11:23] [E-Hentai] [debug] EH API Tokens are 2365440 / 9ff65b23c7`

read more here: https://forums.e-hentai.org/index.php?showtopic=261743&pid=6201235

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/EHentai.pm
+++ b/lib/LANraragi/Plugin/Metadata/EHentai.pm
@@ -134,21 +134,9 @@ sub lookup_gallery {
         #search with image SHA hash
         $URL =
             $domain
-          . "?advsearch=1&f_sname=on&f_sdt2=on&f_spf=&f_spt=&f_sfu=on&f_sft=on&f_sfl=on&f_shash="
+          . "?f_shash="
           . $thumbhash
-          . "&fs_covers=1&fs_similar=1";
-
-        #Include expunged galleries in the search if the option is enabled.
-        if ($expunged) {
-            $URL = $URL . "&fs_exp=1";
-        }
-
-        # Add the language override, if it's defined.
-        if ( $defaultlanguage ne "" ) {
-
-            # Add f_stags to search in tags for language
-            $URL = $URL . "&f_stags=on&f_search=" . uri_escape_utf8("language:$defaultlanguage");
-        }
+          . "&fs_similar=on&fs_covers=on";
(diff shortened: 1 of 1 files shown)


Difegue (changes requested):
Thanks for your work, I wasn't aware of that search change before seeing that PR!  

Could you look into making sure the basic search test doesn't fail? Looks like it's now returning https://exhentai.org/g/2366539/83aa0d7d05 instead of https://exhentai.org/g/618395/0439fa3666/ , which seems a bit wrong.

polak14:
It fails because e-hentai doesn't have the new search system yet, need to use ex right now.

polak14:
After a few updates it looks like its working as expected or even better again. Still waiting for EH to use the new system now.
`[2022-11-09 22:22:37] [E-Hentai] [debug] Using URL https://exhentai.org?advsearch=1&f_sfu=on&f_sft=on&f_sfl=on&f_search=%22%28Kouroumu%208%29%20%5BHandful%E2%98%86Happiness%21%20%28Fuyuki%20Nanahara%29%5D%20TOUHOU%20GUNMANIA%20A2%20%28Touhou%20Project%29%22 (archive title)
[2022-11-09 22:22:37] [E-Hentai] [debug] EH API Tokens are 618395 / 0439fa3666`

Difegue:
Alright, that looks fine -- I guess we can merge this for now, so that people who use ex can move to nightly builds if they want.  

Hopefully the change moves to EH as well before I get to make a release 🙏 Thanks again!

PR #720, opened 2022-12-09
add igneous at cookie

zizzdog:
fix the problem of Ehentai login plugin can not use in non European and American regions

### DIFF
--- a/lib/LANraragi/Plugin/Login/EHentai.pm
+++ b/lib/LANraragi/Plugin/Login/EHentai.pm
@@ -22,8 +22,8 @@ sub plugin_info {
         parameters => [
             { type => "int",    desc => "ipb_member_id cookie" },
             { type => "string", desc => "ipb_pass_hash cookie" },
-            { type => "string", desc => "igneous cookie" },
-            { type => "string", desc => "star cookie (optional, if present you can view fjorded content without exhentai)" }
+            { type => "string", desc => "star cookie (optional, if present you can view fjorded content without exhentai)" },
+            { type => "string", desc => "igneous cookie(optional, if present you can view exhentai outside Europe and America)" }
         ]
     );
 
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Plugin/Login/EHentai.pm
```diff
@@ -16,12 +16,13 @@ sub plugin_info {
...
         namespace => "ehlogin",
         author    => "Difegue",
-        version   => "2.2",
+        version   => "2.3",
         description =>
           "Handles login to E-H. If you have an account that can access fjorded content or exhentai, adding the credentials here will make more archives available for parsing.",
         parameters => [
             { type => "int",    desc => "ipb_member_id cookie" },
             { type => "string", desc => "ipb_pass_hash cookie" },
+            { type => "string", desc => "igneous cookie" },
```
Difegue, line 25: Could you move `igneous` after `star`? Otherwise I'm expecting this'll break upgrades for users that rely on the star cookie.  

It might also be good to mention that this cookie is optional and required only for proxy scenarios.

lib/LANraragi/Plugin/Login/EHentai.pm
```diff
@@ -34,22 +35,22 @@ sub do_login {
 
     # Login plugins only receive the parameters entered by the user.
     shift;
-    my ( $ipb_member_id, $ipb_pass_hash, $star ) = @_;
-    return get_user_agent( $ipb_member_id, $ipb_pass_hash, $star );
+    my ( $ipb_member_id, $ipb_pass_hash, $igneous, $star ) = @_;
```
Difegue, line 38: Likewise, `$igneous` should be after `$star`.

Difegue (changes requested):
👍 Thanks for doing this so quickly! Only have one nitpick but otherwise this is good to go.

lib/LANraragi/Plugin/Login/EHentai.pm
zizzdog: fixed it

Difegue (approved):

Difegue:
Perfect -- Thank you!

Issue #731, opened 2022-12-28
Tooltip broken for "randomly picked"/"new"


nonamethanks:
If you put your mouse over a gallery in the "randomly picked"/"new" archives, the tooltip will appear under the other galleries. This only happens on the left side of the bar (the first few archives depending on your resolution). If you hover on galleries on the right side it works fine.

There's also another issue where the ones that do work, don't overflow so they're cut off by the size of the zone: 
![image](https://user-images.githubusercontent.com/12946050/209886542-c4b480d2-68e2-4c69-a487-bb60fa1ff4ab.png)

In this example you can see the url field is getting cut off because the tooltip is bigger than the zone.

Tested on latest chrome + latest docker. On a 1440p screen the first four archives have broken tooltips.

pastalian:
Rolling back [swiper](https://github.com/Difegue/LANraragi/blob/f3d09bce6e25f20b6f16959e445e7ac651dfe258/package.json#L47) from 8.1.4 to 7.2.0 fixes this issue. It may be due to upstream changes.

Difegue:
The exact CSS parameter that breaks this is [this one.](https://github.com/nolimits4web/swiper/blob/f94322c8be72b1d72a17b081d5527d29d0d11bed/src/core/core.scss#L59)  

It doesn't fix the overflow cutoff though, likely because the tooltips can't escape the flexbox... Not really sure how to fix this further.

PR #738, opened 2023-01-05
Allow the tooltip to be outside of the container

pastalian:
Append the tooltip to the body element so that it is not hidden by other elements.
This fix makes the previous partial fix (26bf9b50997126bdcd75e9f8ad34092df95454e5) unnecessary.

Fixes: #731


Difegue:
Nice, that's much better -- Didn't know the lib allowed for that. Thanks a lot!

PR #742, opened 2023-01-21
Filter more credit pages & move covers to start

Asinin3:
There are a number of scanlators who name their files in a way that puts the credit/translation note pages at the start of a gallery. This PR filters them so they are at the end of a gallery instead.

artist_info - [RedLantern]
9999nhnl - [nhnl]
end_card_save_file - [LunaticSeibah]
9999999* - [MegaFaget]
bumper - [cowboy]
ramble - [degenerate]

And also filters out filenames ending in "note." or "notes." with the period being the file extension, this aims to remove commonly used names for translation notes. I also added a filter to locate filenames ending with "cover." and set them to be the first images in a gallery, so we more reliably find front covers & correct thumbnails for galleries.

I tried to use stricter regex for the commonly used names/characters to avoid false positives, but more may be required.

### DIFF
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -248,9 +248,9 @@ sub get_filelist {
     @files = sort { &expand($a) cmp &expand($b) } @files;
 
     # Move front cover pages to the start of a gallery, and miscelanious pages such as translator credits to the end.
-    my @cover_pages      = grep { /cover\.[^\.]*/i } @files;
-    my @credit_pages     = grep { /^999999|^bumper|^end_card_save_file|notes\.[^\.]*|note\.[^\.]*|artist_info_store_links|credit|999nhnl\.|^group\./i } @files;
-    my @non_credit_pages = grep { !/^999999|^bumper|^end_card_save_file|notes\.[^\.]*|note\.[^\.]*|artist_info_store_links|credit|999nhnl\.|^group\./i } @files;
+    my @cover_pages      = grep { /cover\.[^\.]*$/i } @files;
+    my @credit_pages     = grep { /^999999|^bumper|^end_card_save_file|notes\.[^\.]*$|note\.[^\.]*$|artist_info_store_links|credit|999nhnl\.|^group\./i } @files;
+    my @non_credit_pages = grep { !/^999999|^bumper|^end_card_save_file|notes\.[^\.]*$|note\.[^\.]*$|artist_info_store_links|credit|999nhnl\.|^group\./i } @files;
     @files = ( @cover_pages, @non_credit_pages, @credit_pages );
 
     # Return files and sizes in a hashref
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -248,13 +248,13 @@ sub get_filelist {
     @files = sort { &expand($a) cmp &expand($b) } @files;
 
     # Move front cover pages to the start of a gallery, and miscellaneous pages such as translator credits to the end.
-    my @cover_pages      = grep { /^(?!.*(back|end|rear|recover)).*cover.*/i } @files;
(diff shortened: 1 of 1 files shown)


Asinin3:
Well it looks like I've uncovered a bug. Pushing files to the front of the array created duplicates.
e.g a folder with: cover.jpg 001.jpg became: cover.jpg cover.jpg 001.jpg. I had noticed this kind of duplicate page issue happen before but always thought it was caused by broken cache.

Asinin3:
This still needs to be improved to reduce false positives. The search for cover pages now excludes filenames with "back", "rear" and "end". But I need to make it only match "cover" as a whole word, or a word ending&starting with numbers or special characters. To avoid it matching e.g "discover".

I'm not very good at regex, help would be appreciated.

lib/LANraragi/Utils/Archive.pm
```diff
@@ -247,10 +247,10 @@ sub get_filelist {
 
     @files = sort { &expand($a) cmp &expand($b) } @files;
 
-    # Move any pages containing "credit" to the end of the array.
-    my @credit_pages     = grep { /credit/i } @files;
-    my @non_credit_pages = grep { !/credit/i } @files;
-    @files = ( @non_credit_pages, @credit_pages );
+    # Move front cover pages to the start of a gallery, and miscellaneous pages such as translator credits to the end.
+    my @cover_pages      = grep { /^(?!.*(back|end|rear|recover)).*cover.*/i } @files;
+    my @credit_pages     = grep { /^999999|^bumper|^ramble\.[^\.]*$|^end_card_save_file|notes\.[^\.]*$|note\.[^\.]*$|^artist_info|credit|999nhnl\.|^group\.[^\.]*$/i } @files;
```
Difegue, line 252: Yeah this is starting to feel a bit too specific for me ngl:  

- Aren't the 9999* ones already ending up at the end by default?  
- I'd rather avoid translator-specific lingo since that gets obsolete fast -- `artist_info` and `notes` make sense to me however. (You could also probably keep `end_card_save_file` by just shortening it to `end`? False positives might be too much on that one, although it'd also catch `endcover` which might be convenient)
- As you mentioned it gets exponentially harder to avoid false positives with such a large regex, so I'd rather keep it lean.

Difegue (changes requested):
Thanks for this -- Such a large regex feels a bit unwieldy to maintain imo, but I'm open to accepting a trimmed down version. 👍

lib/LANraragi/Utils/Archive.pm
Asinin3: >Aren't the 9999* ones already ending up at the end by default?

Nope, pages with letters are sorted to be after numbers for instance. There's a ton of galleries where pages are e.g pg001.jpg but then a credit page could start with a number.

>I'd rather avoid translator-specific lingo since that gets obsolete fast -- artist_info and notes make sense to me however. (You could also probably keep end_card_save_file by just shortening it to end? False positives might be too much on that one, although it'd also catch endcover which might be convenient)

I get that, but think we should at least try and do it for popular TL filenames that are overly specific. E.G red lantern or NHNL. Their filenames shouldn't cause any issues with false positives, and it affects a lot of galleries.

lib/LANraragi/Utils/Archive.pm
```diff
@@ -247,10 +247,10 @@ sub get_filelist {
...
+    my @non_credit_pages = grep { !/^999999|^bumper|^ramble\.[^\.]*$|^end_card_save_file|notes\.[^\.]*$|note\.[^\.]*$|^artist_info|credit|999nhnl\.|^group\.[^\.]*$|^(?!.*(back|end|rear|recover)).*cover.*/i } @files;
```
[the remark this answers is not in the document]
Asinin3: Oof totally missed this when making the PR but had it locally.

Difegue (approved):

Difegue:
I didn't test all possible permutations but this seems solid enough, let's get this merged. 👍  

Thanks for taking the time with this! Here's a [Holobyte.](https://www.holopin.io/holobyte/clds00u7p045908mor9poujov)

PR #749, opened 2023-02-03
Chaika plugin adds more tags

kagoromo:
The json from Chaika contains a few more useful info that I wanted to add as tags, namely download URL, gallery ID, category and timestamp. This commit add those, as well as 2 new options. First option allows adding tags without namespace to the 'other:' namespace instead, mimicking the behaviour of E-Hentai. Second option allows tagging galleries found on Chaika with a custom 'source:' tag.

This doesn't fix any existing issue and is mainly for my own use, but I thought others might find it useful as well.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Chaika.pm
+++ b/lib/LANraragi/Plugin/Metadata/Chaika.pm
@@ -19,11 +19,15 @@ sub plugin_info {
         type        => "metadata",
         namespace   => "trabant",
         author      => "Difegue",
-        version     => "2.2",
+        version     => "2.3",
         description => "Searches chaika.moe for tags matching your archive. This will try to use the thumbnail first, and fallback to a default text search.",
         icon =>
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Plugin/Metadata/Chaika.pm
```diff
@@ -43,17 +47,17 @@ sub get_tags {
...
-        ( $newtags, $newtitle ) = tags_from_sha1( $lrr_info->{thumbnail_hash} );
+        ( $newtags, $newtitle ) = tags_from_sha1( $lrr_info->{thumbnail_hash}, $addother, $addsource );
 
         # Try text search if it fails
         if ( $newtags eq "" ) {
             $logger->info("No results, falling back to text search.");
-            ( $newtags, $newtitle ) = search_for_archive( $lrr_info->{archive_title}, $lrr_info->{existing_tags} );
+            ( $newtags, $newtitle ) = search_for_archive( $lrr_info->{archive_title}, $lrr_info->{existing_tags}, $addother, $addsource );
         }
     }
```
Difegue, line 63: Instead of carrying the `$addsource` variable across all subs, you could just check for it here:  

```
if ($addsource ne "") {
    $newtags .= ", source:" . $addsource;
}
```
(I think the plugin system is lenient enough to allow the comma at the start even if no other tags came in, but might be good to check)

lib/LANraragi/Plugin/Metadata/Chaika.pm
```diff
@@ -154,15 +157,49 @@ sub get_json_from_chaika {
...
+            $tag = "other:" . $tag;
+        }
     }
 
-    return ( join( ', ', @$tags ), $json->{"title"} );
+    my $category = lc $json->{"category"};
+    my $download = $json->{"download"} ? $json->{"download"} : $json->{"archives"}->[0]->{"link"};
+    my $gallery = $json->{"gallery"} ? $json->{"gallery"} : $json->{"id"};
+    my $timestamp = $json->{"posted"};
+    if ( $tags ) {
```
Difegue, line 180: People might not want those additional tags, so it'd be good to gate them behind a preference.   

(Maybe this can just use `$addother` as well? I feel like if you want those extra tags you also probably want the E-H style behavior of namespacing everything)

Difegue (changes requested):
Thanks, this indeed looks like it could be useful to some people! 👍  
I have a few suggestions to make the code a bit leaner, if you'd be inclined.  

The matching test at `tests/LANraragi/Plugin/Metadata/Chaika.t` should also be updated. (See CI error log)

lib/LANraragi/Plugin/Metadata/Chaika.pm
kagoromo: That's a good catch, thanks. I will put it after the no tags check on line 65 so adding strings like that should be fine.

lib/LANraragi/Plugin/Metadata/Chaika.pm
kagoromo: That's a good idea. I will add another parameter $addextra in order to make it flexible for users, even at the cost of having to carry it all the way to parse_chaika_json.😅

lib/LANraragi/Plugin/Metadata/Chaika.pm
kagoromo: Actually, it was awkward to test so I decided to carry $addsource across the subs after all.

kagoromo:
The `tests/LANraragi/Plugin/Metadata/Chaika.t` test has passed CI on my fork. Please take a look.

Difegue:
Looks good, thank you very much once again!  

Here are [two](https://www.holopin.io/holobyte/cle4nj8if1108908msgh7fumpr) [Holobytes.](https://www.holopin.io/holobyte/cle4njykh1123108msf1vl6sw8)

PR #753, opened 2023-02-11
Chaika api json

siliconfeces:
So here's a metadata plugin that parses chaika api.json files that are embedded in archives. It has some duplicated code from Chaika.pm, so that's an issue. I can put this shit into the ordinary chaika plugin instead to prevent the copy-paste, but that plugin will get a bit more bloated.

What ya think?

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/ChaikaFile.pm
+++ b/lib/LANraragi/Plugin/Metadata/ChaikaFile.pm
@@ -46,8 +46,8 @@ sub get_tags {
     }
 
     if ( $newtags eq "" ) {
-        $logger->info("No matching Chaika Archive Found!");
-        return ( error => "No matching Chaika Archive Found!" );
+        $logger->info("No Chaika File Found!");
+        return ( error => "No Chaika File Found!" );
     } else {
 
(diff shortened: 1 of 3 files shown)


Difegue:
CI fucked up 'cause the get_temp fix wasn't in yet, but I noticed the new plugin isn't added to `tests/modules.t`.  

I don't mind having some duplicated code in plugins since eventually I want to move most of 'em to a dedicated plugin repo instead of bundling everything in the base server, but that needs some infra work.

siliconfeces:
It fails due to the improvements in the plain chaika plugin as well, so some stuff left to fix.

Difegue:
Looks good, merging! 🎊  

Just for reference (or in case you want to do this for Hentag), you can totally call other plugins' subs from plugin code -- See the [MEMS plugin](https://github.com/Difegue/LANraragi/blob/420238d74adb59be34168fe322c6711e838b32fd/lib/LANraragi/Plugin/Metadata/MEMS.pm#L84) as an example.  

Here are [two](https://www.holopin.io/holobyte/cled870of242608l0rdhx9ewy) [Holobytes.](https://www.holopin.io/holobyte/cled879e2099108l5dtlrb7hk)

PR #756, opened 2023-02-16
Hentag metadata plugin

siliconfeces:
Adds support for embedded hentag metadata files (the ones you get if you pick "Download Metadata"->"Hentag"). Since they default to the filename "info.json" (like many many others), I made this plugin be quite picky about what it accepts to avoid "partially parsing" other file formats.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Hentag.pm
+++ b/lib/LANraragi/Plugin/Metadata/Hentag.pm
@@ -1,0 +1,133 @@
+package LANraragi::Plugin::Metadata::Hentag;
+
+use strict;
+use warnings;
+
+#Plugins can freely use all Perl packages already installed on the system
+#Try however to restrain yourself to the ones already installed for LRR (see tools/cpanfile) to avoid extra installations by the end-user.
+use Mojo::JSON qw(from_json);
+
+#You can also use the LRR Internal API when fitting.
+use LANraragi::Model::Plugins;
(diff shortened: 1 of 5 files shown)


siliconfeces:
Ah, missed that stuff about modules.t. Is there anything left that you want to be fixed in this PR?

I'm working on adding API support as well, but since that may take a while I can make that a separate pull request later'ish.

Difegue:
Nothing else from me in this PR, it just missed the modules.t for a CI check so I tossed it in.   
Since it's green I can just merge this, let me know if you want to make a separate PR or keep working in this one. 👍  

I'm not sure whether the API support should be a separate plugin or melded into this one; Considering we split things up for Chaika, I suppose it makes sense to make this one split as well.  
Although if you end up duplicating 90% of the code, maybe not -- Your call 🤔

siliconfeces:
As long as I can reuse functions across plugins (like in MEMS), I should be able to keep the copy-paste to some boiler-plate-tier minimum. I'll make that a separate plugin (in a separate PR) later for the API lookup version, if keeping things in multiple plugins ends up being stupid it should be easy enough to merge it.

tl;dr, merge away!

Difegue:
Merging this in then, thanks again 👍  

Here are [two](https://www.holopin.io/holobyte/cleeefc19085008jhs48vwbss) [Holobytes.](https://www.holopin.io/holobyte/cleeeg9l9625508l9s08769xi)

PR #758, opened 2023-02-17
[Doc Fix] write-only to read-only

ForsakenRei:
I think it might just be a typeo as we are talking about read-only then the caveat that some folder might need write rights? Or I'm reading it wrong...


Difegue:
Yup that's correct, thanks for noticing!  

Here's a [Holobyte](https://www.holopin.io/holobyte/clebvf4rb168608jvpkgcpcqk).

ForsakenRei:
> Yup that's correct, thanks for noticing!
> 
> Here's a [Holobyte](https://www.holopin.io/holobyte/clebvf4rb168608jvpkgcpcqk).

NP and thanks for the great work!

PR #759, opened 2023-02-18
Added an option to replace existing archive on duplication.

qhua948:
Added a config option "replacedupe" which allows duplicate archive download to continue and replace existing one with the new one.


Difegue:
Looks good! I've reworded the text on the options page a bit but otherwise I'm not seeing anything wrong with this. Thanks!  

Here are [two](https://www.holopin.io/holobyte/clebvputi167308me4yhs20bf) [Holobytes](https://www.holopin.io/holobyte/clebvqjbq180508me8e875kig). 😤

PR #762, opened 2023-02-23
Sort speed increase

djacks6278:
The current sorting algorithm runs hget every time and compares these values, 
which has the disadvantage of heavy load and slow speed.

In addition, there was a huge problem with the existing alignment method.
The problem is that the larger the library, the more exponential it requires.
The load is also severe because hget is executed every operation.
The current method also tries to sort this value against the previous value each time.

To solve this problem,
The key values of the @filtered array and the value of the key value corresponding to the sort key are converted to Hash.
The key sorted based on the value value of this hash has been returned to the @sorted array.

We can process more quickly than before.
Currently, more than 100,000 sortings are often unsuccessful, but if you change them this way, 500,000 sortings can also be successful.

Based on the same performance for 100,000 sortings, traditionally it takes more than 300 seconds and the operation fails, but the improved code takes about 13 seconds.

### DIFF
--- a/lib/LANraragi/Model/Search.pm
+++ b/lib/LANraragi/Model/Search.pm
@@ -370,38 +370,22 @@ sub compute_search_filter {
 sub sort_results {
 
     my ( $sortkey, $sortorder, @filtered ) = @_;
-
     my $redis = LANraragi::Model::Config->get_redis;
-
-    @filtered = sort {
-
-        my $tags_a = $redis->hget( $a, "tags" );
-        my $tags_b = $redis->hget( $b, "tags" );
-
-        # Not a very good way to make items end at the bottom...
-        my $meta1 = "zzzz";
-        my $meta2 = "zzzz";
-
-        if ( $sortkey ne "title" ) {
-            my $re = qr/$sortkey/;
-            if ( $tags_a =~ m/.*${re}:(.*)(\,.*|$)/ ) {
-                $meta1 = $1;
-            }
-
-            if ( $tags_b =~ m/.*${re}:(.*)(\,.*|$)/ ) {
-                $meta2 = $1;
-            }
-        }
-
-        if ($sortorder) {
-            ncmp( lc($meta2), lc($meta1) );
-        } else {
-            ncmp( lc($meta1), lc($meta2) );
-        }
-
-    } @filtered;
-
-    return @filtered;
+	
+    my $re = qr/$sortkey/;
+	
+	my %tmpfilter = map {$_ => ($redis->hget( $_, "tags" ) =~ m/.*${re}:(.*)(\,.*|$)/) ? $1 : "zzzz" } @filtered;
+	
+	my @sorted = map { $_->[0] }
+	  sort { ncmp($a->[1], $b->[1]) }
+	  map  { [$_, lc($tmpfilter{$_})] }
+	  keys %tmpfilter;
+	
+	if ($sortorder) {
+		@sorted = reverse @sorted;
+	}
+	
+    return @sorted;
 }
 
 1;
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Model/Search.pm
```diff
@@ -370,38 +370,22 @@ sub compute_search_filter {
...
-            ncmp( lc($meta1), lc($meta2) );
-        }
-
-    } @filtered;
-
-    return @filtered;
+	
+    my $re = qr/$sortkey/;
+	
+	my %tmpfilter = map {$_ => ($redis->hget( $_, "tags" ) =~ m/.*${re}:(.*)(\,.*|$)/) ? $1 : "zzzz" } @filtered;
```
Difegue, line 377: This needs some comments otherwise I'll shoot myself whenever I need to go back and touch this again.  
I'm pretty sure I got this correctly, but feel free to correct the comment if I'm not. 😅
```suggestion
        # Map our archives to a hash, where the key is the ID and the value is the first tag we found that matches the sortkey/namespace. (If no tag, defaults to "zzzz")
	my %tmpfilter = map {$_ => ($redis->hget( $_, "tags" ) =~ m/.*${re}:(.*)(\,.*|$)/) ? $1 : "zzzz" } @filtered;
```

lib/LANraragi/Model/Search.pm
```diff
@@ -370,38 +370,22 @@ sub compute_search_filter {
...
+	
+	my @sorted = map { $_->[0] }
+	  sort { ncmp($a->[1], $b->[1]) }
+	  map  { [$_, lc($tmpfilter{$_})] }
+	  keys %tmpfilter;
```
Difegue, line 382:
```suggestion
	my @sorted = map { $_->[0] } # Map back to only having the ID
	  sort { ncmp($a->[1], $b->[1]) } # Sort by the tag
	  map  { [$_, lc($tmpfilter{$_})] } # Map to an array containing the ID and the lowercased tag
	  keys %tmpfilter; # List of IDs
```

I guess you could replace `keys %tmpfilter` by `@filtered` here? It doesn't really change much I guess..

Difegue (changes requested):
Thanks, this really helps!  
Non-title sorting was the one part of the search revamp where I kinda ran out of time and defaulted to using the old code, but this looks much leaner in resources.  

This'll still do a redis call for each ID(which is much better than the previous code), but I wonder if it wouldn't be possible to squeeze a bit more speed out of this by bundling all the `hget` calls in a single database call using `multi/exec`... Certainly some food for thought for next time.  

Got a few suggested comments for the code in there, please have a look at 'em and correct me if I'm wrong 🙏

djacks6278:
All comments are correct.
When there is no tag, we use "zzzz" as the default, but if we change it to unicode emoji, we will be able to align it more clearly.

I'd like to add other improvements, but I couldn't spend a lot of time.

How about replacing Redis module to Redis::Fast module for incremental improvement?

lib/LANraragi/Model/Search.pm
djacks6278: The hash value contains {ID: TAG}.
Therefore, this code commonly used hash called %tmpfilter to put the id value sorted based on tag.
I don't know if @filtered will work properly.

djacks6278: Calling multiple variables will result in performance loss, so it seems better to have the same %tmpfilter called.

Difegue (approved):

Difegue:
Looks perfect, thank you! `Redis::Fast` might be an interesting avenue, since it looks to be API-compatible it might not break anything -- But we can think about that at a later time. 😁  

@holopin-bot @djacks6278 
@holopin-bot @djacks6278 holobyte

PR #763, opened 2023-02-24
Hentag.com online lookups plugin

siliconfeces:
This is a first pass on a hentag plugin with online lookups! Big thanks to GodPenguin (Hentag dev) for setting up new API endpoints and generally being very helpful!

The plugin became chunkier than I expected, so feedback and testing would be quite appreciated (and recommended, shit might be buggy).

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/HentagOnline.pm
+++ b/lib/LANraragi/Plugin/Metadata/HentagOnline.pm
@@ -45,6 +45,12 @@ sub get_tags {
     my $logger = get_plugin_logger();
     my $archive_title = $lrr_info->{archive_title};
 
+    # Possible improvement: Detect any currently set hentag URLs, fetch based on that id.
+    # Endpoint: /api/v1/search/vault/id with payload {"ids": [id]}
+
+    # Another possible improvement: Detect any other existing source tags, perform search based on that
+    # Endpoint: /api/v1/search/vault/url with payload {"urls": [url1, url2, ...]}
+
     my $stringjson = get_json_from_api($ua, $archive_title, $logger);
 
     if ($stringjson ne '') {
@@ -67,6 +73,7 @@ sub get_tags {
     return ( error => "No matching Hentag Archive Found!" );
 }
 
+# Fairly good for mocking in tests
 # get_json_from_api(ua, archive_title, logger)
 sub get_json_from_api {
     my ($ua, $archive_title, $logger) = @_;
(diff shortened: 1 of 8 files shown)


siliconfeces:
This passes perlcritic now that sub signatures are allowed, and also handles existing source:hentag.com tags.

lib/LANraragi/Plugin/Metadata/HentagOnline.pm
```diff
@@ -0,0 +1,238 @@
...
+
+# Fetches tags and title, optionally restricted to a language
+sub tags_in_language_from_hentag_api_json($json, $language = undef) {
+    $language =~ s/^\s+|\s+$//g;
+    $language = lc($language);
+
+    # The JSON can contain multiple hits. Loop through them and pick the "best" one.
+    # Possible improvement: Look for hits with "better" metadata (more tags, more tags in namespaces, etc).
+    foreach my $work (@$json) {
+        # Filter out any hits in the wrong language, if requested. Anyone competent in perl, go ahead and golf this shit
```
Difegue, line 188: I think this could be improved upon by mapping `@$json` to a hash with the language as the key, and then looking for `$language` in that hash:  

```
my %json_per_lang = map { LANraragi::Plugin::Metadata::Hentag::language_from_hentag_json($_) => $_ } @$json;

if (exists $json_per_lang{$language}) {
   my ( $tags, $title ) = LANraragi::Plugin::Metadata::Hentag::tags_from_hentag_json($json_per_lang{$language});
   return ($tags, $title);
}
```

See https://perldoc.perl.org/functions/map. I think this code works as-is but haven't tested it :^)

lib/LANraragi/Plugin/Metadata/HentagOnline.pm
```diff
@@ -0,0 +1,238 @@
...
+    if ( $url =~ /https?:\/\/(?:www\.)?hentag\.com\/vault\/([a-zA-Z0-9]*)\/?.*/ ) {
+        return $1;
+    }
+    return;
+}
+
+# Fairly good for mocking in tests
+# get_json_by_title(ua, archive_title, logger)
+sub get_json_by_title($ua, $archive_title, $logger) {
+    my $stringjson = '';
```
Difegue, line 136: should probably be `string_json` to match the other subs

Difegue (changes requested):
Sorry for taking longer to review that one, it was a bit larger after all 😅  

Got a golf suggestion for you.

lib/LANraragi/Plugin/Metadata/HentagOnline.pm
siliconfeces: Aight! It gets a bit trickier as there can be multiple hits for a single language, but it should be solvable.

lib/LANraragi/Plugin/Metadata/HentagOnline.pm
siliconfeces: Fixed!

lib/LANraragi/Plugin/Metadata/HentagOnline.pm
siliconfeces: It's a lot better now I think!

Difegue (approved):

Difegue:
Excellent, it's all good for me. Thanks a lot for going through all the trouble of asking for new API endpoints, let's hope this works out well! 🥳  

it's time... 
@holopin-bot @siliconfeces give-him-the-burg

PR #764, opened 2023-02-24
Koromo multiple artists

siliconfeces:
If the koromo plugin is run on an info.json where the artist field is actually an array you'd get a garbage artist tag (artist:ARRAY-blabla). I'm not sure if it's because of an update to koromo, or if it's because it's a format mixup since info.json is apparently a very popular filename.

 This fixes this by allowing the artist field to be an array. Another way would be to return an error instead, but eh whatever.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Koromo.pm
+++ b/lib/LANraragi/Plugin/Metadata/Koromo.pm
@@ -130,8 +130,19 @@ sub tags_from_koromo_json {
         push( @found_tags, "artist:" . $tag );
     }
 
-    push( @found_tags, "series:" . $parody )     unless !$parody;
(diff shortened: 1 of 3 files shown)


Difegue:
Looks good, merging this! The koromo plugin has historically been used with a ton of other info.jsons since it has one of the most "normalized" formats. It makes sense to try and make it as functional as possible.  
I feel like I should rename it at some point to "catchall info.json plugin that might work for you" or something similar...

Difegue:
Holopin made their bot workflow available to individual users so I don't have to issue holobytes and meme badges by making direct API calls anymore, let's try it!  

@holopin-bot @siliconfeces holobyte

PR #767, opened 2023-02-28
Use Sort::Naturally for titles

siliconfeces:
Restore(?) natural sort for titles.


Difegue:
hmm, I kinda liked saving up on a sort operation by using `zrangebylex`, but it's probably worth taking the small perf hit on this. Thanks!

PR #778, opened 2023-03-17
Avoid complaining about undefined values

siliconfeces:
Fixes a few instances of perl complaining about undefined values.


Difegue:
huh, I actually didn't know about the perl Logical Defined-Or (//) before this! Learn something new every day.

PR #780, opened 2023-03-17
Add category message

siliconfeces:
This PR makes the toast after adding an archive to a category more helpful.
Before:
image
After:
image

More interestingly, it also adds support for backend code (aka perl) to return the success message that appears in toasts.

### DIFF
--- a/lib/LANraragi/Controller/Api/Category.pm
+++ b/lib/LANraragi/Controller/Api/Category.pm
@@ -101,7 +101,15 @@ sub add_to_category {
     my ( $result, $err ) = LANraragi::Model::Category::add_to_category( $catid, $arcid );
 
     if ($result) {
-        render_api_response( $self, "add_to_category" );
+        my $successMessage = "Added $arcid to Category $catid!";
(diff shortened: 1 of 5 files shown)


Difegue (changes requested):
Looks good at a glance -- Could you edit the example API response in the [Docs](https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/api-documentation/category-api.md) as well? Thanks!

siliconfeces:
> Looks good at a glance -- Could you edit the example API response in the [Docs](https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/api-documentation/category-api.md) as well? Thanks!

Done, but not sure how to check what the rendered documentation looks like so I might have fucked something up.

Difegue (approved):

Difegue:
Well, there's no way for me either to know how the doc will look like 'til it lands in dev or master since that's on GitBook's hands, but I'm fairly sure it'll look fine like this.  

Mergin' this 👍

Issue #773, opened 2023-03-12
Implement OPDS-PSE 1.2 (lastReadDate)


Difegue:
PSE 1.2 adds one new parameter, `lastReadDate`, matching the last time a comic was accessed from the server's perspective.  

See https://github.com/anansi-project/opds-pse/pull/6 .

LRR doesn't really have any knowledge of last read times, but this is something that could be added to the DB rather easily alongside the current page when using server-side progress tracking. Not really a priority though 🤔

PR #791, opened 2023-04-16
Add OPDS-PS1 1.2 support for lastReadDate

kuroneko995:
This address https://github.com/Difegue/LANraragi/issues/773 by adding lastReadDate support to the OPDS API endpoints. If server side progress is enabled, lastReadDate is saved whenever progress is updated. OPDS API request will include the lastReadDate field if it's available.

This does not add support for the main LRR app, for example:
- lastReadDate is not rendered in the UI
- lastReadDate is not supported in queries
- Documentation has not been updated

### DIFF
--- a/lib/LANraragi/Controller/Api/Archive.pm
+++ b/lib/LANraragi/Controller/Api/Archive.pm
@@ -206,8 +206,7 @@ sub update_progress {
 
     # Just set the progress value.
     $redis->hset( $id, "progress", $page );
-    my $progressUpdatedTime = time();
-    $redis->hset( $id, "lastreaddate", $progressUpdatedTime);
+    $redis->hset( $id, "lastreadtime", time());
 
     # Update total pages read statistic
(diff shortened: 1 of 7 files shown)


Difegue (changes requested):
This looks perfect, thanks!  

The only request I'd have is that it'd be great if you could add `lastReadDate` to the test for OPDS: It should just be a matter of adding some `lastreadtime` values in the database mocks here: https://github.com/Difegue/LANraragi/blob/dev/tests/mocks.pl 
and then modifying the test output sample here so it matches: https://github.com/Difegue/LANraragi/blob/dev/tests/samples/opds/opds_sample.xml  .

kuroneko995:
I've update the test as requested. Please take another look.

Difegue (approved):

Difegue:
Perfect, thanks once again! 

@holopin-bot @kuroneko995

PR #799, opened 2023-05-03
Allow configuring redis hostname through an env variable

siliconfeces:
Allow setting the redis hostname through an environment variable, using whatever is set in lrr.conf as default. This makes it much easier to run redis in a separate container.


Difegue:
Nice, this works for me. Thanks!

PR #801, opened 2023-05-06
Setup docker-compose for a dev environment

siliconfeces:
Yo!

This adds a docker-compose based solution for bringing up a generic development environment that doesn't rely on the devcontainer standard. Thanks to this, it's easier to use for non-vscode users. It uses the same alpine version as production (unlike the current devcontainer image), **but** keeps redis in a separate container for simplicity.

It's not tested on windows or macos, but I don't see anything inherently incompatible with those platforms.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,28 @@
+---
+# Docker compose file for dev environments.
+services:
+  lanraragi:
+    build:
+      dockerfile: tools/build/docker/Dockerfile-dev
+      context: .
+    volumes:
+      - ./:/home/koyomi/lanraragi
+    ports:
+      - "3000:3000"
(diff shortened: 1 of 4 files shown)


Difegue (changes requested):
👋 This looks good to me, hopefully it can be helpful to others!  

My only minor ask would be to move the `docker-compose.yml` file to the `tools` directory if possible, so that it doesn't linger in the repo root. Thanks!  

(I should probably update https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/extending-lanraragi/index.md with instruction for both this and the codespaces setup at some point..)

Difegue:
Thanks for the documentation, even if it's barebones! 🙏

PR #807, opened 2023-05-16
Use File::Temp rather than homebrew stuff

siliconfeces:
Creating unique temporary files that isn't vulnerable to race conditions is tricky. Delegate that complexity to Somebody Else™


Difegue:
Works for me, it makes for less code as well and File::Temp is already included. Thanks!

polak14:
Importing tags from eze or koromo errors for me with this PR

> Error while fetching tags :
> Error: mkdir /usr/lib/lanraragi/public/temp/plugin/AIrR2CZ_1q: File exists at /usr/lib/lanraragi/script/../lib/LANraragi/Utils/Archive.pm line 303. 

@siliconfeces

PR #811, opened 2023-05-19
Fix broken encoding for titles

siliconfeces:
get_title didn't handle text encoding properly, which caused non-trivial characters getting fucked in the "add to/remove from category" toasts.


Difegue:
Good catch, thanks!

PR #822, opened 2023-07-03
avoid await at top level

URenko:
`await` at top level seems not be allowed in some old browser (Andoird webview in my case).
By using `then` instead, we can enhance compatibility without any loss.


Difegue:
The android webview doesn't support `await`, but front-end ES6 module loading works? ¯\\\_(ツ)_/¯  

Looks good to me if it helps you tho, I'll run some quick testing then merge this.

URenko:
It is not `await` itself now work, but "`await` at top level".
It may relate to second item in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#browser_compatibility

PR #824, opened 2023-07-04
add magazines to koromo

polak14:
the koromo metadata files have a magazine field (e.g COMIC LO..), but the LRR plugin did not add it before. 
Already included in the test samples, so no need to update that. 
oh.. the expected tags do need to be updated in the tests lol


Difegue:
Looks simple enough, thank you.  
I sometimes feel that I could drop `series` from the plugin considering it's basically always `Original Work` these days with the current state of...the "source" of koromo files, but welp might as well leave it in now

PR #827, opened 2023-07-05
'File exists' error when extracting files from archive

polak14:
From what i understand, when `$tmp->filename` is passed as the `$destination` parameter, it includes the filename of the temporary file. 
This caauses :
> Error: mkdir /usr/lib/lanraragi/public/temp/plugin/AIrR2CZ_1q: File exists

when importing tags from koromo or eze for example, because its trying to create a directory with the same name as the temporary file.
 By removing the filename, we ensure $destination is a valid dir path.

### DIFF
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -360,7 +360,7 @@ sub extract_file_from_archive ( $archive, $filename ) {
     my $tmp = File::Temp->new( DIR => $path );
     $tmp->unlink_on_destroy(0);
 
-    return extract_single_file( $archive, $filename, $tmp->filename );
+    return extract_single_file( $archive, $filename, dirname( $tmp->filename ) );
 }
 
 1;
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -357,10 +357,8 @@ sub extract_file_from_archive ( $archive, $filename ) {
     my $path = get_temp . "/plugin";
     mkdir $path;
 
-    my $tmp = File::Temp->new( DIR => $path );
-    $tmp->unlink_on_destroy(0);
-
-    return extract_single_file( $archive, $filename, dirname( $tmp->filename ) );
+    my $tmp = tempdir( DIR => $path, CLEANUP => 1 );
+    return extract_single_file( $archive, $filename, $tmp );
 }
 
 1;


lib/LANraragi/Utils/Archive.pm
```diff
@@ -360,7 +360,7 @@ sub extract_file_from_archive ( $archive, $filename ) {
     my $tmp = File::Temp->new( DIR => $path );
     $tmp->unlink_on_destroy(0);
```
Difegue, line 361:
```suggestion
    my $tmp = File::Temp->tempdir( DIR => $path, CLEANUP => 1 );
```

lib/LANraragi/Utils/Archive.pm
```diff
@@ -360,7 +360,7 @@ sub extract_file_from_archive ( $archive, $filename ) {
...
 
-    return extract_single_file( $archive, $filename, $tmp->filename );
+    return extract_single_file( $archive, $filename, dirname( $tmp->filename ) );
```
Difegue, line 363:
```suggestion
    return extract_single_file( $archive, $filename, $tmp);
```

Difegue (changes requested):
If you use `dirname` on the temporary file created by `$tmp`, aren't you just going to get the initial path we give to File::Temp, aka `my $path = get_temp . "/plugin"`?  

I looked back at the PR (https://github.com/Difegue/LANraragi/pull/807) this change came from, and didn't catch the fact that we moved from creating a temporary folder to a temporary file...  

I believe the easiest way to fix this would be to use File::Temp's [tempdir](https://metacpan.org/pod/File::Temp#tempdir) function instead.

(sorry for the slight delay in reviewing this, it seemed iffy to me so I wanted to spend a bit more time dissecting it, and didn't have said time 🫠)

lib/LANraragi/Utils/Archive.pm
polak14: Right i didn't really understand what it does, this sounds like it works.
Not sure if its my install but i get 

> Error: 'tempdir' can't be called as a method

Difegue: ah, this also needs the `use` statement at the top of the module to add `tempdir` alongside `tempfile`.

Difegue: huh nvm it's already there, I messed up and this doesn't need the OOP interface so:
`my $tmp = tempdir( DIR => $path, CLEANUP => 1 );` 
should work instead.

polak14: Thanks! Seems to work as expected again now.

Difegue:
Perfect, thanks!

PR #833, opened 2023-08-03
Metadata plugin for ksk yaml files

siliconfeces:
This plugin parses the yaml files that exist in _some_ ksk archives. It requires an additional CPAN module, and I haven't tested it on any recent archive. It's a bit old and I've mostly just rebased the branch before opening the PR. At least the tests are green!

Since the metadata doesn't exist in all files, or even the most recent ones, I kind of gave up on this plugin. In other words, no hard feelings if you skip this PR.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Ksk.pm
+++ b/lib/LANraragi/Plugin/Metadata/Ksk.pm
@@ -3,11 +3,11 @@ package LANraragi::Plugin::Metadata::Ksk;
 use strict;
 use warnings;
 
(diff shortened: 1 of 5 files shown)


Difegue:
Looks good! CI is failing on the GH side because it's using a Docker image without YAML::Syck as the base, but that'll fix itself once merged.  

I don't think the plugin itself is that useful since ksk doesn't really include yamls anymore as mentioned, but I still think it's good to have an example available for YAML metadata files people can hack on for their own weird sidecar files.

PR #839, opened 2023-08-28
added hentag-login plugin

Moordp:
Hentag has content that is invisible till you make an account and turn on a setting.
So for making that available I made the hentag-login plugin

At least, it's not available through the website, I don't know if the same applies for the api, but I at least don't get all results even though they are there when the setting is turned on

I have 0 perl experience, so please check if it works before merging


Moordp:
I'll be gone till next monday, will fix this then

PR #840, opened 2023-08-28
Smarter hentag titles

siliconfeces:
This PR improves the success rate of the hentag plugin, when searching by title.

Before this PR, the plugin just picks the first search result if there's multiple. This usually works fine, but for entries in a series this can result in the wrong result being picked.

To solve this, this PR checks which hit seems to be most similar to the requested title. Still not 100% perfect, of course, but a notable improvement. Also, the magic sauce is added as a reusable function as other plugin makers might find it useful.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/HentagOnline.pm
+++ b/lib/LANraragi/Plugin/Metadata/HentagOnline.pm
@@ -14,6 +14,7 @@ use LANraragi::Model::Plugins;
 use LANraragi::Utils::Logging qw(get_plugin_logger);
 
 use LANraragi::Utils::String;
+use String::Similarity;
 
 # Most parsing is reused between the two plugins
 require LANraragi::Plugin::Metadata::Hentag;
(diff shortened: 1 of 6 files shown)


siliconfeces:
It requires a new cpan module, so that might be what causes the failures.

Difegue:
The actions workflow normally uses a nightly container so cpan packages will work in tests post-merge, but I've had CI fail for a while due to the YAML package we added last time so I'm starting to think this isn't a very good way to proceed 🤓  

I'll be modifying the workflow to process the cpanfile before running tests and come back to this PR in a bit.

Difegue:
@siliconfeces fixed, feel free to rebase/merge on top of dev. ✌️ (can't believe that's not something GH allows maintainers to just do yet..)

Difegue:
Thanks! This all looks good to me.  

There technically already is a `remove_spaces` method in the existing Utils module, but I'll probably clean up after this PR to remove it in favor of `Strings::trim`.

PR #841, opened 2023-08-30
Hentag login

Moordp:
Hentag has a controversial content option in the settings that can only be switched on if you are logged in.
so the hentag-login plugin is to set your cookies for allowing getting metadata for those controversial content


Moordp:
not cookie dependend.
unneeded login

PR #856, opened 2023-10-05
Make permission fixing optional (but default)

siliconfeces:
This PR makes it possible to skip most of the chmod/chown fuckery going on when starting the docker version. This can make startup way faster if the content is accessed through NFS. I didn't quite figure out where to document it so I didn't. Permissions is still fixed for redis etc, this ONLY affects the content directory,

The default action is to use the previous behavior.

tl;dr, `-e LRR_AUTOFIX_PERMISSIONS=0` makes startup go wroom wroom.


Difegue:
Sure, looks good to me.  

Environment variables are documented here https://sugoi.gitbook.io/lanraragi/v/dev/extending-lanraragi/architecture#specific-environment-variables-you-can-apply-to-change-lrr-behavior , although since that one is Docker-specific it's a bit different..  
Maybe it should go in the Docker section instead?

Difegue:
Not a super big deal tho so I guess I can merge this as-is. Thanks!

PR #864, opened 2023-10-12
Make stats page faster when file access is slow

siliconfeces:
This is my current attempt at making the stats page faster when using network drives or similar. I try to use redis rather than touching the FS, which is faster but has some issues of it's own.

First, total content size will only consider stuff that's known by LRR, and not the size of the actual directory. Also, you'll need to rescan to get that data into redis.

The archive count will now also show the same count as the default search view, which is a change compared to earlier where you got the actual count of archive files in the content dir.

### DIFF
--- a/lib/LANraragi/Model/Stats.pm
+++ b/lib/LANraragi/Model/Stats.pm
@@ -184,9 +184,17 @@ sub compute_content_size {
     my $redis_db = LANraragi::Model::Config->get_redis;
 
     my @keys = $redis_db->keys('????????????????????????????????????????');
-    my $size = 0;
+
+    $redis_db->multi;
     foreach my $id (@keys) {
-        $size = $size + LANraragi::Utils::Database::get_arcsize($id);
+        LANraragi::Utils::Database::get_arcsize($id, $redis_db);
+    }
+    my @result = $redis_db->exec;
+    $redis_db->quit;
+
+    my $size = 0;
+    foreach my $row (@result) {
+        $size = $size + $row;
     }
 
     return int( $size / 1073741824 * 100 ) / 100;
--- a/lib/LANraragi/Model/Stats.pm
+++ b/lib/LANraragi/Model/Stats.pm
@@ -181,12 +181,13 @@ sub build_tag_stats {
 }
 
 sub compute_content_size {
+    my $redis_db = LANraragi::Model::Config->get_redis;
 
-    #Get size of archive folder
-    my $dirname = LANraragi::Model::Config->get_userdir;
-    my $size    = 0;
-
-    find( sub { $size += -s if -f }, $dirname );
+    my @keys = $redis_db->keys('????????????????????????????????????????');
+    my $size = 0;
+    foreach my $id (@keys) {
+        $size = $size + LANraragi::Utils::Database::get_arcsize($id);
+    }
 
     return int( $size / 1073741824 * 100 ) / 100;
 }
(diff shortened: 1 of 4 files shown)


lib/LANraragi/Utils/Database.pm
```diff
@@ -41,6 +41,7 @@ sub add_archive_to_redis ( $id, $file, $redis ) {
 
     $redis->hset( $id, "name", redis_encode($name) );
     $redis->hset( $id, "tags", "" );
+    set_arcsize( $id, -s $file );
```
Difegue, line 44: I'd add a call to `set_arcsize` to `change_archive_id` in this module as well - This should keep size calculations mostly accurate even for the nutjobs who update their archives in-place.

Difegue: (also wow Utils::Database is getting long I should probably split some of it to a Models::Database at some point)

lib/LANraragi/Model/Stats.pm
```diff
@@ -199,12 +181,13 @@ sub build_tag_stats {
...
+    my $redis_db = LANraragi::Model::Config->get_redis;
 
-    #Get size of archive folder
-    my $dirname = LANraragi::Model::Config->get_userdir;
-    my $size    = 0;
-
-    find( sub { $size += -s if -f }, $dirname );
+    my @keys = $redis_db->keys('????????????????????????????????????????');
+    my $size = 0;
+    foreach my $id (@keys) {
```
Difegue, line 188: It'd probably be a fair bit faster here to use WATCH/MULTI to get the arcsize values for each ID in one redis transaction instead of opening/closing a db connection each time.

Difegue (changes requested):
Looks good to me! Got some minor nitpicks and ramblings.

lib/LANraragi/Model/Stats.pm
```diff
@@ -14,26 +14,8 @@ use LANraragi::Utils::Database qw(redis_decode redis_encode);
...
-                }
-            },
-            no_chdir    => 1,
-            follow_fast => 1
-        },
-        $dirname
-    );
-    return $count;
+    my $redis  = LANraragi::Model::Config->get_redis_search;
+    return $redis->zcard("LRR_TITLES") + 0;    # Total number of archives (as int)
```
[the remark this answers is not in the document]
siliconfeces: Yeah, this was my biggest concern, but unfortunately that was the reason stats grinds to a halt on external mounts and stuff. Could always fetch that stat through javascript and at least get the rest of the stats showing quickly.

lib/LANraragi/Model/Stats.pm
siliconfeces: Oops, missed that every call to get_redis created a new connection.

lib/LANraragi/Utils/Database.pm
siliconfeces: Fixed!

siliconfeces:
That was just a temporary failure due to chaika, right?

Difegue:
Looks like it - I've queued a rerun.  
(And I need to look at those new commits)

lib/LANraragi/Model/Stats.pm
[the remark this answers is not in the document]
Difegue: Yeah, I don't think it's worth adding extra JS cruft just for a number (although people might yell at me for it so we'll see?)  
Works for me for the time being anyway. 👍

PR #868, opened 2023-10-14
Reuse generate_archive_list to make code more DRY

siliconfeces:
The "Modify Categories" page was rolling some custom code for listing all archives. This PR changes that to reuse the same function as in "Batch Operations", which has the nice side effect of being faster as well.

### DIFF
--- a/lib/LANraragi/Controller/Category.pm
+++ b/lib/LANraragi/Controller/Category.pm
@@ -21,25 +21,18 @@ sub index {
 
     $redis->quit();
 
-    #Then complete it with the rest from the database.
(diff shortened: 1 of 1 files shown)


Difegue:
Looks good. In hindsight I'm not quite sure why the categories page still loads the IDs server-side while Batch does an API call, but I'm constantly flip-flopping on which approach is best so might as well keep it like that for now..  

I'll need to make yet another "list all archives and checkbox select some" page for tankoubons soon so maybe I'll make up my mind then

Difegue:
(At least now they both end up using the list code with MULTI so that's good speedwise as mentioned.)

PR #871, opened 2023-10-17
Remove duplicate tags when updating DB

### LINKED ISSUE #870: Tag duplicates are not properly blocked
Duplicate tags are not blocked everywhere, such as when using batch operations or when doing an API call to update metadata.

(This bug report is a bit shitty but I'm planning on fixing this myself)

siliconfeces:
Fixes #870


Difegue:
👌 Works for me. Thanks!

PR #872, opened 2023-10-19
(chore): ajax cache to reduce server load

ndbiaw:
normal, ajax will add timestamp at the end of request. this config remove it

### DIFF
--- a/public/js/common.js
+++ b/public/js/common.js
@@ -308,6 +308,7 @@ LRR.getImgSize = function (target) {
     $.ajax({
         async: false,
         url: target,
+        cache: true,
         type: "HEAD",
         success: (data, textStatus, request) => {
             imgSize = parseInt(request.getResponseHeader("Content-Length") / 1024, 10);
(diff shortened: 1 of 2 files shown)


Difegue:
👋 Wouldn't this lead to users not getting up-to-date data when they modify metadata for their archives, which changes the internal indexes/the result of the search API call?  

There's already serverside cache for searches, so I don't think this would be very useful 🤔

ndbiaw:
> 👋 Wouldn't this lead to users not getting up-to-date data when they modify metadata for their archives, which changes the internal indexes/the result of the search API call?

cache: true doesn't actually change anything, cache: false just adds a timestamp at the end of the URL to prevent content from being cached if the appropriate header is present. User can still get up-to-date results, this only helps if the user uses the header to set the cache but ajax always adds timestamp to the end of the URL so making browser cache useless. This is useful for reducing the request to the server.

> There's already serverside cache for searches, so I don't think this would be very useful 🤔
This change reducing the request to the server if the user sets the appropriate header, so I don't think server-side caching is really relevant, because if the request is cached it won't make the request to the server.

Difegue:
Thanks for the explanation - Don't see anything wrong with merging this in that case. Thanks!

PR #878, opened 2023-10-25
Fix Redis password authentication

TomaSajt:
I was packaging the app for Nixpkgs (merged now, btw) when I realised that the Redis password wasn't working correctly.

I looked up the docs of the Minion Redis Backend (https://metacpan.org/pod/Minion::Backend::Redis), and it turns out you need an `x:` before the password (or atleast that's how it worked for me).

Also, there was another place (the main plugin registering part) where the password wasn't even trying to be passed over to Minion. This was the main source of the errors it was throwing.

This PR addresses this by prepending a `x:` to the password, when given, and also providing a password at the other location.


Difegue:
Looks about right - it's kinda embarrassing to have missed that considering I worked on the minion backend side as well 😅 goes to show how often I test with redis passwords

Thanks for the fix!

PR #879, opened 2023-10-25
ToC metadata for archive

EfronC:
This is just a proposal for a feature to add a ToC to files that could require them (Compilations of works). I'm not sure how much could this synergize with the platform, so I didn't put a lot of effort in it, I just kept it simple so it is easier to test. This feature comes from the need to know which titles are included on a certain zip file, and probably get direct links to every one of them and index them inside the reader view.

Basically this adds a "toc" key to the archive table in Redis, which would save the whole object as a JSON string, which we will be able to enconde/decode as needed to provide it. I added 2 methods to Archive.pm, one to update the toc data, and the second to remove an element from it, and on the same note, I also added 2 methdos to database.pm, one is to get the ToC already ordered(Because it is not saved this way, every new key is appended), this one in case we want to evade redundance from writting those lines every time we want to decode the ToC and keep it organized, and the second one is to get the list of titles of the ToC(The values of the hash), so if eventually this could be used to include those names as aliases for the archive in the search, well to have a practical way to get that data.

Lastly, just for the purpose of testing, I added a context menu(That only appears if you're logged in) in the reader view, in the pop up where you can see the list of pages as thumbs, which include 3 options: add, edit, and delete a toc content(add and edit are basically the same BTW). Those are for testing only, since I'm not sure which would be a good way to implement the ToC in the Frontend, so maybe it could be removed later.

image

This is just to present the feature, honestly I'm not sure if is a good feature on first instance, but well, if it might be a good addition, we can check to work it a little more.


Difegue:
I took a cursory look - Won't focus on this until post-0.9.0, but I think the idea is sound!  

There's some good combo potential with meta-archives/tankoubons UI-wise, as you could end up with a tiered combobox that'd combine both:  

```
- Tank Archive 1
  - ToC page 1
  - ToC page 20
  - etc
- Tank Archive 2
- Tank Archive 3

```  

I'm not sure a context menu is the best solution, I think buttons on the main UI to add/remove (add becomes edit) might be faster.

EfronC:
> There's some good combo potential with meta-archives/tankoubons UI-wise, as you could end up with a tiered combobox that'd combine both:

Yes, depending in how Tanks affect the search, it could be enhanced further with the ToC for aliases for a single tank object. I think I can imagine many scenarios where this could be useful in how I expect to use Tanks actually.

> I'm not sure a context menu is the best solution, I think buttons on the main UI to add/remove (add becomes edit) might be faster.

Yeah, me neither, for me it's too hidden this way, I just put that context menu there for testing while working on it, I was thinking of removing it before creating this PR because of that, but I left it there since otherwise there would not be a good way for others to check it.

Difegue:
It took me an eternity to get back to this, so I ended up just building the entire UI 😳  

I moved away from the context menus and made it so that the page overlay is split by ToC/Chapters, alongside new buttons on each page to add ToC entries and set the given page as the archive thumbnail. (Which is more functional and better UI than the bundle of buttons in Admin Options)   

image  

This relies on a nested JS object which I'm hoping will make Tank integration easier.

PR #886, opened 2023-10-31
Delete archive error handling

siliconfeces:
Time to drown you in work again!

I noticed that "delete archive" was a bit optimistic and tended to always report success. This fixes that, and also ensures that the archive file was actually deleted.

### DIFF
--- a/lib/LANraragi/Utils/Database.pm
+++ b/lib/LANraragi/Utils/Database.pm
@@ -248,7 +248,7 @@ sub delete_archive ($id) {
         # Delete the thumbpages folder
         remove_tree("$thumbdir/$subfolder/$id/");
 
-        return $status?$filename:"0";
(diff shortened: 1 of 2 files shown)


lib/LANraragi/Utils/Database.pm
```diff
@@ -248,7 +248,7 @@ sub delete_archive ($id) {
         # Delete the thumbpages folder
         remove_tree("$thumbdir/$subfolder/$id/");
 
-        return $filename;
+        return $status?$filename:"0";
```
Difegue, line 251:
```suggestion
        return $status ? $filename : "0";
```

Difegue:
Thanks -- This one is small enough I can't consider it too heavy a workload to review.  
Just added some spacing to the ternary expression so it's more readable but otherwise good to go!

PR #894, opened 2023-11-03
chore: make modern theme more "modern"

ndbiaw:
Makes the default theme more "modern". Inspired by Reddit's color palette.

New:

image

Old:

image


Difegue:
hmmm - nope, don't like it!

PR #900, opened 2023-11-06
Set missing ghostscript device value

TomaSajt:
I was testing out pdf conversion on WSL (running NixOS), and ghostscript was complaining that the xserver wasn't running. I found that one line in the pdf info querying code was missing an explicitly specified device, so I'm guessing the xserver was the fallback, which WSL didn't have.

Also tested it on Linux, when I viewed the pdf on the website, the server's monitor started opening many pdf-viewer windows just for a split second, which, because I am using `i3`, was continuously messing up the window layout.

This simple change should fix both of these issues.

I didn't notice any quality regressions (the quality was not that great before and after the fix, but maybe that's just NixOS)


Difegue:
Thanks - I'd encountered this in recent WSL testing but didn't have time to figure out the issue.  
GS PDF quality has never been very good, no worries there! 👍

Difegue:
wow is this thing actually broken for real now  

@holopin-bot @TomaSajt cmon

PR #903, opened 2023-11-06
Use password directly when connecting to Redis

TomaSajt:
It turns out I didn't find every issue with using Redis with a password in a previous PR.
I found that it would sometimes throw `NOAUTH` errors, but only when finishing a Minion job(??) The main thing is that file upload and downloaders were not working at all because of this.
After a little more time than I dare to admit, I found a solution: don't use `auth`, just use the `password` key in the constructor.
I don't know a lot of Perl, so after a bit of searching, I found a way to do conditional key-value pairs. If there's a better way, please do tell me.


Difegue:
huh, I didn't know you could do conditional kvps in Perl either! god bless larry wall for his gift of wacky syntaxes.  

@holopin-bot @TomaSajt holobyte

PR #917, opened 2023-11-25
Allow filtering on number of read pages

owldyn:
Would like to be able to search for not started (but not flagged as "new") and barely started galleries, this should allow that. Figured I would just extend the page number filtering since it can use the same logic.


Difegue:
Looks good and tests well - No qualms from me on merging this 👍 

I'd been thinking of adding an "On Deck" section to the carousel at some point, this'd probably be useful. Thanks!  

@holopin-bot @owldyn

Difegue:
(oh hey holobytes work again that's good to know)

PR #940, opened 2024-01-11
Update the download URL of Tachiyomi extension

Seflyx:


Difegue:
Thanks -- I've only barely followed the recent Tachi extension discourse so this is helpful!

PR #941, opened 2024-01-13
Fix eH login script so that downloader can download galleries hidden behind a warning

f8upd8:
Add nw=1 cookie to skip offensive warning screen


Difegue:
Looks good, thanks!

PR #944, opened 2024-01-28
add plugin (Hentai@Home)

sakuyamaij:
a plugin that can collects metadata embedded into your archives by HentaiAtHome Downloader's galleryinfo.txt files.


sakuyamaij:
Thanks!

Difegue (changes requested):
Sorry for the delay in reviewing this - I'd just fixed the package name to see if compilation passed and left it at that.  

Plugin itself looks fine -- Could you add some tests for it? You can look at https://github.com/Difegue/LANraragi/tree/dev/tests/LANraragi/Plugin/Metadata for examples.  
A sample galleryinfo.txt should also be added in https://github.com/Difegue/LANraragi/tree/dev/tests/samples and used by the tests.

PR #945, opened 2024-01-29
Update Tachiyomi docs

RePod:
With added Keiyoushi bias.


Difegue:
Looks good. Thanks a lot for the new screenshot as well!

PR #963, opened 2024-04-04
Enable SOCKS proxy support

ypoluektovich:
By default Mojolicious doesn't pull the Socks module that's required for SOCKS5 proxy use.


Difegue:
Looks accurate according to https://docs.mojolicious.org/Mojo/UserAgent#DESCRIPTION . Thanks!

Salvora:
Current documentation covers only the http(s) proxy setup for docker environments. Is this not possible to set up when deploying with docker?

Issue #943, opened 2024-01-21
[Feature request] save japanese title from Chaika.moe metadata


sakuyamaij:
E-Hentai v.2.5.2 metadata plugin has a great opinion called
**Save the original title when available instead of the English or romanised title**.

Will it be possible to add this opinon to Chaika.moe v.2.3.1 metadata plugin?

I found almost every archives have "title_jpn" in chaika's api, so I guess it is possible?

Difegue:
Sounds easy enough, tagging as volunteers.

PR #967, opened 2024-04-17
(#943) Save japanese title from Chaika.moe metadata

BWsix:
This PR adds option `Save the original title when available instead of the English or romanised title` in Chaika metadata plugin, as proposed in https://github.com/Difegue/LANraragi/issues/943:

### Screenshots

Settings page:
Screenshot from 2024-04-17 16-35-11

Before enabling the option:
Screenshot from 2024-04-17 16-48-19

After enabling the option, I think it's working as intended (the first one has jp title and the second one does not):
Screenshot from 2024-04-17 16-40-32

### Notes

This is my first time coding in perl, but I literally copied how EHentai plugin did <https://github.com/Difegue/LANraragi/blob/dev/lib/LANraragi/Plugin/Metadata/EHentai.pm> so it should be fine..?

Before commit, I ran `npm run critic` and the file I touched passed the lint:
image

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Chaika.pm
+++ b/lib/LANraragi/Plugin/Metadata/Chaika.pm
@@ -32,6 +32,9 @@ sub plugin_info {
             },
             {   type => "string",
                 desc => "Add a custom 'source:' tag to your archive. Example: chaika. Will NOT add a tag if blank"
+            },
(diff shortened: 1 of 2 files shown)


Difegue:
Looks good to me. Sorry for the delay, this PR completely slipped by my notifications..

PR #968, opened 2024-04-20
Render HTML for summaries

psilabs-dev:
Support HTML rendering features for displaying summaries, such as new lines, font styles, hyperlinks, etc. Should be backwards compatible with text rendering.

There are some security concerns, mainly if scripts are found in the summary. In case one archive points to metadata containing a malicious summary. I think how to resolve this is up for debate; whether to sanitize client-side or server-side or both. But I think this would be a better visual experience.

not too familiar with JS

image

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -125,7 +125,7 @@ Reader.initializeAll = function () {
 
             if (data.summary) {
                 $("#tagContainer").append(`<div class="archive-summary"/>`);
-                $(".archive-summary").text(data.summary);
+                $(".archive-summary").html(data.summary);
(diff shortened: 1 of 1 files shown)


Difegue:
Yeah I explicitly disabled this for security concerns; I'd rather not encourage people putting html in their summaries either since it means third-party clients would also need to parse/render HTML which is a massive can of worms.

psilabs-dev:
Fair enough. Pixiv illustration summaries use html so I'm more motivated to push this haha

There are some ways to address those issues, let's just put a pin in this for now

PR #969, opened 2024-04-20
automatically add summary for archive upload and edits

psilabs-dev:
Prior to PR, summary is not automatically added during archive file uploads (when metadata plugin is configured to run automatically) (addressed by Plugins.pm), and summary data is not automatically filled in to the summary box when editing an archive and calling metadata plugins (edit.js). This PR attempts to address that.

Contains edits (readme.js) from #968, a separate issue.

I'm not sure how do deal with rhombus git flows so I'm just keeping it linear now

### DIFF
--- a/public/js/edit.js
+++ b/public/js/edit.js
@@ -174,6 +174,15 @@ Edit.getTags = function () {
                 });
             }
 
+            if (result.data.summary && result.data.summary !=="") {
+                $("#summary").val(result.data.summary);
+                LRR.toast({
+                    heading: "Archive summary changed to :",
+                    text: result.data.summary,
+                    icon: "info",
+                });
+            }
+
             if (result.data.new_tags !== "") {
                 result.data.new_tags.split(/,\s?/).forEach((tag) => {
                     // Remove trailing/leading spaces from tag before adding it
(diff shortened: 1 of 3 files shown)


public/js/edit.js
```diff
@@ -174,6 +174,15 @@ Edit.getTags = function () {
                 });
             }
 
+            if (result.data.summary && result.data.summary !=="") {
+                $("#summary").val(result.data.summary);
+                LRR.toast({
+                    heading: "Archive summary changed to :",
+                    text: result.data.summary,
```
Difegue, line 181: Since summaries can be pretty long it's probably better to just mention "Archive summary updated!" without the text.  

Mentioning the value is kinda superflous already since users can just like, see it in the text box..

public/js/reader.js
```diff
@@ -125,7 +125,7 @@ Reader.initializeAll = function () {
 
             if (data.summary) {
                 $("#tagContainer").append(`<div class="archive-summary"/>`);
-                $(".archive-summary").text(data.summary);
+                $(".archive-summary").html(data.summary);
```
Difegue, line 128: As mentioned in the other PR, probably better to remove this

Difegue (changes requested):
Thanks, I'd missed those two!

Difegue (approved):

PR #973, opened 2024-05-07
Change YAML::Syck to YAML::PP

shinji257:
This parser handles the longer \u unicode sequences without any changes to existing LANraragi code.  It is a literal drop in swap.


Difegue:
Ace; I wasn't expecting both libraries to have the exact same API.. I don't think there was any specific reason to go with Syck and the tests pass, so I'm happy to merge this.   
Thanks!

PR #977, opened 2024-05-29
Add web app manifest

jacobgkau:
This allows installing the website as a progressive web app (PWA) on mobile devices and certain web browsers. That's useful to be able to use the app without having the URL bar/tab bar getting in the way.

Relevant docs:
- Mozilla: https://developer.mozilla.org/en-US/docs/Web/Manifest 
- Google: https://web.dev/learn/pwa/web-app-manifest

I've tested this on two Android devices (Android 9 and 14).

### DIFF
--- a/public/app.webappmanifest
+++ b/public/app.webappmanifest
@@ -1,0 +1,14 @@
+{
+  "name": "LANraragi",
+  "short_name": "LANraragi",
+  "start_url": "/",
+  "display": "standalone",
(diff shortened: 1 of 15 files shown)


Difegue:
Looks good to me. Does it work fine when you start browsing through the web front? (ie edit, reader, etc)  
Since we're not a singel-page app, I always thought browsers would just start popping additional windows for everything if we declared ourselves as a PWA.

PR #984, opened 2024-06-09
Fixed #983 The FAKKU plugin now works again.

### LINKED ISSUE #983: FAKKU plugin has stopped working
**LRR Version and OS**
LRR 0.9.0 
Windows 11

**Bug Details**
The metadata plugin for FAKKU has stopped working, both in automatic and manual mode. The automatic mode stopped working a few months back but the manual mode was still fine until a few days ago, I suspect is because they made some modifications to the website. 
I know the plugin has been kinda wonky since the beginning and the suggested method was to use the chaika plugin, but they have stopped uploading any FAKKU related content for more than a year so is not really an option anymore.

**Matching Logs**
Log attached because of character limit
[fakkupluginlog.txt](https://github.com/user-attachments/files/15746269/fakkupluginlog.txt)

**Screenshots**
image

Nixis198:
Simple fix for the FAKKU plugin problem found in #983 . Literally just a one character fix.
From my testing, everything works the way it used too before FAKKU updated their website.
I don't really ever make pull requests so I don't really know what else to put. Just let me know and I will do my best to fix it.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Fakku.pm
+++ b/lib/LANraragi/Plugin/Metadata/Fakku.pm
@@ -160,7 +160,7 @@ sub get_search_result_dom {
 
     my $res = $ua->max_redirects(5)->get($URL)->result;
 
-    # $logger->debug( "Got this HTML: " . $res->body );
+    $logger->debug( "Got this HTML: " . $res->body );
 
     return $res->dom;
 }
@@ -175,7 +175,7 @@ sub get_dom_from_fakku {
 
     my $html = $res->body;
 
-    # $logger->debug( "Got this HTML: " . $html );
+    $logger->debug( "Got this HTML: " . $html );
     if ( $html =~ /.*error code: (\d*).*/gim ) {
         $logger->debug("Blocked by Cloudflare, aborting for now. (Error code $1)");
         die "The plugin has been blocked by Cloudflare. (Error code $1) Try opening FAKKU in your browser to bypass this.";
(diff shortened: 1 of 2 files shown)


Nixis198:
@pandafeliz noticed that my fixed didn't work with automatic mode. The latest commit fixes that.

Nixis198:
I think the tests are out of date. They are from before FAKKU updated their website.
I'm going to try to get LANraragi setup for development on my computer so I can test the tests without having to deploy to Github.

Difegue:
Yeah, it's likely the sample HTML needs to be updated.

Nixis198:
I discovered a major issue with the plugin.
It only works if you are logged in on FAKKU using the login plugin providing your FAKKU sid.
I am working on changing what the plugin looks for to get the tags

Nixis198:
I did update the test HTML file but currently the plugin will fail.

Difegue:
Hasn't that been the case for a while? The login plugin is there to help circumvent the cloudflare block.

Nixis198:
> Hasn't that been the case for a while? The login plugin is there to help circumvent the cloudflare block.

The reason for it to fail isn't directly caused by Cloudflare. The plugin looks for the "Suggest More Tags" button on the page to find the tags. If you are not logged in that button doesn't show up, so the plugin errors out.

With Cloudflare being an issue, I will change the plugin to throw an error if you are not logged in. I will also change the test to simulate being logged in, so it should pass.

Nixis198:
I added checks to make sure the user is logged in to FAKKU. The plugin will return an error if the proper cookie if not found.
Because FAKKU accounts are free to make, I don't see this as a major setback for users.
I did update the test HTML file to better represent the current FAKKU site (simulating that a user is logged in).
I don't think any other changes are needed for this plugin.

Difegue:
All looks good here. Thanks again!

PR #985, opened 2024-06-09
Log error message if open_filename fails

siliconfeces:
Add some logging if opening archive fails.
I can add tests if you want, but it's just a bit of logging so I decided to YOLO.


Difegue:
Looks good, thanks 👍

PR #986, opened 2024-06-11
Fixes #980 Adds Plugins to Persistent volumes in Dockerfile

### LINKED ISSUE #980: uploaded plugins gone if docker container recreated
**LRR Version and OS**
Docker
LANraragi v0.9.0

**Bug Details**
Uploaded plugins are store in `home/koyomi/lanraragi/lib/LANraragi/Plugin` along with the default plugins. This directory is come from the image.
As a result, when docker container recreate(e.g. for version upgrading, env vars updating), the directory will be restored, the uploaded plugins would gone. I have to upload the plugins, and configurate them again.

Expected: user uploaded plugins are put in an independent directory, so we can mount a local path to it.

**Matching Logs**
Post any logs you have encountered when reproducing the bug.  
If possible, switch the app to Debug Mode in order to have more logs.
No Logs.

**Screenshots**

Nixis198:


Difegue:
I wonder how this will work with updates -- If you install a new version of the server that updates the bundled plugins, will they overwrite the ones in an existing volume correctly?  
That's the main reason I always left the plugin directory nonexposed, but I've never tested this stuff in detail..

siliconfeces:
What about scanning for plugins recursively in the path? In that case, one can split it up fairly easily without causing too much issues with existing installs. Leave the bundled plugins in the base `Plugins` directory and store uploaded plugins in  `Plugins/Uploaded`. The latter can be exposed as a docker volume/path if you want persistent uploaded plugins.

It will also be possible to install sets of plugins through additional mounts like `Plugins/MyPlugins`.

Difegue:
The issue with that is that Perl uses the folder structure to do namespace lookups, so uploaded plugins would need a new namespace.. Although thinking on it there's no reason we shouldn't be able to bodge this with symlinks in the container

Difegue:
I've hijacked the PR to add a `Sideloaded` folder that Plugin uploads will now land into.  
Seems like there's no issue having a bunch of differently namespaced files in that random folder, so it should now carry over between Docker installs just fine.

PR #987, opened 2024-06-12
Add patterns for timestamp tag rendering

chierinyan:
Render tags as timestamp when namespace starts with `date` or `time`, as discussed in https://github.com/Difegue/LANraragi/discussions/981

> First time trying to create a PR... Please forgive me if I made any stupid mistakes :(

PR #989, opened 2024-06-16
Changed the Koushoku.yaml plugin to add the "Released" tag.

Nixis198:
Adds the option to add the "Released" tag using the Koushoku.yaml plugin.
Needs to be enabled in the plugin settings.


Difegue:
Looks good to me. Should `released:` be prefixed by `date_` or `time_` so it can leverage https://github.com/Difegue/LANraragi/pull/987 in the UI?

Nixis198:
Updated to include the date prefix.

Issue #982, opened 2024-06-07
Error while processing archive.


swithun-liu:
<!-- hol up senpai have you checked out the GH Discussions first ? 
Your issue/idea might've already been reported by someone else! 🐱‍👤
https://github.com/Difegue/LANraragi/discussions
->

**LRR Version and OS**
Please tell which version you're running, as well as your OS and installation method. (Windows, Docker, Native)

Docker (arm chip)

**Bug Details**
Please detail the bug here. 

 Error while processing archive.
(The file couldn't be moved to your content folder!)

**Matching Logs**
Post any logs you have encountered when reproducing the bug.  
If possible, switch the app to Debug Mode in order to have more logs.

**Screenshots**
<img width="1248" alt="image" src="https://github.com/Difegue/LANraragi/assets/42160297/f5451c55-4a90-4bea-a264-9afe78633a2b">

Difegue:
Need more logs/info here I'm afraid - check write permissions on the content folder maybe

PR #990, opened 2024-06-17
Show syscall error message if file move failed

siliconfeces:
Fairly minor change that shows why moving uploaded files fail. Should make stuff like #982 easier to troubleshoot.

### DIFF
--- a/lib/LANraragi/Model/Upload.pm
+++ b/lib/LANraragi/Model/Upload.pm
@@ -110,12 +110,10 @@ sub handle_incoming_file {
 
     # Move the file to the content folder.
     # Move to a .upload first in case copy to the content folder takes a while...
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Model/Upload.pm
```diff
@@ -113,8 +113,12 @@ sub handle_incoming_file {
     move( $tempfile, $output_file . ".upload" );
```
Difegue, line 113: If the move were to fail due to permission errors on the content folder, I believe you should check the output of this first move operation instead, as it's the one that actually places the file in the correct spot.  

We could probably return early here and not bother with the second move if this one fails too.

Difegue (changes requested):

siliconfeces:
Made it a bit more "perlthonic"

PR #991, opened 2024-06-21
Fix #975 Preload image size asynchronously

### LINKED ISSUE #975: Respect preload setting when resizing in reader enabled
Currently it seems that images are not preloaded and/or pre-resized when resize in reader is enabled. Advancing the page results in a noticeable lag every time. It would be a significantly better user experience if the same number of "preload" images were resized ahead of time (assuming they require resizing) to reduce this reading lag. Another small thing to note is that it seems resized images are not kept in the cache after exiting the reader and returning, but it would be better to just leave them there.

anonstash:
Was a pretty low hanging fruit so tried to do it in the simplest way possible. Adds an async version of the `getImageSize` function and calls it in `Reader.loadImage`. Other line changes are formatting stuff done by the autoformatter specified in `devcontainer.json`.

### DIFF
--- a/public/js/common.js
+++ b/public/js/common.js
@@ -317,6 +317,14 @@ LRR.getImgSize = function (target) {
     return imgSize;
 };
 
+LRR.getImgSizeAsync = function (target) {
+    return $.ajax({
+        url: target,
+        cache: true,
+        type: "HEAD",
+    });
+};
+
 /**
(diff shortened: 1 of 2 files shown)


Difegue:
This looks about right to me. Happy to finally cross out this precaching HEAD item from my todolist! Thanks a bunch.

PR #992, opened 2024-06-25
Ignore source tag when building untagged index

anonstash:
When a metadata plugin fails, tags are not added to an archive. However, since the `source:` tag is still attached to the archive, it isn't automatically detected as an untagged archive, which makes it difficult to batch tag later on. All this PR does is to add `source` to the list of ignored tags (other changes are the auto-formatter).


Difegue:
Looks good to me. Thanks!

Issue #907, opened 2023-11-08
EH plugin gets tags from wrong url sometimes


brachna:
**LRR Version and OS**
Windows 10, WSL2, Lanraragi v0.9.0 Nightly msi

**Bug Details**
I'm having an issue with stock EH plugin. Downloading 4 archives and having auto-tagging enabled I get tags from wrong url for some of them.
Downloaded archives themselves are correct even if their tags are not, so it's tagging plugin issue.

*Update*
I just had it happen with only one link, so it's not only due to downloading multiple archives.

**Logs**
```
[Minion] [info] URL downloaded to /tmp/dHnupMX57M/(C97) [Ruiketsuan (Namidame)] Emiya-ke Futei Koukou Ryouiki Yon ~Matou Sakura no Baai~ (Fate stay night).zip
[Archive] [info] Using current date
[Auto-Plugin] [info] Executing enabled metadata plugins on archive with id 4404ffaef8f1c616be036fc2af08c886ece21e57.
[Plugin System] [info] Thumbnail hash invalid, regenerating.
[E-Hentai Login] [info] Cookies provided (mycookies)!
[E-Hentai] [info] Sending the following tags to LRR: language:russian, language:translated, parody:fate stay night, character:sakura matou, character:shirou emiya, group:ruiketsuan, artist:namidame, male:bald, male:bbm, male:big penis, female:ahegao, female:anal, female:big breasts, female:corruption, female:focus anal, female:netorare, female:paizuri, female:sole female, other:multi-work series, category:doujinshi, uploader:Lecks, timestamp:1605050459
[Plugin System] [info] Applying tag rules...
```

**Screenshots**
One of the archives have a [Russian] translation. Previously I had [French] ones as well.
![Screenshot 2023-11-08 at 04-09-23 LANraragi - Upload Center](https://github.com/Difegue/LANraragi/assets/5756107/fb43236e-8050-4e67-b194-02289aac5215)

1st source is correct, 2nd (added by tagger) is wrong.
![Screenshot 2023-11-08 at 04-15-01 (C97) Ruiketsuan (Namidame) Emiya-ke Futei Koukou Ryouiki Yon ~Matou Sakura no Baai~ (Fate_stay night) Russian by namidame](https://github.com/Difegue/LANraragi/assets/5756107/49a381f6-2844-49cf-9a3e-a411651dc9f4)

Plugin settings
[...]

PR #993, opened 2024-06-25
Enable EH script to handle source tag with https prefix

anonstash:
When pasting links directly into the downloader, the links are often prefixed with `https://` or potentially `http://`. The `source:` tag is attached to the archive with the URL containing the https prefix, but can't be picked up by the current EH metadata script because the regex does not handle it. Here I add the non-capture group `(?:https?:\/\/)?` to the regex to allow these links to be used by the script.


Difegue:
This fix shouldn't be necessary with the changes that were introduced for #907/#920. I guess we could still include this to help users with fucked tags but it'd only fix EH..

PR #994, opened 2024-06-25
Added docker-compose.yml way of installing

RexYuan:
I added a docker compose file example to the documentation. Note that the thumbnail directory bind mount is necessary because if it wasn't there already and the content folder isn't writable, it will fail to mkdir the thumb /content/thumb.

### DIFF
--- a/tools/Documentation/installing-lanraragi/docker.md
+++ b/tools/Documentation/installing-lanraragi/docker.md
@@ -30,6 +30,24 @@ difegue/lanraragi
 You can tell Docker to auto-restart the LRR container on boot by adding the `--restart always` flag to this command.
 {% endhint %}
 
+{% hint style="info" %}
+Alternatively, if you prefer a `docker-compose.yml` file, use:
+```yaml
+---
+services:
+  lanraragi:
+    image: difegue/lanraragi
+    container_name: lanraragi
+    ports:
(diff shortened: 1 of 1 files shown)


Difegue:
👋 Were you aware of the compose file already located at https://github.com/Difegue/LANraragi/blob/dev/tools/build/docker/docker-compose.yml ? I wonder if this doc change shouldn't mention this file instead.

RexYuan:
Thanks for updating the [Dockerfile](https://github.com/Difegue/LANraragi/commit/cb0ce347b45ed1620f51cdb26dca647a3448271b#diff-51eb5c812fe4d8224d70d7ee15250f4d5bdf86eacf081a0f7a7cca29825daee2)! As for the dev environment docker-compose.yml, I think it's okay to not mention it here because people won't need it for a simple user setup. Additionally, the first time I found this project I did a search of "docker-compose" in the doc site and the first thing it showed was that file and it was sorta confusing since I wasn't expecting to be reading about the dev environment the first time setting the project up

siliconfeces:
Yeah that compose file is mostly useful for developers, or if you are peculiar about your container setup and want to keep redis separate.

Difegue:
Yeah, I suppose it makes sense to include a simple one in the doc. Works for me!

PR #1012, opened 2024-07-03
Updated the FAKKU plugin to add summary support and source tag lookup.

Nixis198:
With the latest version adding summaries, the FAKKU plugin now grabs the summary as well. It now also adds the option to include the given URL as a source tag. It does need to be enabled in the plugin settings.

Also if a source tag is already in to tags, the plugin will detect it and use that if no URL is provided by the user.
This is useful for updating galleries to now add the summary.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Fakku.pm
+++ b/lib/LANraragi/Plugin/Metadata/Fakku.pm
@@ -235,7 +235,7 @@ sub get_tags_from_fakku {
 
     # If the Summary DIV doesn't exist, return a blank string.
     if ( defined $summ_div ) {
-        $summary = $summ_div->text;
+        $summary = $summ_div->all_text;
     } else {
         $summary = "";
     }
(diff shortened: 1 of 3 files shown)


lib/LANraragi/Plugin/Metadata/Fakku.pm
```diff
@@ -77,14 +78,29 @@ sub get_tags {
...
         $fakku_URL = $lrr_info->{oneshot_param};
     } else {
+        if ( $fakku_URL eq "" ) {
 
-        # Search for a FAKKU URL if the user didn't specify one
-        $fakku_URL = search_for_fakku_url( $lrr_info->{archive_title}, $ua );
+            # Search for a FAKKU URL if the user didn't specify one or none found in the tags.
+            $fakku_URL = search_for_fakku_url( $lrr_info->{archive_title}, $ua );
+        }
     }
```
Difegue, line 104: You can remove a level of nesting here:

```
if ( $lrr_info->{oneshot_param} ) {
        $fakku_URL = $lrr_info->{oneshot_param};
}

if ($fakku_URL eq "") {
     # Search for a FAKKU URL if the user didn't specify one or none found in the tags.
     $fakku_URL = search_for_fakku_url( $lrr_info->{archive_title}, $ua );
}

```

Difegue (changes requested):

PR #1013, opened 2024-07-03
Ignore tags start with "source:" or "date_added:" when build word cloud

nuclear06:


Difegue:
What issue are you having that'd be fixed by this?  
The `$minscore` filtering should already keep out source:/date: tags, unless you have multiples of the same one.

nuclear06:
For date_added: I initially didn't enable the automatic timestamp feature, resulting in a batch of archives without the date_added tag. I then use batch operation to add same timestamp for them, causing multiple archives to have the same timestamp.

For source: The archives come from multiple sources, which means different archive files might point to the same source. such as different resolutions.

Difegue:
Fair enough! Thanks for clarifying 👍

PR #1014, opened 2024-07-04
New comicinfo plugin

BEIVSIQIN:
I modified a new local comicinfo plugin based on the original comicinfo plugin to read the comicinfo information in the cbz package downloaded by FooIbar/Ehviewer. Compared to the original version, I added the reading of titles, the acquisition of some tags, and adapted to the tag conversion of E-Hentai Tag Converter v.1.0 by GrayZhao&Guerra24

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/ComicInfoToEhviewer.pm
+++ b/lib/LANraragi/Plugin/Metadata/ComicInfoToEhviewer.pm
@@ -1,0 +1,174 @@
+package LANraragi::Plugin::Metadata::ComicInfoToEhviewer;
+
+use strict;
+use warnings;
+
+use Mojo::DOM;
+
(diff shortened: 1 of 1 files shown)


Difegue:
Thank you. Is there a reason those changes couldn't simply be added to the initial comicinfo plugin?  
I'm not seeing any breaking differences here, all of these seem to just be really nice additions to go on top!

BEIVSIQIN:
sorry. Currently, it seems that there is none. I'm not sure if that plugin is still being maintained because it hasn't been updated for a long time. So I created a new one

Difegue:
No problem! In that case though, I think it would be better to just add those changes to the original bundled plugin instead of making a new one. Makes things easier related to tests as well 👍

PR #1015, opened 2024-07-04
Introduce a new tag rule

nuclear06:
The functionality of ns_a:tag_a => ns_b:tag_b is identical to ns_a:tag_a -> ns_b:tag_b, but internally utilizes a hash table to enhance speed. I attempted to implement a hash table within the existing -> structure. However, I found it challenging to accelerate the process while ensuring that the rules execute sequentially. Therefore, I added a new tag rule type called hash_replace, which performs replacements separately after all other rules have been executed.

Usage: This new rule type improves execution speed when there are a large number of replacements in the tag rules.

I use this project: [EhTagTranslation/Database](https://github.com/EhTagTranslation/Database) to generate tag rules for translation. With over 30,000 tag rules, performing batch operations on 100 archives took approximately 2 minutes. I deemed this speed to be insufficient and hence introduced this new tag rule.


Difegue (changes requested):
Nothing wrong with this at a glance, but I'd like to see the [tests for tag rules](https://github.com/Difegue/LANraragi/blob/dev/tests/LANraragi/Utils/Tags.t) and the [documentation](https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/advanced-usage/tag-rules.md?plain=1) [updated](https://github.com/Difegue/LANraragi/blob/3800e1f2da2cce7b8e8bfe9106dba5012df281e8/templates/templates_config/config_tags.html.tt2#L109) accordingly.

templates/templates_config/config_tags.html.tt2
```diff
@@ -108,6 +108,7 @@
         <br><b>-namespace:*</b> : removes all tags within this namespace
         <br><b>~namespace</b> : strips the namespace from the tags
         <br><b>tag -> new-tag</b> : replaces one tag
+        <br><b>tag => new-tag</b> : replaces one tag, but use a hash table internally for faster performance
```
Difegue, line 111: Better mention here as well this is executed after all other rules

tests/LANraragi/Utils/Tags.t
```diff
@@ -118,6 +118,125 @@ note("testing tag rules conversion...");
...
+{
+    my $text_rules = 'group:alpha => lone wolf';
+
+    my @rules = LANraragi::Utils::Tags::tags_rules_to_array($text_rules);
+
+    cmp_deeply( \@rules, [ [ 'hash_replace', 'group:alpha', 'lone wolf' ] ], 'hash_replace: simple substitution with namespace' );
+}
+
+{
+    my $text_rules = '     ping    =>   pong
```
Difegue, line 177: duplicated tests starting from here

tools/Documentation/advanced-usage/tag-rules.md
```diff
@@ -14,6 +14,7 @@ The format is as follows:
 * `-namespace:*` : removes all tags within this namespace
 * `~namespace` : strips the namespace from the tags
 * `tag -> new-tag` : replaces one tag
+* `tag => new-tag` : replaces one tag, but uses a hash table internally for faster performance. This will be executed __once__ after all other rules.
```
Difegue, line 17:
```suggestion
* `tag => new-tag` : replaces one tag, but uses a hash table internally for faster performance. These rules will be executed __once__ after all other rules.
```

lib/LANraragi/Utils/Tags.pm
```diff
@@ -87,21 +99,42 @@ sub tags_rules_to_array {
     return @rules;
 }
 
+# build hash_replace rules and return the remaining rules
+sub build_tag_replace_hash {
```
Difegue, line 103: Shouldn't this method just be merged with `tags_rules_to_array`, which should then return both the array of standard rules and the hash?  

I don't really see the point in temporarily putting `hash_replace` items in the array just to remove them in the very next function.

Difegue (changes requested):
A few more nitpicks from me but otherwise this looks good.  

I wondered if we should really have two kind of replace rules since hashes seem to be a straight upgrade, but I think there can be some value in having a final replace operation that runs after all the other rules (including other replaces) to merge tags or similar.

lib/LANraragi/Utils/Tags.pm
nuclear06: I am doing this just in case someone, somewhere, sometime, wants to get all tags array😀
you're right, I'll just merge it.

nuclear06:
> I wondered if we should really have two kind of replace rules since hashes seem to be a straight upgrade

`replace` can perform multiple replacements
```
a->b
# other rules...
b->c
```
To perform multiple replacements, all tag rules must be traversed in order, which becomes a performance bottleneck when there are a large number of rules. That's why I introduced `hash_replace`.

ps: I couldn't think of any scenario where multiple `replace` can't be replaced by a single `replace`, but who knows😀?

lib/LANraragi/Utils/Tags.pm
nuclear06: oh! I remember why I didn't merge it at that time. The original rule array is needed for persistence [here](https://github.com/Difegue/LANraragi/blob/96a9c33b6625584ccaace95e4131127407fe244f/lib/LANraragi/Controller/Config.pm#L153)

Difegue: Right, I forgot about that 😩 
Well it can't be helped in that case I guess; I don't think it's worth spinning up a secondary database field just to save the hashreplace rules.

Difegue (approved):

Difegue:
Thanks again!

Issue #966, opened 2024-04-16
Introduce a plugin that can copy metadata from another archive in the db.


soarpasser:
As the title suggests, it would be neat to have a plugin that allows copying the metadata (tags, categories etc.) from an existing archive in the database to another, using the archive ID.
This can come in handy when trying to sort archives that have unique tags added by the user themselves, instead of fetched from a website. This can also work as a workaround for when archives lose their tag because of bugs like #942.
Example:
![image](https://github.com/Difegue/LANraragi/assets/74951691/76762256-90fe-4e15-a639-b4a56aa66c3b)
![image](https://github.com/Difegue/LANraragi/assets/74951691/8dff5cb0-c766-451d-b152-c390b83b8d4c)

PR #1021, opened 2024-07-11
adds plugin CopyArchiveTags

IceBreeze:
I took advantage of the issue #966 to try some suggestions like:
- exception handling for the plugins
- named parameters

They are implemented inside the plugin because I think they are too invasive to be put elsewhere at the moment.

Also I added a little helper method to retrieve the tags from an archive and I slightly reworked various `get_archive*` methods internally, but I couldn't come up with a better name for `h_get_archive` for the one that needs the Redis instance to be passed.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
+++ b/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
@@ -25,7 +25,7 @@ sub plugin_info {
                 desc => "Enable to also copy the date (but it's up to you to remove the old one)"
             }
         ],
-        oneshot_arg => "LRR Gallery URI or ID:"
+        oneshot_arg => "LRR Gallery URL or ID:"
     );
 
 }
--- a/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
+++ b/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
@@ -4,16 +4,17 @@ use strict;
 use warnings;
 
 use LANraragi::Model::Plugins;
-use LANraragi::Utils::Database qw(get_archive_tags);
-use LANraragi::Utils::Logging  qw(get_plugin_logger);
-use LANraragi::Utils::Tags     qw(join_tags_to_string);
+use LANraragi::Utils::Database;
+use LANraragi::Utils::Logging qw(get_plugin_logger);
+use LANraragi::Utils::Tags    qw(join_tags_to_string split_tags_to_array);
+use LANraragi::Utils::String  qw(trim);
 
 #Meta-information about your plugin.
 sub plugin_info {
 
     return (
         #Standard metadata
-        name        => "CopyArchiveTags",
+        name        => "Copy Archive Tags",
         type        => "metadata",
         namespace   => "copy-archive-tags",
         author      => "IceBreeze",
@@ -48,8 +49,10 @@ sub get_tags {
 sub internal_get_tags {
     my ( $logger, $params ) = @_;
 
-    my $lrr_gid = $params->{'oneshot'};
-    $lrr_gid =~ s/^.*id=//i;    # extract the ID from the URI if necessary
+    my $lrr_gid = extract_archive_id( $params->{'oneshot'} );
+    if ( !$lrr_gid ) {
+        die "oneshot_param doesn't contain a valid archive ID\n";
+    }
 
     if ( $lrr_gid eq $params->{'lrr_info'}{'archive_id'} ) {
         die "You are using the current archive ID\n";
@@ -57,17 +60,25 @@ sub internal_get_tags {
 
     $logger->info("Copying tags from archive \"${lrr_gid}\"");
 
-    my $tags;
-    if ( $params->{'copy_date_added'} ) {
-        $tags = get_archive_tags($lrr_gid);
-    } else {
-        my @tags = get_archive_tags($lrr_gid);
+    my $tags = LANraragi::Utils::Database::get_tags($lrr_gid);
+
+    if ( !$params->{'copy_date_added'} ) {
+        my @tags = split_tags_to_array($tags);
         $tags = join_tags_to_string( grep( !m/date_added/, @tags ) );
     }
 
     my %hashdata = ( tags => $tags );
 
     return %hashdata;
+}
+
+sub extract_archive_id {
+    my ($oneshot) = @_;
+    return if ( !$oneshot || length($oneshot) < 40 );
+    if ( ( lc $oneshot ) =~ m/([0-9a-f]{40,})/ ) {
+        return $1 if length($1) == 40;
+    }
+    return;
 }
 
 sub read_params {
--- a/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
+++ b/lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
@@ -1,5 +1,8 @@
 package LANraragi::Plugin::Metadata::CopyArchiveTags;
 
+use v5.36;
+use experimental 'try';
+
 use strict;
 use warnings;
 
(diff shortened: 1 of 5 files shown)


lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -0,0 +1,88 @@
...
+use LANraragi::Utils::Database qw(get_archive_tags);
+use LANraragi::Utils::Logging  qw(get_plugin_logger);
+use LANraragi::Utils::Tags     qw(join_tags_to_string);
+
+#Meta-information about your plugin.
+sub plugin_info {
+
+    return (
+        #Standard metadata
+        name        => "CopyArchiveTags",
```
Difegue, line 16: Plugin names can have spaces in them, unlike the namespaces

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -0,0 +1,88 @@
...
+        author      => "IceBreeze",
+        version     => "1.0",
+        description => "Copy tags from another LRR archive given either the URI or the ID.",
+        parameters  => [
+            {   type => "bool",
+                name => 'copy_date_added',
+                desc => "Enable to also copy the date (but it's up to you to remove the old one)"
+            }
+        ],
+        oneshot_arg => "LRR Gallery URI or ID:"
```
Difegue, line 28:
```suggestion
        oneshot_arg => "LRR Gallery URL or ID:"
```

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    $logger->info( "Sending the following tags to LRR: " . ( $hashdata{tags} || '-' ) );
+    return %hashdata;
+}
+
+sub internal_get_tags {
+    my ( $logger, $params ) = @_;
+
+    my $lrr_gid = $params->{'oneshot'};
```
Difegue, line 51: This doesn't seem to have any fail states in case an invalid ID is provided (or no ID at all, which could happen if you enable this plugin for automatic execution for some reason)

lib/LANraragi/Utils/Database.pm
```diff
@@ -187,6 +190,14 @@ sub get_archive_json_multi (@ids) {
     return @archives;
 }
 
+sub get_archive_tags ($id) {
+    my %archive_info = get_archive($id);
+    return undef if ( !%archive_info );
+    return wantarray
```
Difegue, line 196: This is too much lmaoperl even for me here, I don't think subs should be able to return different types depending on what you ask for - especially without any kind of comments to specify the behavior 🤔  

Let's just return the normal tag string here, people can chain it to split_tags_to_array if needed.

lib/LANraragi/Utils/Database.pm
```diff
@@ -187,6 +190,14 @@ sub get_archive_json_multi (@ids) {
     return @archives;
 }
 
+sub get_archive_tags ($id) {
```
Difegue, line 193: This should match with `set_tags`. 
```suggestion
sub get_tags ($id) {
```

tests/LANraragi/Plugin/Metadata/CopyArchiveTags.t
```diff
@@ -0,0 +1,224 @@
+use strict;
```
Difegue, line 1: Nice; but this is missing the "using namespace" of the new plugin in the global test file

Difegue (changes requested):

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -0,0 +1,88 @@
...
+    );
+
+}
+
+sub get_tags {
+    my $params = read_params(@_);
+    my $logger = get_plugin_logger();
+
+    # should be handled in the caller
+    my %hashdata = eval { internal_get_tags( $logger, $params ); };
```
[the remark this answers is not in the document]
IceBreeze: > How about using perl's try/catch feature for this? It recently got out of experimental: https://www.perl.com/article/what-is-new-in-perl/#try_catch

Honestly I've been using it for a while, just not here. I'll gladly do it.

> Eventually I wouldn't mind moving this up to the main plugin execution, it shouldn't be too complex.

This was my intention. It should simplify the error management inside the plugins.

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -0,0 +1,88 @@
...
+        my @tags = get_archive_tags($lrr_gid);
+        $tags = join_tags_to_string( grep( !m/date_added/, @tags ) );
+    }
+
+    my %hashdata = ( tags => $tags );
+
+    return %hashdata;
+}
+
+sub read_params {
```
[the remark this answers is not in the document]
IceBreeze: I agree, especially because there is only one parameters here.
This is another thing that could be convenient to have when building the parameters in the caller, but it's more invasive and I don't know if there's a reason why named parameters haven't been already introduced, so I put it inside here just as an example.

lib/LANraragi/Utils/Database.pm
```diff
@@ -135,21 +136,23 @@ sub add_pagecount ( $redis, $id ) {
...
 
+# Retrieves the archive's info as hash (empty if not found)
+sub get_archive ($id) {
+    my $redis = LANraragi::Model::Config->get_redis;
+    my %hash  = h_get_archive( $redis, $id );
+    $redis->quit();
+    return %hash;
+}
+
+sub h_get_archive ( $redis, $id ) {
```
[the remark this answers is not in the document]
IceBreeze: Here I wanted to name the call `$redis->hgetall($id);` to make it immediately understandable and searchable (during a refactoring). It can also be used transactionally, like in `get_archive_json_multi()` and it restricts the call to `hgetall` inside the Database package. I think it might help in building tests.

Anyway being the only method changed like this, is not really helpful and I wanted to read your opinion before replacing its other calls around.

lib/LANraragi/Utils/Database.pm
IceBreeze: OK

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
IceBreeze: It does nothing unless the ID matches something, but we can definitely avoid the call.

tests/LANraragi/Plugin/Metadata/CopyArchiveTags.t
IceBreeze: Are you referring to `modules.t`?

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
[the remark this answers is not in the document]
IceBreeze: Maybe it is better to edit Plugins.pm in a dedicated PR.

tests/LANraragi/Plugin/Metadata/CopyArchiveTags.t
Difegue: yep, ty

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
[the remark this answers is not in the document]
Difegue: Nothing againtst named parameters in particular, I just hadn't thought of it back then.  
I think it'd be a nice improvement, but it'd require retooling all the built-in plugins 🫠

lib/LANraragi/Utils/Database.pm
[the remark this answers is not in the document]
Difegue: I'm not really a fan, sorry!  
I don't see either how useful it'd be for testing compared to just mocking hgetall directly like we do now. 🤔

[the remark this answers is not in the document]
IceBreeze: It's fine, I shouldn't have left it in from the beginning.

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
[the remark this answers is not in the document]
IceBreeze: So I'll look into moving this outside of the plugin.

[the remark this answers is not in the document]
Difegue: Sounds good - I think I can merge this in the meantime now. Thanks!

Difegue (approved):

PR #1023, opened 2024-07-17
Add support for yet another version of info.yaml

siliconfeces:
New site has different data in the info.yaml file. This adds support for that, and fixes an issue where missing data in info.yaml could make tags ugly (empty namespaced tags, etc).


siliconfeces:
Errors should be unrelated (unmocked ehentai tests)

Difegue:
lgtm. I'll have to look into those test failures at some point.

PR #1027, opened 2024-07-23
Rewrite given/when as if-chains

### LINKED ISSUE #882: `switch` will be removed in future Perl releases
https://perldoc.perl.org/perl5380delta#Switch-and-Smart-Match-operator  

> These features will be entirely removed from perl in v5.42.0.  

There are some `use feature "switch";` and `given/when`s in the codebase I should get rid of before that.

siliconfeces:
given/when is deprecated (and causes whining in logs).


Difegue (changes requested):
You should remove the `use feature "switch" ` statements at the top of the files as well.

Difegue (approved):

PR #1041, opened 2024-07-30
changes exception handling for plugins

IceBreeze:
This is a work in progress, but I would like to know your opinion before going further.

At this point the refactoring is limited to the metadata plugins.
I handled the exception inside the sub `exec_metadata_plugin`, this way I was able to remove some "eval" without replacing them.


lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -65,16 +47,16 @@ sub internal_get_tags {
...
     if ( !$params->{'copy_date_added'} ) {
         my @tags = split_tags_to_array($tags);
         $tags = join_tags_to_string( grep( !m/date_added/, @tags ) );
     }
 
-    my %hashdata = ( tags => $tags );
+    $logger->info( "Sending the following tags to LRR: " . ( $tags || '-' ) );
 
-    return %hashdata;
+    return ( tags => $tags || '' );
```
IceBreeze, line 59: I was wondering why not returning the tags as an array? It should reduce the number of splits an joins around the project.

lib/LANraragi/Model/Plugins.pm
```diff
@@ -195,42 +193,43 @@ sub exec_metadata_plugin {
...
+    }
+    $redis->quit();
 
+    my %returnhash;
+    try {
         # Hand it off to the plugin here.
         # If the plugin requires a login, execute that first to get a UserAgent
         my %pluginfo = $plugin->plugin_info();
-        my $ua       = exec_login_plugin( $pluginfo{login_from} );
+        my $ua       = exec_login_plugin( $pluginfo{login_from} ) if ( exists $pluginfo{login_from} );
```
Difegue, line 232: I believe this would make `$ua` undef if there's no login plugin set.  
This is technically different from the old behavior where you would always have a Mojo::UserAgent object available in `%infohash`, even if you don't depend on a login plugin.  

I don't know if external plugins came to depend on this, but it might be safer to keep the old behavior here.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -163,10 +163,8 @@ sub use_plugin {
...
             eval { %plugin_result = LANraragi::Model::Plugins::exec_script_plugin( $plugin, $input, @settings ); };
-        }
-
-        if ( $pluginfo{type} eq "metadata" ) {
-            eval { %plugin_result = LANraragi::Model::Plugins::exec_metadata_plugin( $plugin, $id, $input, @settings ); };
+        } elsif ( $pluginfo{type} eq "metadata" ) {
+            %plugin_result = LANraragi::Model::Plugins::exec_metadata_plugin( $plugin, $id, $input, @settings );
         }
 
         if ($@) {
```
Difegue, line 170: You're probably already aware, but this check would be removed once script plugins have an internal try/catch as well

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
Difegue: Metadata plugin return tags are just string-concatenated to the existing tags, so this is technically faster than doing a split/join dance server-wise.  

Although this was written before the introduction of search indexes where we still need to do splits regardless.. Since this is front-facing behavior used by plugin devs I wouldn't try changing it now unfortunately 🤷  

(also since most metadata sources just spit out tags as a comma-separated string anyway I think it usually makes their lives easier)

lib/LANraragi/Plugin/Metadata/CopyArchiveTags.pm
```diff
@@ -1,10 +1,8 @@
 package LANraragi::Plugin::Metadata::CopyArchiveTags;
 
-use v5.36;
-use experimental 'try';
-
 use strict;
 use warnings;
+use Data::Dumper;
```
Difegue, line 5: leftover I guess

lib/LANraragi/Model/Plugins.pm
```diff
@@ -1,5 +1,9 @@
 package LANraragi::Model::Plugins;
 
+use v5.36;
+use experimental 'try';
+no warnings 'experimental::try';
```
Difegue, line 5: Does this work for you? My perl 5.36 still yells at me about try/catch being experimental when I added this in the copyarchivetags plugin on my own 🤔

Difegue (commented):
Thanks for the test code as usual, this is much appreciated.  

I think it's fine to keep going in this direction, it's certainly nice to cut down on having to eval check everywhere.

lib/LANraragi/Model/Plugins.pm
IceBreeze: I don't know why at that time I was really sure that it was created here only for plugins that required to login. Shame on me!

lib/LANraragi/Utils/Plugins.pm
IceBreeze: Yes, I wanted to treat it in the same way as the metadata plugins.

lib/LANraragi/Model/Plugins.pm
IceBreeze: I know, it only removed a couple of lines from the log in foreground mode, but it keeps spamming the tests.
So I'm going to put it inside every sub with a "try/catch". I don't like it very much, but it seems the way, at least until you decide to upgrade to v5.40.

Difegue: Right, I guess we'll just have to deal with it then.

I can't really require 5.40 until it lands in debian stable, that'd be annoying for everyone. We'll probably have to wait a few years for that though 😤

IceBreeze: Maybe no, unless there are specific reasons, you can try the image "perl:5.40.0-bullseye". It's an official image.

lib/LANraragi/Model/Plugins.pm
```diff
@@ -243,8 +248,11 @@ sub exec_metadata_plugin {
...
         );
 
-        my %newmetadata = $plugin->get_tags( \%infohash, @args );
+        my %newmetadata;
 
+        %newmetadata = $plugin->get_tags( \%infohash, @args );
+
+        # TODO: remove this block after changing all the metadata plugins
         #Error checking
         if ( exists $newmetadata{error} ) {
```
IceBreeze, line 257: Recently I saw a repository with some other plugins, so I'm unsure whether to remove this block of code without notice.
We should leave it for some time and add a warning in the documentation.

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -139,7 +141,7 @@ sub lookup_gallery {
         my ( $gId, $gToken ) = &ehentai_parse( $URL, $ua );
 
         if ( $gId ne "" && $gToken ne "" ) {
-            return ( $gId, $gToken );
+            die "Reverse Image Search failed\n";
```
Difegue, line 144: This is incorrect, the sub is meant to return early if it finds a gID/gToken.  
Any errors would just bubble up from `ehentai_parse` now I believe

lib/LANraragi/Plugin/Metadata/EHentai.pm
```diff
@@ -153,7 +155,7 @@ sub lookup_gallery {
         my ( $gId, $gToken ) = &ehentai_parse( $URL, $ua );
 
         if ( $gId ne "" && $gToken ne "" ) {
-            return ( $gId, $gToken );
+            die "The archive title doesn't contain a valid gID\n";
```
Difegue, line 158: Same here

lib/LANraragi/Plugin/Metadata/nHentai.pm
```diff
@@ -220,12 +216,6 @@ sub get_tags_from_NH {
     my %hashdata = ( tags => "" );
 
     my $html = get_html_from_NH( $gID, $ua );
-
-    # If the string starts with "error", we couldn't retrieve data from NH.
-    if ( $html =~ /^error/ ) {
```
Difegue, line 225: Why was this block removed?

Difegue (changes requested):
Looking great. A few more notes from me on existing plugin updates but I think we're at the finish line here 👍

lib/LANraragi/Model/Plugins.pm
Difegue: Right might as well keep it in - It's not a lot of code anyway so it's better to avoid the breaking change.

lib/LANraragi/Plugin/Metadata/EHentai.pm
IceBreeze: Yes, I misread the negation... :man_facepalming:

lib/LANraragi/Plugin/Metadata/EHentai.pm
IceBreeze: As above.

lib/LANraragi/Plugin/Metadata/nHentai.pm
IceBreeze: Now it fails inside `get_html_from_NH(...)` and dies.

Difegue: right I hadn't read this fully - thanks 👍

Difegue (approved):
All good for me! Thank you very much for this.

PR #1045, opened 2024-08-02
Support for additional fields in Tanks and filter in retrieve

EfronC:
[POSSIBLY INCOMPLETE]

This is a small PR just to show the idea of the initial added support for extra fields in Tanks. The commit adds a new file called Constants.pm in utils, since I wanted to have a variable with the extra fields separated of the Tank code to keep the structure defined and easy to modify, and because I saw Generic.pm was full with functions, I decided to create a file exclusively for variables, otherwise both work the same.

This PR also adds support in the "Get All Tankoubons" EP to send a param name, so you can filter the result, still I'm not sure if ti should be done this way, since if someone ever gets to have so many Tanks(Which I doubt, but is possible), this might take some time to do the search, since is visiting the whole list twice at worst before paginating.

This probably has a few issues, so I keep pending for comments about changes that can be done to improve this idea.


EfronC:
There is a bug that I just noted, where the get tankoubon EP appends an extra field "5":0 in each tank, but I've no idea why this is happening, I've tracked that it might be related to how the Tanks are created, since reducing the number of flag fields reduce the number in it, but still I've no idea where is that field being inserted.

lib/LANraragi/Controller/Api/Tankoubon.pm
```diff
@@ -92,6 +93,24 @@ sub update_archive_list {
     }
 }
 
+sub update_metadata {
```
Difegue, line 96: I'm unconvinced we need a new API endpoint here -- Couldn't the metadata json be another, optional parameter in the original update tankoubon endpoint?  

(I'm not a fan of the routing for the original endpoint through.. see comment below)

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -27,6 +28,21 @@ sub get_tankoubon_list ( $page = 0 ) {
     # Tankoubons are represented by TANK_[timestamp] in DB. Can't wait for 2038!
     my @tanks = $redis->keys('TANK_??????????');
 
+    # This can potentially get very slow searches!!!
+    if ( length($name) ) {
```
Difegue, line 32: I wouldn't bother implementing this -- all tank names are now included in LRR_TITLES when building the stat indexes, so even in the event that we need to add a search bar to a "here are all your tanks" page, I would rather rely on the existing search infrastructure.  

(probably by just swapping the initial id list to be all the tank IDs instead of the archive/tankID mix we have for base search rn)  

I'd even argue for just doing this clientside, but it might become problematic later if users start having 1000s of tanks so ¯\\\_(ツ)_/¯

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -84,13 +101,59 @@ sub create_tankoubon ( $name, $tank_id ) {
 
     # Default values for new group
     # Score 0 will be reserved for the name of the tank
-    $redis->zadd( $tank_id, 0, redis_encode($name) );
+    $redis->zadd( $tank_id, $TANK_METADATA{"name"}, redis_encode("name_${name}") );
+    $redis->zadd( $tank_id, $TANK_METADATA{"summary"}, "summary_" );
+    $redis->zadd( $tank_id, $TANK_METADATA{"thumbhash"}, "thumbhash_" );
```
Difegue, line 106: Do those fields need default values? I'd just leave em as empty strings

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -84,13 +101,59 @@ sub create_tankoubon ( $name, $tank_id ) {
...
+    my $redis           = LANraragi::Model::Config->get_redis;
+    my $err             = "";
+    my $name            = $data->{"name"};
+    my $summary         = $data->{"summary"};
+    my $thumbhash       = $data->{"thumbhash"};
+    my $tags            = $data->{"tags"};
+    my $alias           = $data->{"alias"};
+
+    if ( $redis->exists($tank_id) ) {
+        $redis->zremrangebyscore( $tank_id, $TANK_METADATA{"name"}, $TANK_METADATA{"name"} );
```
Difegue, line 130: Those zremrangebyscore/zadd combos could probably be refactored into a reusable subroutine  

(And again, do we need the prefixes here for each value? The original name property didn't need one)

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -118,8 +181,29 @@ sub get_tankoubon ( $tank_id, $fulldata = 0, $page = 0 ) {
     my @limit = split( ' ', "LIMIT " . ( $keysperpage * $page ) . " $keysperpage" );
 
     # Get name
-    my @name = $redis->zrangebyscore( $tank_id, 0, 0, qw{LIMIT 0 1} );
-    $tank{name} = redis_decode( $name[0] );
+    my @name = $redis->zrangebyscore( $tank_id, $TANK_METADATA{"name"}, $TANK_METADATA{"name"}, qw{LIMIT 0 1} );
```
Difegue, line 184: In a similar way, this can probably be refactored into an individual sub.  

(And now that I'm thinking of it, couldn't you do a single zrangebyscore from 0 to -5 to get all the metadata items at once?)

lib/LANraragi/Utils/Constants.pm
```diff
@@ -0,0 +1,20 @@
...
+use strict;
+use warnings;
+use utf8;
+no warnings 'experimental';
+
+use Exporter 'import';
+our @EXPORT_OK = qw(%TANK_METADATA);
+
+
+our %TANK_METADATA = (
```
Difegue, line 12: `TANK_METADATA` is only used in Model/Tankoubon.pm -- I'm not convinced it needs a spare class atm tbh

lib/LANraragi/Utils/Constants.pm
```diff
@@ -0,0 +1,20 @@
...
+        "name", 0,
+        "summary", -1,
+        "thumbhash", -2,
```
Difegue, line 15: Using negative indices for the metadata is clever btw, I hadn't thought of that!  

My opinion might be changed on this, but I don't think thumbhash and alias are needed here:  

- Since you can't execute plugins on tanks there's no need to store a thumb hash.  
  - But we might need another thumbnail-related field here in the future; I'm not 100% on what the story is for tank thumbnails yet: Should they just use the thumbnail of the first archive? A composite of all archive thumbs done thru magick? An arbitrary image path? I'll probably keep it simple for now and just use the first archive thumb  
- What is alias even meant to be?  

Tags should always just be the aggregate of all archive tags, but one could argue a field here would be useful for caching.. I was thinking about that while doing the search implementation so might as well keep it for now.

lib/LANraragi/Utils/Routing.pm
```diff
@@ -158,6 +158,7 @@ sub apply_routes {
     $public_api->get('/api/tankoubons')->to('api-tankoubon#get_tankoubon_list');
     $public_api->get('/api/tankoubons/:id')->to('api-tankoubon#get_tankoubon');
     $logged_in_api->put('/api/tankoubons')->to('api-tankoubon#create_tankoubon');
+    $logged_in_api->put('/api/tankoubons/:id')->to('api-tankoubon#update_metadata');
```
Difegue, line 161: I think this API should be the definite endpoint for updating tankoubon data instead of PUT` /api/tanks/:id/archive` .  

It follows the convention of the /api/archive endpoints -- Not sure why I accepted the current one to begin with but since the feature isn't public yet, might as well fix it here.

Difegue (changes requested):
Needs additional tests imo, but this makes sense to me.  
A few general code quality comments and Qs. 👍

lib/LANraragi/Model/Tankoubon.pm
EfronC: Then we can remove it, TBH it never ended up growing in me, and clientside searches are not out of place either, I think Ichaival does something similar with file searches.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Here is the problem: Sorted sets require a unique key for the name, otherwise it overwrites the existent score, so, for example, if you have a Tank called "Name A"(Score 0), and then you add a summary field with the string "Name A"(Score -1), it will overwrite the first, so now your tank will not have score 0, hence unnamed, and this can happen to every field that the user can manually write, that's why I added a prefix to all metadata fields, so they are unique, even if they are blank, and I need them this way since I added a function to remove the prefix.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Will check if I can refactor that, it's mostly a copy paste, so I don't think will be hard to do. For the prefixes check my previous comment, is an DB integrity protection.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Supposing we do not increase the amount of metadata in the future, yes, we could get all in one harcoded call. Or maybe even better I can use the dict of keys to get the lower score... Will try a few things.

lib/LANraragi/Utils/Constants.pm
EfronC: Yeah, it's not for the present, but for the future, if any other model wants to eventually get the tanks to retrieve a piece of metadata, it's better to be able to do it by the name of the field, instead of guessing the number. Still I'm open to remove it and harcode the scores if you are still not convinced.

lib/LANraragi/Controller/Api/Tankoubon.pm
EfronC: We can, I just wanted to separate the metadata update from the content, but I can join them without problem.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Tried to do this, but the fact that a sorted set can have multiple values in the same score is opening the door to many bugs if someone managed to mess with the update_tankoubon EP. Basically it grabs all the elements within those scores, and then matches the names using the dict of keys, but if one scores has multiple elements, it causes a mismatch. It could be added that the zrangebyscore returns the elements with their scores to try to manage that, but sounds like a lot of things when the number of lines we are replacing will not decrease.

For the moment I added the method that resolves this inside tankoubon_tools method, but I think is a good idea for you to give it a look before implementing it.

lib/LANraragi/Utils/Constants.pm
Difegue: Oh I meant moving the hash into Model/Tank, I do think being able to put a name onto them is good

lib/LANraragi/Model/Tankoubon.pm
Difegue: I see, that makes sense. Let's keep the prefixes then I suppose 🤔

(I'm starting to wonder if we shouldn't just have used hashes here as well instead of trying to wrangle sorted sets)

lib/LANraragi/Utils/Constants.pm
Difegue: I mean, the frontend doesn't use thumbhash in any form or fashion, that's just used by plugins for reverse image searching.  
Thumbnail lookup only needs an archive ID, which is why I was planning to only use the first archive ID for now.  

As for alias, I think we could add it later if needed, yeah.

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -283,6 +223,50 @@ sub delete_tankoubon ($tank_id) {
...
+
+# update_metadata(tankoubonid, data)
+#   Updates the metadata in the Tankoubon.
+#   Returns 1 on success, 0 on failure alongside an error message.
+sub update_metadata( $tank_id, $data ) {
+
+    my $logger          = get_logger( "Tankoubon", "lanraragi" );
+    my $redis           = LANraragi::Model::Config->get_redis;
+    my $err             = "";
+    my $name            = $data->{"metadata"}->{"name"};
```
Difegue, line 247: I didn't catch it the first time but requiring all the fields to be set is pretty annoying for API clients and doesn't match how the update archive endpoint functions -- Would it be possible to change that?  

Feels like we could just undef check before calling update_meta for each field

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -509,4 +493,41 @@ sub get_tankoubons_containing_archive ($arcid) {
     return @tankoubons;
 }
 
+sub tankoubon_tools ( $action, $tank_id, $field, $value ) {
```
Difegue, line 496: Wouldn't this be better as just two distinct subs?

Difegue (changes requested):
I'll have to give `fetch_data` a more indepth think, but did find some other stuff in the meantime.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Yeah, I did it this way because the rule is that for PUT operations, you should set all the fields, while PATCH ones are the ones you send only one field, but I guess it's not necessary to follow that here. I'll do the change.

lib/LANraragi/Model/Tankoubon.pm
EfronC: Ok, I joined as a way to have a single method to do all operations related to metadata, but it can be done separated.

lib/LANraragi/Utils/Constants.pm
EfronC: Yeah, I was talking about that too, since having a dedicated file for all shared constants is usually good practice, but we are not using this a lot in this project, so I guess putting it into Model/Tank is ok.

lib/LANraragi/Model/Tankoubon.pm
EfronC: So, I asked our friend ChatGPT for a way to fix the problem with this new method, the solution doubles the complexity, but at least ensure we get the pairings correctly(There might still be bugs, but that will get there little by little). Also added a new method to clean the result of get_tankoubon, since the result was generating some thrash fields.

tools/Documentation/api-documentation/tankoubon-api.md
```diff
@@ -166,26 +174,29 @@ Name of the Category.
 {% endswagger-response %}
 {% endswagger %}
 
-{% swagger baseUrl="http://lrr.tvc-16.science" path="/api/tankoubons/:id/archive" method="put" summary="🔑Update a Tankoubon" %}
+{% swagger baseUrl="http://lrr.tvc-16.science" path="/api/tankoubons/:id" method="put" summary="Update a Tankoubon" %}
```
Difegue, line 177: This API still requires a login so the key emoji indicator should remain

tools/Documentation/api-documentation/tankoubon-api.md
```diff
@@ -166,26 +174,29 @@ Name of the Category.
...
 {% swagger-parameter name="id" type="string" required="true" in="path" %}
 ID of the Tankoubon to update.
 {% endswagger-parameter %}
 
-{% swagger-parameter name="archives" type="array" required="true" in="body" %}
-Ordered array with the IDs of the archives.
+{% swagger-parameter name="archives" type="json" required="true" in="body" %}
+Json with 2 keys "archives" and "metadata" defining:
+
+- archives: Ordered array with the IDs of the archives.
```
Difegue, line 189: Should we mention that this key is technically optional?  
It is possible to supply the API with just a metadata JSON if they only need to do a title/summary swap.  

(the reverse isn't really possible, but they can just add an empty `metadata` JSON which might be helpful to mention as well)

tools/Documentation/api-documentation/tankoubon-api.md
```diff
@@ -166,26 +174,29 @@ Name of the Category.
...
+- metadata: Json with the metadata parameters: name, summary, tags.
```
Difegue, line 190: In a similar fashion, the parameters within the metadata json are optional

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -254,9 +322,12 @@ sub update_archive_list ( $tank_id, $data ) {
         }
 
         # Update
-        $redis->zadd( $tank_id, @update );
+        if ( $len > 0 ) { # Move this line and the previous for inside the previous else?
```
Difegue, line 325: Sounds good to me yeah

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -419,4 +490,40 @@ sub get_tankoubons_containing_archive ($arcid) {
     return @tankoubons;
 }
 
+sub update_metadata_field ( $tank_id, $field, $value ) {
+    my $redis        = LANraragi::Model::Config->get_redis;
+    my $logger       = get_logger( "Tankoubon", "lanraragi" );
+    my $err          = "";
```
Difegue, line 496: This variable does nothing here now, right?  
There's no error checking at all here so you don't even need a return statement at this point it seems..

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -419,4 +490,40 @@ sub get_tankoubons_containing_archive ($arcid) {
...
+
+    $redis->zremrangebyscore( $tank_id, $TANK_METADATA{$field}, $TANK_METADATA{$field} );
+    $redis->zadd( $tank_id, $TANK_METADATA{$field}, redis_encode("${field}_${value}") );
+
+    return ( 1, $err );
+}
+
+sub fetch_metadata_fields ( $tank_id ) {
+    my $redis        = LANraragi::Model::Config->get_redis;
+    my $err          = "";
```
Difegue, line 506: Ditto here

Difegue (changes requested):
This looks markedly better to me. 👍 Just a bit more cleanup I've noticed.

tools/Documentation/api-documentation/tankoubon-api.md
EfronC: Both actions have a piece of code like this, which ensures that if you don't send the key, then it doesn't do anything, so both are mutually optional.

```
if ( not defined $data->{"archives"} ) {
    return ( 1, "" );
}
```

Difegue (approved):
Perfect, thanks!

PR #1046, opened 2024-08-06
fix: check exist upload_date tag before processing

lnhpm:
Just checking `upload_date` tag before processing because some archives are missing this field


Difegue:
Works for me, thanks!

PR #1050, opened 2024-08-11
Avoid ARRAY garbage tags for unexpected koromo info.json files

siliconfeces:
Encountered koromo'ish info.json files with multiple magazines, which caused good ol' `magazine:ARRAY(0xbadb00b13)` tags. This PR fixes that, and ensures the same shit doesn't happen to other tags.


Difegue:
backup tests are brokey but that's my fault again; Changes here look good ty 👍

Issue #363, opened 2020-11-15
[Feature Request] Add base url support


fullmetal1:
The purpose is to provide the ability to run lanraragi behind a reverse proxy as a subpath of a domain instead of a subdomain, since many free services do not allow ddns domains to have subdomains.

Difegue:
According to the links in the [previous issue](https://github.com/Difegue/LANraragi/issues/32) we touched upon this subject, implementation could be as simple as adding a hook if a base url is configured:  

```
app->hook( before_dispatch => sub {
        my $c = shift;
        unless ($c->req->url->path->to_route =~ m#/baseurl#) {
            $c->req->url->path( "/baseurl" . $c->req->url->path->to_string);
        }
    } );
```
(https://stackoverflow.com/questions/25170147/base-url-in-mojolicious-template-rendering)  

The bigger question is where to put the base url config?  
I honestly don't think this is a setting that would be used much, and it can't be set from the web interface since you likely won't have access to it, so it'd probably be fine in `lrr.conf`. Might also want to add an environment variable override for Docker users?

fullmetal1:
You already have a hook into before dispatch, and that module has access to the LANraragi settings module. I was testing it out but couldn't understand mojolicious' url object. Had some wierd behaviour with url->base that would make it use the wrong address, and also the web app pages didn't seem to exist at the rewritten locations.I couldn't figure it out after a few hours.

Difegue:
Oh yeah, the SIGTERM catcher does hook in before_dispatch.  
You can add code in there no problem but I still think it'd be better/easier to put the base url setting in lrr.conf, you could then access it with something as dumb as `$config->{base_url}`.  

If you can't get stuff working in a hook, I think `Routing.pm` would also be an acceptable place to try? You could store the base url setting in a string (defaulting to empty of course) and prepend that to each routed URL path.

alethiophile:
I'm currently looking to deploy LANraragi behind a reverse proxy under a base URL path, so I thought I would look at this issue and see what can be done about it.

The underlying requirement is to prepend the configured base URL path to every link LANraragi generates, including all the static files. You can do this for links generated via `url_for` with the hook code mentioned above. However, one big problem is that the templates and the frontend JS are all littered with hard-coded URL strings, meaning every one of those would have to be altered accordingly -- either to use `url_for`, to depend on a central base URL variable, or (depending) to use a path-relative URL.

[...]

PR #1052, opened 2024-08-12
Add base URL feature

alethiophile:
This is toward adding a base URL feature, as discussed in https://github.com/Difegue/LANraragi/issues/363.

This works as follows: you can configure a reverse proxy server to forward traffic to LANraragi only under a given path. The proxy must be configured to strip the path prefix from the forwarded requests (e.g. in Caddy this is the `handle_path` directive, as opposed to `handle`). Once this is configured, LANraragi can be viewed as usual using subdirectories of that path. E.g. if a top-level deployment had the URL `manga.example.com`, a subdirectory deployment might have the URL `example.com/manga`. LANraragi can be used as normal, with all its internal URLs rewritten to be under the `/manga` subdirectory.

Internally, this works as follows: the reverse proxy strips the path prefix from requests as they pass through. Thus, LANraragi still sees ordinary requests to `/api`, `/js`, etc. However, all internal links generated by LANraragi are generated relative to the new path prefix. The user will click a link to `/manga/reader`, which will be processed into `/reader` by the reverse proxy. The main server-side system for this is the Mojo `url_for` helper (https://docs.mojolicious.org/Mojolicious/Controller#url_for). `url_for` constructs URLs relative to the incoming request URL's `base` member. Thus, when a base URL is configured, the `before_dispatch` hook alters the URL base to include that path prefix. This makes all URLs generated via `url_for` include it.

For the Javascript frontend, the base URL will be passed to the frontend via code in the template. Then, if configured, all API calls will be rewritten to include the base URL. This is not yet finished as I submit this draft.

### DIFF
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -196,6 +196,9 @@ sub startup {
                     $c->req->url->base->path($prefix);
                 }
             }
+            # SameSite=Lax is the default behavior here; I set it
+            # explicitly to get rid of a warning in the browser
+            $c->cookie("lrr_baseurl" => $prefix, { samesite => "lax" });
         }
     );
 
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -191,6 +191,8 @@ sub startup {
             if ($prefix) {
                 if (!$prefix =~ m|^/[^"]*[^/"]$|) {
                     say "Warning: configured URL prefix '$prefix' invalid, ignoring";
+                    # if prefix is invalid, then set it to empty for the cookie
+                    $prefix = "";
                 }
                 else {
                     $c->req->url->base->path($prefix);
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -187,7 +187,7 @@ sub startup {
             my $c = shift;
             state $unused = add_sigint_handler();
 
-            my $prefix = $self->LRR_CONF->get_baseurl;
+            my $prefix = $self->LRR_BASEURL;
             if ($prefix) {
                 if (!$prefix =~ m|^/[^"]*[^/"]$|) {
                     say "Warning: configured URL prefix '$prefix' invalid, ignoring";
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -179,7 +179,18 @@ sub startup {
     # (https://stackoverflow.com/questions/60814220/how-to-manage-myself-sigint-and-sigterm-signals)
     $self->hook(
         before_dispatch => sub {
+            my $c = shift;
             state $unused = add_sigint_handler();
+
+            my $prefix = $self->LRR_CONF->get_baseurl;
+            if ($prefix) {
+                if (!$prefix =~ m|^/.*[^/]$|) {
+                    say "Warning: configured URL prefix '$prefix' invalid, ignoring";
+                }
+                else {
+                    $c->req->url->base->path($prefix);
+                }
+            }
         }
     );
 
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -72,6 +72,11 @@ sub startup {
         }
     );
 
+    # for some reason I can't call the one under LRR_CONF from the
+    # templates, so create a separate helper here
+    my $prefix = $self->LRR_CONF->get_baseurl();
+    $self->helper( LRR_BASEURL => sub { return $prefix } );
+
     #Check if a Redis server is running on the provided address/port
     eval { $self->LRR_CONF->get_redis->ping(); };
     if ($@) {
@@ -184,7 +189,7 @@ sub startup {
 
             my $prefix = $self->LRR_CONF->get_baseurl;
             if ($prefix) {
-                if (!$prefix =~ m|^/.*[^/]$|) {
+                if (!$prefix =~ m|^/[^"]*[^/"]$|) {
                     say "Warning: configured URL prefix '$prefix' invalid, ignoring";
                 }
                 else {
--- a/public/js/common.js
+++ b/public/js/common.js
@@ -251,7 +251,7 @@ LRR.buildThumbnailDiv = function (data, tagTooltip = true) {
     const thumbCss = (localStorage.cropthumbs === "true") ? "id3" : "id3 nocrop";
     // The ID can be in a different field depending on the archive object...
     const id = data.arcid || data.id;
-    let reader_url = new LRR.apiURL(`reader?id=${id}`);
+    let reader_url = new LRR.apiURL(`/reader?id=${id}`);
 
     return `<div class="id1 context-menu swiper-slide" id="${id}">
                 <div class="id2">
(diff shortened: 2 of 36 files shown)


alethiophile:
The updates to the Javascript frontend and the test suite are now complete. My local testing shows this is working.

To use the feature, configure your reverse proxy appropriately, then set the variable `base_url_path` in `lrr.conf`. The `base_url_path` value is an empty string by default; this maintains the old behavior. If it is set, it must be a valid URL path fragment, beginning with a slash but with no trailing slash.

It also remains to update the documentation, since this is a user-visible feature.

lib/LANraragi.pm
```diff
@@ -179,7 +184,18 @@ sub startup {
     # (https://stackoverflow.com/questions/60814220/how-to-manage-myself-sigint-and-sigterm-signals)
     $self->hook(
         before_dispatch => sub {
+            my $c = shift;
             state $unused = add_sigint_handler();
+
+            my $prefix = $self->LRR_CONF->get_baseurl;
```
Difegue, line 190: Small optimization here but ig you could use LRR_BASEURL directly?

public/js/common.js
```diff
@@ -3,6 +3,41 @@
  */
 const LRR = {};
 
+// This class is used to wrap URLs that point into the app, including
+// API endpoints and browser targets. It reads the configured base URL
+// from the template, then prepends it to the provided URL in the
+// toString() method.
+
+// Every operation involving an app-internal URL should be wrapped in
```
Difegue, line 11: Is there a way we could potentially enforce this thru linter rules?  
I'm aware it's potentially impossible since we can't always know which URLs are internal and which aren't, but I thought I'd ask regardless.  

This class looks great to me but I am worried that it's easy to forget about 🤔

public/js/common.js
```diff
@@ -3,6 +3,41 @@
...
+        if (!this.load_url.startsWith("/")) {
+            console.trace("passed non-absolute URL to apiURL");
+            this.load_url = "/" + this.load_url;
+        }
+    }
+
+    toString() {
+        // in the default case, this will be empty string and will
+        // leave the load URL unchanged
+        let base_url = window.LRR_BASEURL;
```
Difegue, line 32: Is there a way we could do this with an additional header from the mojo/server side instead of requiring a property to be set in every template?  

There's already a lot of boilerplate in all the templates so it's not really a big deal, but I feel it'd be less code.

public/js/index.js
```diff
@@ -551,10 +551,10 @@ Index.handleContextMenu = function (option, id) {
         });
         break;
     case "read":
-        LRR.openInNewTab(`./reader?id=${id}`);
+        LRR.openInNewTab(new LRR.apiURL(`/reader?id=${id}`).toString());
```
Difegue, line 554: Is the `toString()` required here? I haven't been seeing it in other uses of the class like with fetch and the like

public/js/stats.js
```diff
@@ -26,6 +26,7 @@ Stats.initializeAll = function () {
                     return;
                 const namespacedTag = LRR.buildNamespacedTag(tag.namespace, tag.text);
                 const url = LRR.getTagSearchURL(tag.namespace, tag.text);
+
```
Difegue, line 29: leftover whitespace

tests/opds.t
```diff
@@ -49,6 +49,11 @@ $mojo->mock(
         my $tt = Template->new( { INCLUDE_PATH => $cwd . "/templates" } ) || die "$Template::ERROR\n";
         my $output;
 
+        # this mocks the Mojo url_for helper function, which is now
+        # used in the templates
+        $vars{c} = {
```
Difegue, line 54: Maybe one to put in mocks.pl?

Difegue (changes requested):
This looks good to me overall, I mostly have a few misc questions.  
Thanks for going through the trouble of updating all the URL refs in the app!  

I would like to see the documentation updated in step with this PR before merging it.  
(Although I'm not expecting a wall of text, probably just a few lines next to where we mention manual redis-server setup since both rely on lrr.conf)

public/js/common.js
alethiophile: I would love it if this were possible; you're right that this is easy to forget about. Unfortunately, I can't think of an easy way to do it. The app generates URLs in all sorts of ways from all sorts of places, including in inline HTML inside string constants. Most of those are path-only URLs that aren't necessarily programmatically distinguishable from any other string.

I don't know how a linter would even be able to determine what strings are URLs and what aren't, leaving aside sorting them into internal vs. external URLs.

public/js/index.js
alethiophile: The API documentation for window.open is ambiguous, but a quick test suggests that it is not required.

Also caught some more URL references I'd missed.

tests/opds.t
alethiophile: The mock for `url_for` is specific to this `render_to_string` mock, and needs to be in the same place.

It might be reasonable to remove this whole Mojo mock, including `render_to_string`, into mocks.pl, if we wanted to use it in some other test.

lib/LANraragi.pm
alethiophile: You're right. Probably not a material difference in processing time, but this does run every request, so it's good form.

public/js/common.js
alethiophile: There's no way to read an arbitrary header from the same-page request using Javascript. However, it turns out you can set a cookie and then read that.

I do like this better than putting it in template code. That's just another thing to maybe forget later.

alethiophile:
Regarding documentation, I'm happy to write a docs update. However, I can't immediately find where the documentation source is. Is it hiding in some subdirectory I'm missing?

public/js/common.js
Difegue: huh, I thought headers were accessible somehow but I'll admit I never tried to do it 🙇 The cookie approach sounds good to me.

lib/LANraragi.pm
Difegue: Well, now that we went with the cookie approach I wonder if the extra helper is necessary at all now..  
But I guess it doesn't really hurt keeping it in.

public/js/common.js
Difegue: Yup, fair enough!

Difegue:
The docs are hidden in https://github.com/Difegue/LANraragi/tree/dev/tools/Documentation -- Rest of the PR looks good to me as-is.

Difegue (approved):

alethiophile:
I've added some docs. I put it alongside the reverse proxy configuration, since it's tied in with that system and that's where I looked when I was trying to find this a few weeks ago.

Difegue:
Makes sense to me. Docs looks good, I'll be merging this.  
Thanks again!

kebien6020:
Hi. Not sure if this is being tracked somewhere else, but there is a missing quote introduced by this change in the "Add Archive" page, and it's pretty broken at least in firefox and chromium, because the `<a>` tag is going on for much longer than intended.

Difegue:
Thanks for the report, I'll have that fixed shortly.

PR #1053, opened 2024-08-13
Improve HDoujin info.txt parsing

HDoujinDownloader:
Currently, tags are only extracted from the `TAGS` field for HDoujin's info.txt files. I've updated the plugin to extract tags from other fields as well (artist, series, language, parody, etc.), and namespace them accordingly.


lib/LANraragi/Plugin/Metadata/Hdoujin.pm
```diff
@@ -1,107 +1,170 @@
-package LANraragi::Plugin::Metadata::Hdoujin;
+package LANraragi::Plugin::Metadata::HDoujin;
```
Difegue, line 1: Changing the namespace is fine but I believe you'll have to change the tests to match

lib/LANraragi/Plugin/Metadata/Hdoujin.pm
```diff
@@ -1,107 +1,170 @@
...
-        #Standard metadata
-        name        => "Hdoujin",
+        # Standard metadata
+        name        => "HDoujin",
         type        => "metadata",
-        namespace   => "Hdoujinplugin",
-        author      => "Pao",
-        version     => "0.5",
-        description => "Collects metadata embedded into your archives by HDoujin Downloader's json or txt files.",
+        namespace   => "hdoujinplugin",
```
Difegue, line 22: I would avoid changing this namespace however since it's what's used for keeping track of user preferences

lib/LANraragi/Plugin/Metadata/Hdoujin.pm
```diff
@@ -1,107 +1,170 @@
...
+    $logger->info("Sending the following tags to LRR: $tags");
+
+    return (tags => $tags);
+
+}
+
+#get_tags_from_hdoujin_json(decodedjson)
 #Goes through the JSON hash obtained from an info.json file and return the contained tags.
-sub tags_from_Hdoujin_json {
+sub get_tags_from_hdoujin_json_hash {
```
Difegue, line 165: I'm guessing the json variant can't have summaries?

Difegue (changes requested):
LGTM overall - I think the tests should be updated to reflect the new summary/description parsing here. Thanks!

lib/LANraragi/Plugin/Metadata/Hdoujin.pm
HDoujinDownloader: Done!

HDoujinDownloader:
Thank you for the feedback! 

I updated the JSON parser to read the summary and make it more consistent with the output from the TXT file parser. It was adding all the fields as tags (including titles and URLs), but I've limited it to a more relevant subset. I also updated it to work with different JSON configurations (the outer `manga_info` may or may not be present based on user settings). The namespace-related issues should be resolved now as well.

> I think the tests should be updated to reflect the new summary/description parsing here

Correct me if I'm wrong, but it doesn't look like there are any tests for this format right now. I could possibly add some.

Difegue:
Thanks! The JSON parser was a pretty old bit of code so I'm not surprised if it was worse than the txt version.  
There are indeed no specific tests for the HDoujin plugin - Adding some with samples like the other plugins have would be welcome, but that's not blocking me from merging this in the meantime.

Difegue (approved):

Boontato:
> I updated the JSON parser to read the summary and make it more consistent with the output from the TXT file parser. It was adding all the fields as tags (including titles and URLs), but I've limited it to a more relevant subset. I also updated it to work with different JSON configurations (the outer `manga_info` may or may not be present based on user settings). The namespace-related issues should be resolved now as well.

Thanks squiddy for working on this, I actually enjoyed that it would pull URLs since in mihon/tachi i could search nhentai codes and it would resolve because the url is part of the tags and it was useful at least for me.

when i saw this PR i was hoping that it would fix the ability for this plugin to pull the title too because right now im using a secondary plugin just to pull title information from the metadata file.

HDoujinDownloader:
@Boontato Oh! I didn't even notice plugins could specify a gallery title. I'll get that fixed and submit a new PR in a bit.

@Difegue What's your take on having URLs in the tags (e.g. `url:https://nhentai.net/g/XXXXXX/`)? If the use case is just being able to search by NHentai code, maybe there's a better way to do it.

Difegue:
You should use `source:nhentai.net/xxxx` tags, there's support for those in the browser extension and a few other spots.

Boontato:
Yes I have been using tag rules to convert url namespace to source namespaces.  mihon also allowed specifying which namespace to use to pull the url too.

Issue #893, opened 2023-11-03
Add a simple string filter to BatchOperations


fatecyx:
I can filter new mangas by executing js in the console. such as
document.querySelectorAll('label').forEach(label => {
  if (label.textContent.includes('🆕')) {
    label.previousElementSibling.checked = true;
  }
});

I hope to add it to the page functions.

Difegue:
I'm planning to eventually improve the batch selection experience via #570, but I wouldn't mind approving a PR with this short improvement in the interim.

Difegue:
Implemented through #826 .

Issue #570, opened 2021-12-27
Batch Operations features label:enhancement 


nikgtasa:
I really like how LRR picks up tags automatically most of the time but sometimes it doesn't and sometimes i need to do it again. Currently batch operations are offering barebones solution of giving you a huge list with no way of managing it. I have no idea how any of it works so i'm obviously qualified for making suggestions. 
![firefox_BE5IZ5dHDI](https://user-images.githubusercontent.com/38408810/147439174-6c44b4bf-a6f6-464c-b54c-b697498af4db.jpg)

nikgtasa:
![firefox_uTe758eyXo](https://user-images.githubusercontent.com/38408810/147439569-f9be4ddd-3c1b-4e95-95c5-bf63aab39de4.jpg)
Also something for categories.

Difegue:
There was a similar mockup on Discord a few weeks ago, except it'd use the existing index UI:  
![image](https://user-images.githubusercontent.com/8237712/147466909-19820124-85e7-4cd8-b8e6-69786468a6cb.png)  

I'd like to work towards this instead, as it'd bring more code reuse (although it also means I have to untangle the index JS even more to make it usable as a standalone component)

Difegue:
I'm starting to wonder if I shouldn't bring batch into the main index as an overlay or something that'd apply on the current search -- Even with this issue done, you'd still have to repeat your search on the batch tagging page to actually do the operation you want..

IceBreeze:
> whereas this would probably need a new API endpoint and queue up a Minion job to execute the command?

Yes, I think so too.

Abusing the "script plugin system", I have already wrote a concept for this which is limited to adding tags.
It can't be ported as is because it includes a search functionality. For example:

```
|title:HUNTER X|addtags:tag1, tag2
\filter:fantasy, -oneshot\addtags:tag1, tag2
```
`title:` searches for substrings in titles case-sensitively while `filter:` calls the usual search method. The first character of the string specifies the separator between the filter and the command, so that I can avoid escaping strings in the titles.
But I don't want to change how the search box works, so I don't mind to stick with the usual filter here.

I was planning to support search by file and folder and to add commands to remove tags and to call plugins, but I wasn't comfortable writing in the small textbox that is the argument of the script and this seemed like the obvious alternative to me.

jmhickman:
[...]

PR #1054, opened 2024-08-15
Added search option to batch operations

alethiophile:
This is an initial implementation of some basic filtering for batch operations, as previously discussed in e.g. https://github.com/Difegue/LANraragi/issues/893 and https://github.com/Difegue/LANraragi/issues/570.

This basically just copies the search UI from the index page to the batch page, then populates the batch archive list from the search results instead of the all-archive list. The old behavior is available behind the "Use all archives" button. The batch page's search bar has tag autocomplete as well; to facilitate this, I moved the loadTagSuggestions function from index into common. That specific function could be pulled out with very little change.

The search UI is not complete here; it lacks categories or any way to apply other search filters. I'm primarily looking for feedback on the basic concept. Personally, this version satisfies one of my big problems with the batch operation interface.


Difegue:
Given Batch is moving towards being paired to the Index and its searches with the multi-select feature, I don't think I want to give it its own search bar for the time being.  

I'll close this for now but I'm open for revisiting Batch UI once people get to use the new selection bits.

PR #1060, opened 2024-08-19
Add testing for RegexParse

siliconfeces:
RegexParse didn't have tests, so now it does. Also ensure that the title gets trimmed (previously there was a space at the end).


Difegue (approved):

Difegue:
Good stuff, ty.

PR #1063, opened 2024-08-24
Add support for .zst and .tar.zst archives

baysonfox:
As libarchive has now supported zstandard archives, this pull request adds zstd archive support ( .zst / .tar.zst ) for it.


Difegue (approved):

Difegue:
no idea who'd use these for manga but sure, why not.

PR #1065, opened 2024-08-27
Add i18n

Haibersut:
Taking into consideration the existing framework, I have briefly incorporated the corresponding i18n components. I have adapted much of the text within "index" and "config" to the i18n format; however, many template files have yet to be replaced.
image

There is still a significant amount of work to be done to complete the i18n translation. 

**I am sharing this to request assistance. Also looking for a better method to name JSON keys and would appreciate any suggestions.**

Additionally, since new dependencies have been introduced, the `package-lock.json` file will need to be updated if you wish to run tests.

For translating text within JavaScript files using i18n, you simply need to ensure that i18n.js and its dependencies are loaded prior to the JavaScript files that require translation. Then use `i18n.t()`.

For translating HTML text, you need to include the following three JavaScript files and place them before any other scripts in your HTML document.
```html
<script src="[% c.url_for("/js/vendor/i18next.min.js") %]" type="text/JAVASCRIPT"></script>
<script src="[% c.url_for("/js/vendor/i18nextHttpBackend.min.js") %]" type="text/JAVASCRIPT"></script>
<script src="[% c.url_for("/js/i18n.js?$version") %]" type="text/JAVASCRIPT"></script>
```
follow these examples and the modified files as guidance for replacement.

### 1. Replacing Text Content Using the `data-i18n` Attribute

By adding the `data-i18n` attribute to HTML elements, you can instruct the i18next library to replace the content of those elements with translated text.

```html
<p data-i18n="welcome.message"></p>
```

In JavaScript, after the translation is done, `i18next.t('welcome.message')` will replace the original text.

### 2. Replacing HTML Attributes Using the `data-i18n-attr` Attribute

The `data-i18n-attr` attribute allows you to insert translated text into specific HTML attributes. For example, to update a button's `title` attribute:

```html
<button data-i18n="button.save" data-i18n-attr="title">Save</button>
```

In JavaScript, the `setAttribute` method will be used to update the specified attribute.

### 3. Handling Option Elements within Select Elements

For `select` elements, both the `select` tag itself and its internal `option` elements need the `data-i18n` attribute for translation.

```html
<select data-i18n="select.options">
    <option data-i18n="option.first">First Option</option>
    <option data-i18n="option.second">Second Option</option>
</select>
```

In JavaScript, you should iterate through the `option` child elements of the `select` element and replace the text based on the `data-i18n` attribute.

### 4. Special Handling of Input Elements of Type Button

For `input` elements with a type of button, you need to set the translated text using the `value` attribute.

```html
<input type="button" data-i18n="input.button.label" value="Click me" />
```

In JavaScript, update the translated text by setting the `value` attribute of the `input` element.

### 5. Replacing HTML Content Using the `data-i18n-html` Attribute

In some situations, you might need to replace the entire HTML content rather than just the text. Use the `data-i18n-html` attribute for this purpose.

```html
<div data-i18n="div.content" data-i18n-html></div>
```

Following these guidelines will help you successfully and smoothly integrate i18n translation within your HTML content.

### DIFF
--- a/lib/LANraragi/Model/Archive.pm
+++ b/lib/LANraragi/Model/Archive.pm
@@ -114,7 +114,7 @@ sub generate_page_thumbnails {
 
     my $should_queue_job = 0;
 
-    for ( my $page = 1; $page <= $pages; $page++ ) {
+    for ( my $page = 2; $page <= $pages; $page++ ) {
         my $thumbname = ( $page - 1 > 0 ) ? "$thumbdir/$subfolder/$id/$page.$format" : "$thumbdir/$subfolder/$id.$format";
 
         unless ( $force == 0 && -e $thumbname ) {
@@ -379,18 +379,18 @@ sub delete_archive ($id) {
 
     # Remove matching data from the search indexes
     my $redis_search = LANraragi::Model::Config->get_redis_search;
-    $redis_search->zrem( "LRR_TITLES",       "$oldtitle\0$id" );
-    $redis_search->srem( "LRR_NEW",          $id );
-    $redis_search->srem( "LRR_UNTAGGED",     $id );
-    $redis_search->srem( "LRR_TANKGROUPED",  $id );
+    $redis_search->zrem( "LRR_TITLES", "$oldtitle\0$id" );
+    $redis_search->srem( "LRR_NEW",         $id );
+    $redis_search->srem( "LRR_UNTAGGED",    $id );
+    $redis_search->srem( "LRR_TANKGROUPED", $id );
     $redis_search->quit();
 
     # Remove from tanks/collections
-    foreach my $tank_id (LANraragi::Model::Tankoubon::get_tankoubons_containing_archive($id)) {
+    foreach my $tank_id ( LANraragi::Model::Tankoubon::get_tankoubons_containing_archive($id) ) {
         LANraragi::Model::Tankoubon::remove_from_tankoubon( $tank_id, $id );
     }
 
-    foreach my $cat (LANraragi::Model::Category::get_categories_containing_archive($id)) {
+    foreach my $cat ( LANraragi::Model::Category::get_categories_containing_archive($id) ) {
         my $catid = %{$cat}{"id"};
         LANraragi::Model::Category::remove_from_category( $catid, $id );
     }
--- a/lib/LANraragi/Utils/Minion.pm
+++ b/lib/LANraragi/Utils/Minion.pm
@@ -19,8 +19,6 @@ use LANraragi::Model::Upload;
 use LANraragi::Model::Config;
 use LANraragi::Model::Stats;
 
-use Data::Dumper;
-
 # Add Tasks to the Minion instance.
 sub add_tasks {
     my $minion = shift;
@@ -76,7 +74,7 @@ sub add_tasks {
 
             $logger->debug("Number of available cores for processing: $numCpus");
 
-            # Generate thumbnails for all pages -- Cover should already be handled
+            # Generate thumbnails for all pages -- Cover should already be handled in higher resolution
             my @keys = ();
             for ( my $i = 2; $i <= $pages; $i++ ) {
                 push @keys, $i;


public/locales/en.json
```diff
@@ -0,0 +1,207 @@
+{
```
Difegue, line 1: Finding a good naming scheme here is tricky..  

- It's convenient that since this just JSON we can group keys; I would suggest grouping keys by the place in the UI/app they're used -- subgroups should also be used for stuff like notification header/text. A "common" group for text that is often reused is also good to have.  

- For elements that are similar(notifications, errors), I would personally have them all start the same (`notification-xxx`), just so they're easy to parse in the list. 

- Key names themselves should probably all be formatted similarly.
  - For LRR, I use `snake_case` in the backend code, `kebab-case` for CSS and `camelCase` for frontend/javascript.  
  - Since loc keys are more of a frontend styling thing, I would tend to go for kebab-case just to align with CSS? That looks like what you've mostly done already and it feels fine to me. 

So, something like this maybe? I didn't go through everything but hopefully this can be good enough guidance:  
```
"index": {
    "add-archives": "Add Archives",
    "batch-operations": "Batch Operations",
    "settings": "Settings",
    [...]
    "notification-default-password": {
      "heading": "You're using the default password and that's super baka of you",
      "text": "<a href=\"{{loginUrl}}\">Login</a> with password \"kamimamita\" and <a href=\"{{configUrl}}\">change that shit</a> on the double.<br/>...Or just disable it! <br/>Why not check the configuration options afterwards, while you're at it?"
    },
    "notification-debug-mode": {
      "heading": "<i class=\"fas fa-bug\"></i> You're running in Debug Mode!",
      "text": "Advanced server statistics can be viewed <a href=\"{{url}}\">here.</a>"
    },
    "notification-welcome": {
      "heading": "Welcome to LANraragi {{version}}!",
      "text": "If you want to perform advanced operations on an archive, remember to just right-click its name. Happy reading!"
    },
    "context-menu": {
        "read": "Read",
        "download": "Download",
        "edit": "Edit Metadata",
        "delete": "Delete",
        "add-to-category": "Add to Category"
    },
   "error-tag-suggestions": "Couldn't load tag suggestions",
   "error-loading-categories": "Couldn't load categories",
   "error-loading-changelog": "Error getting changelog for new version",
   "error-finding-categories": "Error finding categories for {{id}}!",
},
"settings": {
   "admin-settings": "Admin Settings",
   "select-category": "Select a category to show the matching settings.",
   "save-settings": "Save Settings",
   "plugin-configuration": "Plugin Configuration",
   "database-backup-restore": "Database Backup/Restore",
   "global": {
       "config_global.html.tt2": "------Placeholder------",
       "site-title": "Site Title",
       "site-title-description": "The site title appears on most pages as...their title.",
       "motd-description": "Slang for Message of the Day. Appears on top of the main Library view.",
       [...]
    },
    "security":{
       "enable-password": "Enable Password",
       "password-description": "If enabled, everything that isn't reading will require a password.",
       "new-password": "New Password",
       "new-password-confirm": "New Password Confirmation",
       [...]
    },
    [...]
},
"common": {
    "destructive-operation": "This is a destructive operation!",
    "title": "Title",
    "tags": "Tags",
    "return-to-library": "Return to Library",
}
```

Difegue (changes requested):
Thank you for trying to tackle this on -- I'm aware it is a massive job.  

I think we can first aim for getting the most-used pages of the app TL'd, then slowly localize-ify the rest of the app over time.  
A lot of people seem to want a Chinese translation in-app, so I'm sure people will want to chip in. 😎  

This is a good opportunity to get the initial tools and guidelines in place though, so my comments are more about that to begin with.

public/js/i18n.js
```diff
@@ -0,0 +1,115 @@
...
+            document.dispatchEvent(new CustomEvent('i18nextInitialized'));
+        }
+    });
+}
+
+/**
+ * Update the translations of the page content
+ */
+function updateContent() {
+    document.querySelectorAll('[data-i18n]').forEach(el => {
```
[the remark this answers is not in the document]
Haibersut: Thank you for your response. I've been a bit occupied with other matters these past few days, so progress has been somewhat slow. I will strive to complete the overall translation, as well as adjust the replacement methods and JSON text, within the next week.

public/locales/en.json
siliconfeces: This is of course your call, but I'd still like to bike-shed a touch if that's cool.

Grouping/namespacing translations based on location haven't ended up well for me so far. In my experience, it tends to result in duplicate translations and makes it annoying if you move stuff between pages. Grouping on component can work better (like navigation, menus, forms, etc).

Difegue: My thinking behind the "common" group was to avoid duplicating translations, although I suppose it could become a clusterfuck if you end up having a lot of reused strings.  

Grouping on component feels like it'd make building the initial list much harder to write.. In the end I wouldn't really mind either option I suppose

Haibersut: I apologize for the recent delays, but I’ve still been working on modifying the relevant code as much as possible.  

I’ve want to continue using my own implementation for text replacement, as the translation methods provided by the official i18n framework don’t quite fit the way I’ve integrated it into my current setup.

Regarding the translated text, I’ve opted to use the "filename" as the parent key for all new entries.  This approach allows for better localization, even though it might result in some redundant translations.

PR #1068, opened 2024-08-31
Add title and source extraction for HDoujin info files

HDoujinDownloader:
Added title and source metadata extraction as discussed in #1053.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/HDoujin.pm
+++ b/lib/LANraragi/Plugin/Metadata/HDoujin.pm
@@ -9,8 +9,9 @@ use Mojo::JSON qw(from_json);
 
 # You can also use the LRR Internal API when fitting.
 use LANraragi::Model::Plugins;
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Plugin/Metadata/HDoujin.pm
```diff
@@ -271,9 +284,27 @@ sub normalize_namespace {
...
     }
+    elsif ($namespace eq "url") {
+        return "source";
+    }
 
     return $namespace;
 
 }
 
+sub format_source {
```
Difegue, line 295: You can replace this with `trim_url` from `LANraragi::Utils::String`:  

```
use LANraragi::Utils::String qw(trim_url);

[...]
$values = trim_url($values);

```

Difegue (changes requested):

Difegue:
Excellent, thank you!

PR #1071, opened 2024-09-01
plugin loader should do all the checks

IceBreeze:
I think we should rely on the plugin loader to check the required subs. This would make the code simpler.

With this PR we loose the visibility of the invalid plugins, but it's not a problem for the official ones.
If it's necessary, `get_plugins()` can also return a list of invalid plugins to be used inside the startup sub.


Difegue:
Sorry for the delay in reviewing this - The changes make sense to me, now that we have more robust error handling around plugins it's possible to be a bit less paranoid around the sub checks.  

Thanks!

PR #1074, opened 2024-09-07
Fix searched as grabed text

natanane:
fixing an issue where the URL used for searching on nh was garbled.


Difegue:
Makes sense to me. Thanks!

PR #1077, opened 2024-09-15
[nHentai] Add support to set upload date in timestamp field using epoch stamp

shinji257:
Adds support for retrieving and using the epoch upload date stamp in the json data and including it as a ~~date_uploaded~~ timestamp tag.  I also added a toggle so that if a user doesn't want it then it doesn't bother to parse it out.  This is consistent with other plugins that support this functionality.


Difegue:
Looks good to me. Thanks!

Issue #873, opened 2023-10-23
Request a rating system


kokkeng1:
I checked it on the discussion board, but it's too old an issue

**Please describe your suggestion, and the problem it'd solve.**

First of all, thank you for developing a good app.
I saw a comment saying that the rating system has low priority and should be managed with tags.
But it's very strange that the Content Manager app doesn't have a scoring system. I want you to know that there is a demand to distinguish whether this was good or not by horoscope because a person's memory is limited.
It is very cumbersome to modify the tag you suggested before. Since the app doesn't have a tag preset, I always have to add an emoji to modify it.

**Additional context**
I request modify the score with a click in the Archive Overview menu.

Difegue:
👋 I still maintain this can and should be handled thru a tag instead of spinning off yet another DB field, as it makes sorting by rating "just work" with the current design.  

A rating control could be added to the edit UI to quickly add/update/remove a matching `rating:` tag, I suppose.

alok5050:
Hey, I am new to this and would like to contribute. Do you have any pointers on how to get started with this issue?

Difegue:
Sure! Thanks for stepping up 🙏 I believe this can be done thru frontend only. 
You can use Codespaces to spin up a development version of the app, and just upload any random pdf or zip with one image in it to get an entry you can add metadata to. 

You can look at https://github.com/Difegue/LANraragi/blob/dev/templates/edit.html.tt2 and https://github.com/Difegue/LANraragi/blob/dev/public/js/edit.js for the HTML/JS code used by the edit metadata page.  

Tags in LRR are in a simple comma-delimited format, so implementing this should be as easy as adding a rating control that when interacted with, adds a `rating:⭐⭐⭐⭐,` tag to the textfield. If a rating tag is already present, it should probably update it with the matching star count.  
I suppose the rating should go from 1 to 5 stars, and selecting 0 stars should remove the tag from the textfield.

chu-shen:
Below is an example that works for me by [editing the UI](https://github.com/chu-shen/LANraragi/blob/4ee1b9e04866f7602eab7445f4f09c67dd1270de/templates/index.html.tt2#L247) and using [plugins](https://github.com/chu-shen/LANraragi/blob/4ee1b9e04866f7602eab7445f4f09c67dd1270de/lib/LANraragi/Plugin/Metadata/Rating.pm)

[...]

PR #1079, opened 2024-09-22
Added Rating option on contextMenu

Hanga7yr:
Added a rating option on the contextMenu.

When an option is selected it updates the tags of the selected archive and add or modifies the rating tag to the selected stars.
if selected the white one, it will remove the rating tag.

Just made it in a few hours so it's probably not really good but it's a start.
Changes are only client side and doesn't need many more things done unless the rating is added to the archive thumb view.

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -525,6 +525,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
 );
 
 /**
+ * Build rating options for contextMenu and select the one for the current ID.
+ * @param {*} id The ID of the archive to check
+ * @returns Ratings
+ */
+Index.loadContextMenuRatings = (id) => Server.callAPI(`/api/archives/${id}/metadata`, "GET", null, `Error finding metadata for ${id}!`,
+    (data) => {
+        const items = {};
+        const ratings = [{
+            id: 1,
+            name: "⭐",
+        }, {
+            id: 2,
+            name: "⭐⭐",
+        }, {
+            id: 3,
+            name: "⭐⭐⭐",
+        }, {
+            id: 4,
+            name: "⭐⭐⭐⭐",
+        }, {
+            id: 5,
+            name: "⭐⭐⭐⭐⭐",
+        }];
+        const tags = data.tags.split(",").map((x) => x.split(":"));
+        const hasRating = tags.some((x) => x[0] === "rating");
+
+        items.push({
+            name: "",
+            events: {
+                click() {
+                    Server.updateTagsFromArchive(id, tags.filter((x) => x[0] !== "rating").map((x) => x.join(":")).join(","));
+                },
+            },
+        });
+
+        for (let i = 0; i < ratings.length; i++) {
+            items[i] = ratings[i];
+            items[i].events = {
+                click() {
+                    if (hasRating) {
+                        const ratingIndex = tags.findIndex((x) => x[0] === "rating");
+                        tags[ratingIndex][1] = ratings[i].name;
+                    } else {
+                        tags.push(["rating", ratings[i]]);
+                    }
+
+                    Server.updateTagsFromArchive(id, tags.map((x) => x.join(":")).join(","));
+                },
+            };
+        }
+
+        return items;
+    },
+);
+
+/**
  * Handle context menu clicks.
  * @param {*} option The clicked option
  * @param {*} id The Archive ID
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -533,45 +533,38 @@ Index.loadContextMenuRatings = (id) => Server.callAPI(`/api/archives/${id}/metad
     (data) => {
         const items = {};
         const ratings = [{
-            id: 1,
+            name: "☆"
+        }, {
             name: "⭐",
         }, {
-            id: 2,
             name: "⭐⭐",
         }, {
-            id: 3,
             name: "⭐⭐⭐",
         }, {
-            id: 4,
             name: "⭐⭐⭐⭐",
         }, {
-            id: 5,
             name: "⭐⭐⭐⭐⭐",
         }];
-        const tags = data.tags.split(",").map((x) => x.split(":"));
-        const hasRating = tags.some((x) => x[0] === "rating");
-
-        items.push({
-            name: "",
-            events: {
-                click() {
-                    Server.updateTagsFromArchive(id, tags.filter((x) => x[0] !== "rating").map((x) => x.join(":")).join(","));
-                },
-            },
-        });
+        const tags = LRR.splitTagsByNamespace(data.tags);
+        const hasRating = Object.keys(tags).some(x => x === "rating");
+        const ratingValue = hasRating ? tags["rating"] : [0];
 
         for (let i = 0; i < ratings.length; i++) {
             items[i] = ratings[i];
+            items[i].type = "checkbox";
+
+            if (items[i].name === ratingValue[0]) { items[i].selected = true; }
             items[i].events = {
                 click() {
-                    if (hasRating) {
-                        const ratingIndex = tags.findIndex((x) => x[0] === "rating");
-                        tags[ratingIndex][1] = ratings[i].name;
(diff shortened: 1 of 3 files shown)


Hanga7yr:
#873

Hanga7yr:
Adds functionality requested on #873

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
...
+            id: 3,
+            name: "⭐⭐⭐",
+        }, {
+            id: 4,
+            name: "⭐⭐⭐⭐",
+        }, {
+            id: 5,
+            name: "⭐⭐⭐⭐⭐",
+        }];
+        const tags = data.tags.split(",").map((x) => x.split(":"));
```
Difegue, line 551: You can use `LRR.splitTagsByNamespace(data.tags)` here, then directly look if the "rating" key exists to get the existing rating tag. I think it should also make the underlying logic easier to write.

templates/index.html.tt2
```diff
@@ -233,6 +233,11 @@
 							"sep1": "---------",
 							"edit": { name: "Edit Metadata", icon: "fas fa-pencil-alt" },
 							"delete": { name: "Delete", icon: "fas fa-trash-alt" },
+							"rating": {
+							    "name": "Add Rating",
+							    "icon": "",
```
Difegue, line 238: Needs an icon. How about `fas fa-star` ?

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
...
+        const hasRating = tags.some((x) => x[0] === "rating");
+
+        items.push({
+            name: "",
+            events: {
+                click() {
+                    Server.updateTagsFromArchive(id, tags.filter((x) => x[0] !== "rating").map((x) => x.join(":")).join(","));
+                },
+            },
+        });
```
Difegue, line 561: This block fails for me on Firefox as `push` only works with arrays, not objects.  
Also, I'm not quite sure what this click event even does?

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
...
+
+        for (let i = 0; i < ratings.length; i++) {
+            items[i] = ratings[i];
+            items[i].events = {
+                click() {
+                    if (hasRating) {
```
Difegue, line 567: It would be nice to show in the UI the current rating, if there already was one.  
Maybe a checkmark can be added to the matching item or something?

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
     },
 );
 
+/**
+ * Build rating options for contextMenu and select the one for the current ID.
+ * @param {*} id The ID of the archive to check
+ * @returns Ratings
+ */
+Index.loadContextMenuRatings = (id) => Server.callAPI(`/api/archives/${id}/metadata`, "GET", null, `Error finding metadata for ${id}!`,
```
Difegue, line 532: I'm not a massive fan of making a new API request to get metadata again when it's already technically present in the page.. But it'd be a bit annoying to hook into the datatables part to copy the tags somewhere so it's fine for now.

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
...
+    (data) => {
+        const items = {};
+        const ratings = [{
+            id: 1,
```
Difegue, line 536: The `id` parameters are unnecessary here, but you'll need to add `type:` if you want the click events to work.  
`type:"checkbox"` works, and you could precheck the checkbox that matches the current rating tag if there's one.

public/js/server.js
```diff
@@ -325,3 +325,11 @@ Server.deleteArchive = function (arcId, callback) {
         })
         .catch((error) => LRR.showErrorToast("Error while deleting archive", error));
 };
+
+/**
+ * Sends a UPDATE request for the metadata of the archive ID
+ * @param {*} arcId Archive ID
+ */
+Server.updateTagsFromArchive = function (arcId, tags) {
+    Server.callAPIBody(`/api/archives/${arcId}/metadata`, "PUT", { tags }, `Updated tags from Archive ${arcId}!`, "Error updating tags from archive", null);
```
Difegue, line 334: APIBody needs a `FormData` object, you can't just pass the API a random string and expect it to work.  

```suggestion
    const formData = new FormData();
    formData.append("tags", tags);

    Server.callAPIBody(`/api/archives/${arcId}/metadata`, "PUT", formData, `Updated tags from Archive ${arcId}!`, "Error updating tags from archive", null);
```

public/js/index.js
```diff
@@ -524,6 +524,62 @@ Index.loadContextMenuCategories = (catList, id) => Server.callAPI(`/api/archives
...
+            items[i].events = {
+                click() {
+                    if (hasRating) {
+                        const ratingIndex = tags.findIndex((x) => x[0] === "rating");
+                        tags[ratingIndex][1] = ratings[i].name;
+                    } else {
+                        tags.push(["rating", ratings[i]]);
+                    }
+
+                    Server.updateTagsFromArchive(id, tags.map((x) => x.join(":")).join(","));
```
Difegue, line 574: Once this goes through, the tags displayed in the main search index don't get updated.  
Not quite sure what'd be the best way to fix this... Maybe just reloading the table would be enough.

Difegue (changes requested):
Is this AI-generated code? I'm afraid there are quite a few logic errors...  

I did get it to work after a few tweaks though, so I believe this can still be salvaged.

Hanga7yr:
Hi,
No, it was not AI, however, now that I look around better, there seems to be some changes that are not reflected on the commit.
It must have been me tweaking around as I did have some problems cleaning up the commit, because most changes you point I believe I encountered too.

I will try to have a proper update soon. Sorry to have you waste time looking at this mess.
I will, however, use `LRR.splitTagsByNamespace(data.tags)` now that I now of it.

Difegue:
Ah, apologies for the false accusation in that case. 🙇 
There's no problem having multiple commits in the PR for cleanup if necessary, we can just squash things in the end. 🤠  

Good luck!

templates/index.html.tt2
Hanga7yr: Fixed on new commit

public/js/index.js
Hanga7yr: Fixed on new commit.

public/js/index.js
Hanga7yr: Yep, tried to get the data but it was quite messy.

public/js/index.js
Hanga7yr: Fixed on new commit.

public/js/server.js
Hanga7yr: Fixed on new commit.

public/js/index.js
Hanga7yr: Fixed on new commit.

public/js/index.js
Hanga7yr: Fixed on new commit.

Hanga7yr:
> Ah, apologies for the false accusation in that case. 🙇 There's no problem having multiple commits in the PR for cleanup if necessary, we can just squash things in the end. 🤠
> 
> Good luck!

I uploaded a fix of the commit, it resolves most of the problems while adding the improvement of `LRR.splitTagsByNamespace(data.tags)`.
It didn't take long as it was something I already did.
The only thing that I couldn't really fix was removing the API call.

Hope everything is alright now.

Difegue (approved):
Looks good now! Sorry for the delay in re-reviewing this.

Difegue:
Using the blank star emoji for "no rating/remove rating" is a bit odd so I might tweak that later, but the base featureset is good. Thanks!

Issue #1005, opened 2024-06-28
Feature Request: Add api for upload archive 


nuclear06:
I have reviewed the documentation, but I dont find an API for uploading archive. Using token to upload through this url `http://example.com/upload` is also ineffective (must login by password at first).

API for uploading would be helpful for writing scripts to automate tasks such as downloading from somewhere and uploading. Although Lanraragi provides plugin mechanism to achieve these purposes, unfortunately, I am not familiar with Perl. : (

I'm a bit puzzled ; I searche issue and discussion but not find similar suggestions.
Although wirte a script to login by password is simple, but why not API? : )

PR #1085, opened 2024-10-05
Implement upload API

psilabs-dev:
Implementation of an archive upload API, that uploads the content of an archive and all metadata, if any, to the LRR server. (#1005)

While it is possible to write a LRR plugin for a website W, it is often the case to use dedicated, feature-rich, and actively maintained downloaders (e.g. PixivUtil2 for Pixiv) that specifically pulls metadata/archives from W. However, moving downloaded content from these downloaders to LRR is still a human process. In the case of PixivUtil2, metadata is stored in a sqlite db, and the storage/language/architecture is different for each downloader.

This API implementation (using multipart/form-data encoding) allows users to perform language-agnostic archive migrations or mirrors from their existing archive servers and downloaders to the LRR server via the API, or write comparatively thin HTTP client wrappers around their downloaders that negotiate with and push to LRR during that application's runtime. The user may wish to attach the SHA1 hash of the file to check for file integrity and avoid corruptions during transit.
```
PUT /api/archives/upload
```

Return values (see documentation)

Include in body: file content (binary), filename, title (optional), summary (optional), tags (optional), category ID (optional), plugin calls (?)

Example of a `PUT /api/archives/upload` in Python with the "example.zip" archive:
```python
import hashlib
import requests

def calculate_sha1(file_path):
    sha1 = hashlib.sha1()
    with open(file_path, 'rb') as fb:
        while chunk := fb.read(8192):
            sha1.update(chunk)
    return sha1.hexdigest()

file_path = "example.zip" # required
file_name = "example.zip" # required, this will be where the file is stored in the server
upload_mime = "application/zip"

# metadata
title = "example title"
tags = "example_date,key:value,artist:example artist"
summary = "example summary"
category_id = None

# calculate checksum (optional)
file_checksum = calculate_sha1(file_path)

data = {
    "title": title,
    "tags": tags,
    "summary": summary,
    "file_checksum": file_checksum
}

with open(file_path, 'rb') as fb:
    files = {'file': (file_name, fb, upload_mime)}
    response = requests.put(
        "http://localhost:3000/api/archives/upload",
        files=files,
        data=data
    )
    print(response.json())
```

~~This is a synchronous API implementation. Theoretically, we can also run this API as a Minion job but this is not a long-running task and doesn't justify creating a job, also error handling is more streamlined and the API should not take too many requests to begin with (TBD).~~


psilabs-dev:
Open for review and testing. I've tested this on ~50k cbz archives with metadata (without obvious problems). I've also tested this in combination with/wo automatic metadata plugins, checksum validation, optional metadata (title, tags, summaries), mandatory fields (filename, file data), localhost/reverse proxy.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -107,6 +114,171 @@ sub serve_file {
...
+        sub {
+            return if -d $_;
+            $tempfile = $File::Find::name;
+            $filename = $_;
+        },
+        $tempdir
+    );
+
+    # utf downgrade (see LANraragi::Utils::Minion)
+    unless (utf8::downgrade($filename, 1)) {
```
Difegue, line 213: Since this API doesn't run through a minion job, you might not need the atrocious utf hack.

Difegue (changes requested):
This looks good at a glance to me - I'll have to play with it for a bit.  

Would it be possible to get the documentation updated?  
This would land in https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/api-documentation/archive-api.md - Please duplicate an existing {swagger} element and edit it to match. Even if it's not perfect I'll help with suggestions afterwards 👍

Difegue:
I guess one additional thought: You mention only one upload can be done at a time with this API, but I'm not seeing any obvious limitations that would prevent running it sequentially.  
Is there something that breaks when you call it multiple times in parallel? If so, it might be good to add a lock of some sort to prevent further API calls when an upload is in progress.

psilabs-dev:
Ahh you're right, LRR does support concurrent API calls now that I tested it again. I don't know why last time for some reason my second worker would always hang and lose connections...

After testing concurrent uploads, I'm seeing two (probably fixable) issues with concurrency.

Type 1: when the same archive is uploaded concurrently by two clients, on rare occasions the server will throw a "The file couldn't be renamed in your content folder: No such file or directory" error to the later client. This can happen when the two uploads differ by as much as 30ms.

2) this is the wierder bug characterized by this response:
```html
<!DOCTYPE html>

<head>
	<title>LANraragi Server Error</title>

	<meta name="viewport" content="width=device-width" />
	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
<meta name="apple-mobile-web-status-bar-style" content="black" />
<meta name="mobile-web-app-capable" content="yes" />
<meta name="apple-mobile-web-app-capable" content="yes" />

	<link type="image/png" rel="icon" href="favicon.ico" />
	<link rel="manifest" href="app.webappmanifest" />
</head>

<body style='background: none repeat scroll 0% 0% brown; color: white; font-family: sans-serif; text-align: center'>

	<img src='/img/flubbed.gif' />
	<br />
	<h2>I flubbed it while rendering your page!</h2>
	It's likely there's a bug with the server at this point.<br />
	<p>
		 at /home/koyomi/lanraragi/script/../lib/LANraragi/Utils/Archive.pm line 336.

	</p>

	<h3>Some more info below if available:</h3>
	<pre>{
  &quot;action&quot; =&gt; &quot;create_archive&quot;,
  &quot;controller&quot; =&gt; &quot;api-archive&quot;
}
</pre>

</body>

</html>
```
I've noticed that when one worker encounters this error, the other worker will encounter a concurrency error of the first type. This is more rare than the first type of error

psilabs-dev:
> This looks good at a glance to me - I'll have to play with it for a bit.
> 
> Would it be possible to get the documentation updated? This would land in https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/api-documentation/archive-api.md - Please duplicate an existing {swagger} element and edit it to match. Even if it's not perfect I'll help with suggestions afterwards 👍

Sure! I'll take a look at this

psilabs-dev:
I added an upload documentation, though I'm not finding a lot of info online about the `% swagger-parameter` format. I tried editing through gitbook, it seems to change a lot of things though.

The `query` parameters should actually be part of the request body in JSON form ([request body](https://swagger.io/docs/specification/v3_0/describing-request-body/describing-request-body/)), and the API must take a filename and file binary argument as part of the [multipart](https://swagger.io/docs/specification/v3_0/describing-request-body/multipart-requests/) format.

psilabs-dev:
Ugh... there is a race condition that's happening when two identical uploads are being processed at almost the exact same time. I thought adding a redis lock would fix it but then implementing it causes more issues it seems, Utils::Archive will complain at 242 🧐

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: I've re-added this, since redis (used for locking) seems to throw errors when certain files are not downgraded.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -188,11 +218,14 @@ sub create_archive {
...
     if ( !$upload->move_to($tempfile) ) {
+        $logger->error("Could not move uploaded file $filename to $tempfile");
+        $redis->del("upload:$filename");
+        $redis->quit();
         return $self->render(
             json => {
                 operation   => "upload",
                 success     => 0,
-                error       => "Couldn't move uploaded file to $tempfile."
+                error       => "Server Error"
```
Difegue, line 228: Not the biggest fan of generic "server error" messages which are always unhelpful - especially since we have the reason for the error here.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -239,41 +256,47 @@ sub create_archive {
...
     if ( $success_status==0 ) {
         $status = 500;
-        if ( index($message, "Unsupported File Extension") != -1 ) {
-            $status = 415;
-        } elsif ( index($message, "Enable replace duplicated archive in config to replace old ones") != -1 ) {
-            $status = 409;
-        } elsif ( index($message, "The file couldn't be moved to your content folder") != -1 ) {
-            $status = 500;
+        my $error = "Server Error";
+        if ( index($message, "Enable replace duplicated archive in config to replace old ones") != -1 ) {
```
Difegue, line 266: Doing a string comparison from the user-facing message of `handle_incoming_file` feels very brittle.. Maybe we should have different values for `success_status` instead? Like 1, 2, -1.. Or maybe even http error codes at this point.

Difegue: Alternatively I guess we could just return 400 for all errors and rely on clients to look at `message` for details -- That's how other endpoints in the server do it.

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: I can change this back. My opinion is that server-side errors should not be exposed to the client for security reasons/necessity, but this obviously is a tradeoff with debugging.

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: Yeahhhh I can't think of a really good way to do this, this was just one implementation I thought of. I'm tending towards returning all 400, but that mixes in server-side errors and passes the brittleness over to the client-side.

An option that removes the need for string comparison is to copy the logic of `handle_incoming_file` to the controller level and inject response handling there, which I'm fine doing but might look ugly. Or refactor and break apart `handle_incoming_file` so that it can be used by the controller.

Adding an http status code return to `handle_incoming_file` can work in a backwards-compatible way. Otherwise I don't have a super strong opinion on these approaches.

psilabs-dev: decided to fix this by replacing `handle_incoming_file` with http status codes, and doing status code => success_status mapping for minion downstream

lib/LANraragi/Controller/Api/Archive.pm
Difegue: Right, I do agree it's better to avoid exposing information about the server file structure or similar.  
Would `Couldn't move uploaded file to temporary location` work as a compromise? The detailed error is already logged server-side so I'd be fine with that.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -259,46 +259,25 @@ sub create_archive {
...
-            return $self->render(
-                json => {
-                    operation   => "upload",
-                    success     => $success_status,
-                    error       => $error
-                },
-                status => $status
-            );
-        }
+    unless ( $success_status ) {
```
Difegue, line 262: The success_status variable feels a bit superfluous with this now.
```suggestion
    unless ( $status_code == 200 ) {
```

Difegue (changes requested):

lib/LANraragi/Controller/Api/Archive.pm
Difegue: Yep, that's perfectly fine for me. We'll probably want to mention those various status code in the documentation.

psilabs-dev:
Ok, don't think I'll have anything else to add, I'll be doing final builds and stress tests and opening it up for review later.

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: No complaints here, should be reflected in the new commits.

Difegue:
Sounds good, feel free to undraft once you think it's ready and I'll give it one last passover.

Difegue:
No, if there's no limitations on the server I wouldn't implement an artificial one in the API.  
I imagine people running this behind nginx or similar will encounter the same session size limits they already had with the non-API version.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -107,6 +114,173 @@ sub serve_file {
...
+    $redis->expire( "upload:$filename", 10 );
+
+    # metadata extraction
+    my $catid           = $self->req->param('category_id');
+    my $tags            = $self->req->param('tags');
+    my $title           = $self->req->param('title');
+    my $summary         = $self->req->param('summary');
+
+    # return error if archive is not supported.
+    if ( !is_archive($filename) ) {
```
Difegue, line 195: You don't really need this bit if it's already done in `handle_incoming_file`, right? 

(I'm aware Controller->Upload does the same, but I think I'll eventually deprecate that code path in favor of just using the API in the webclient)

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -107,6 +114,173 @@ sub serve_file {
...
+            },
+            status => 415
+        );
+    }
+
+    # Move file to a temp folder (not the default LRR one)
+    my $tempdir                 = tempdir();
+
+    my ( $fn, $path, $ext )     = fileparse( $filename, qr/\.[^.]*/ );
+    my $byte_limit              = LANraragi::Model::Config->enable_cryptofs ? 143 : 255;
```
Difegue, line 212: I wonder if this shouldn't be enforced at the `handle_incoming_file` level rather than both in here, in Model->Upload->Download_url, and the old Controller upload.  

Might be one for later tho

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -107,6 +114,173 @@ sub serve_file {
...
+                status => 417
+            );
+        }
+    }
+
+    my $filename        = $upload->filename;
+    my $uploadMime      = $upload->headers->content_type;
+
+    # utf downgrade (see LANraragi::Utils::Minion)
+    $filename = encode('UTF-8', $filename);
```
Difegue, line 161: Why is this necessary? To me it looks like you're just passing the filename through utf-8 and back out for no real reason now.

tools/Documentation/api-documentation/archive-api.md
```diff
@@ -444,6 +444,112 @@ Current page to update the reading progress to. **Must** be a positive integer,
...
+Title of the Archive.
+{% endswagger-parameter %}
+{% swagger-parameter name="tags" type="string" required="false" in="query" %}
+Tags of the Archive.
+{% endswagger-parameter %}
+{% swagger-parameter name="summary" type="string" required="false" in="query" %}
+summary
+{% endswagger-parameter %}
+{% swagger-parameter name="category_id" type="int" required="false" in="query" %}
+Category ID of the Archive.
```
Difegue, line 463:
```suggestion
Category ID you'd want the archive to be added to.
```

tools/Documentation/api-documentation/archive-api.md
```diff
@@ -444,6 +444,112 @@ Current page to update the reading progress to. **Must** be a positive integer,
...
+{% swagger-description %}
+Upload an Archive to the server.
+If a SHA1 checksum of the Archive is included, the server will perform an optional in-transit, file integrity validation, and reject the upload if the server-side checksum does not match.
+{% endswagger-description %}
+
+{% swagger-parameter name="title" type="string" required="false" in="query" %}
+Title of the Archive.
+{% endswagger-parameter %}
+{% swagger-parameter name="tags" type="string" required="false" in="query" %}
+Tags of the Archive.
```
Difegue, line 457:
```suggestion
Set of tags you want to insert in the database alongside the archive.
```

tools/Documentation/api-documentation/archive-api.md
```diff
@@ -444,6 +444,112 @@ Current page to update the reading progress to. **Must** be a positive integer,
...
+{% endswagger-parameter %}
+{% swagger-parameter name="summary" type="string" required="false" in="query" %}
+summary
+{% endswagger-parameter %}
+{% swagger-parameter name="category_id" type="int" required="false" in="query" %}
+Category ID of the Archive.
+{% endswagger-parameter %}
+{% swagger-parameter name="file_checksum" type="string" required="false" in="query" %}
+SHA1 checksum of the Archive.
```
Difegue, line 466:
```suggestion
SHA1 checksum of the archive for in-transit validation.
```

Difegue (changes requested):
One last round of comments from me after reviewing the whole thing anew.

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: tbh I don't think I can confidently explain this part of the code because I don't get perl encoding, it was a product of experimentation and was necessary because it prevented errors and ensured that API uploads behave the same as manual uploads (i.e. `LANraragi::Controller::Upload::process_upload`). I'll describe a timeline explaining the rationale.

For the minimal example consider an archive with a special character called "～.cbz". Note that this type of file can already be uploaded successfully through the manual upload portal. (It goes without saying that the behavior of the manual upload dictates the expected behavior of the API upload.)

The entire utf encode/downgrade block exists to guard against issues with redis when a file with a special character like this is uploaded. Redis locking is required before any writes to guard against race conditions. Without the utf block, an unsanitized filename with special characters will cause redis to throw an exception.

When utf-downgrade is (re)introduced, any upload that would've incurred a redis exception is now "handled" at the downgrade stage, which returns an HTTP error. All filenames that pass this downgrade test also do not trigger redis errors. However, this does not *fix* the issue that a file with special characters *should not* fail a downgrade in the API at all, when it never failed this stage during a manual upload, so why does this happen

During the manual process, the file is written to a temporary location first, and the filename is extracted using perl's built-in `File::Find` module. This means somewhere perl already performed encoding/normalization work on the file, which is not present in the API, hence it passing the downgrade stage. (Alternatively, this encoding may be happening during the javascript phase of the upload.)

> to verify my assertion that `File::Find` performs encoding work (which I haven't done since I think the issue was resolved), you can do this: in the API stage, take the file bytes and filename and write them to a temporary directory and read it again with `File::Find` overwriting the filename. If I'm correct then this should pass utf downgrade.

The obvious solution would be to just copy the `File::Find` logic of the manual upload to the API, but the API method cannot afford to move the file to disk due to Redis, as it needs to happen before any writes. So a way to do this is to emulate the effect of `File::Find` and manually encode the file, hence the utf encode stage. Adding this stage makes the API have expected behavior.

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: `is_archive` ensures that the filename has a valid extension. I don't know how removing this will affect the tempdir logic, I'll need to test it out

lib/LANraragi/Controller/Api/Archive.pm
Difegue: Well, it's mostly there for synology hardware which has that 255 byte cap, and the notably insane who use cryptoFS on top of that, which limits it down to a further 143 bytes.  

I believe this would break even temporary filenames if exceeded.

lib/LANraragi/Controller/Api/Archive.pm
Difegue: huh, I guess is_pdf wasn't written to handle invalid filenames at all and the issue just never popped up - good to know.

lib/LANraragi/Controller/Api/Archive.pm
Difegue: right, the File::Find way was designed to make sure we get the accurate filesystem representation of the file, since filesystems can and will mangle names.  

But here, you can't wait for that since you need to lock the database on the filename... If the encode/downgrade works I don't mind it I suppose, I don't want to hang the PR forever over this.  

Although... have you tried just using `redis_encode`/`redis_decode`? Those are the intended methods for treating strings before putting them in/taking them out of the database, the downgrade is very much a nuclear/unsafe option.

psilabs-dev: > Although... have you tried just using `redis_encode`/`redis_decode`? Those are the intended methods for treating strings before putting them in/taking them out of the database, the downgrade is very much a nuclear/unsafe option.

Ah... their existence kind of just left my mind 😅 replaced them and tested it, seems like there's no issue. Unfortunately/admittedly, I'm not knowledgable enough about encodings

psilabs-dev: Thanks for taking the time to look over all this, sorry if the PR has gotten too long.

Difegue: No problem at all - Thanks for having put up with my review for all this time.

Difegue (approved):

Difegue:
This is good for me - the leftover comments aren't essential and I might get to them at some point during a refactor.  

Thanks again for the work!

PR #1091, opened 2024-10-10
Named params

IceBreeze:
This PR introduces named parameters for the plugins while maintaining backwards compatibility.
It should also allow for conversion of existing parameters without losing the configuration.

To help you in the review, basically `get_plugin_parameters()` now return an hash instead of an array.
It can be feed directly to the new plugins entry point, but, for the old plugins, it contains the key `customargs` with the array of values from the DB.
The choice of what to use is made inside the `exec_*_plugin` subs.

Some issues I want to point out:

- the hashes `$lrr_info` and `$params` can be merged, but this requires adding checks to avoid collisions between the names of internal keys and those chosen by the plugin contributors. Alternatively you could change the names of the "internal" keys (eg: `archive_id` to `_archive_id`), but this will break the compatibility.
- `oneshot_param` is now duplicated in `$params` as `$params->{'oneshot'}` because I think it should be moved there if we keeps the two hashes separated.
- login plugins load, but I don't have an account to fully test them

### DIFF
--- a/lib/LANraragi/Utils/Plugins.pm
+++ b/lib/LANraragi/Utils/Plugins.pm
@@ -140,6 +140,12 @@ sub get_plugin_parameters {
     } elsif ( ref( $pluginfo{parameters} ) eq 'HASH' ) {
 
         my %params = $redis->hgetall($namerds);
+
+        # TODO: param conversion block, remove after deprecation period
+        if ( $pluginfo{to_named_params} && exists $params{customargs} ) {
+            %params = convert_to_named_params_and_persist( $redis, $namerds, $pluginfo{to_named_params}, %params );
+        }
+
         %args = map { $_ => redis_decode( $params{$_} ) } keys %params;
 
     }    # else { # should we die here? }
@@ -194,6 +200,28 @@ sub use_plugin {
     }
 
     return ( \%pluginfo, \%plugin_result );
+}
+
+sub convert_to_named_params_and_persist {
+    my ( $redis, $namerds, $old_params_order, %params ) = @_;
+    my $customargs = redis_decode( $params{customargs} );
+    my $logger     = get_logger( "Plugin System", "lanraragi" );
+
+    $logger->info("converting $namerds to named parameters ...");
+
+    #Decode it to an array for proper use
+    if ($customargs) {
+        my @args = @{ decode_json($customargs) };
+        while ( my ( $idx, $key ) = each @{$old_params_order} ) {
+            $params{$key} = $args[$idx];
+            $redis->hset( $namerds, $key, $params{$key} );
+        }
+        $logger->info("conversion completed: removing 'customargs' value");
+        $redis->hdel( $namerds, 'customargs' );
+        delete $params{customargs};
+    }
+
+    return %params;
 }
 
 1;
--- a/lib/LANraragi/Utils/Plugins.pm
+++ b/lib/LANraragi/Utils/Plugins.pm
@@ -127,28 +127,39 @@ sub get_plugin_parameters {
 
         # Replace with saved values if they exist
         if ( $redis->hexists( $namerds, "enabled" ) ) {
-            my $argsjson = $redis->hget( $namerds, "customargs" );
-            $argsjson = redis_decode($argsjson);
+            my $saved_config = redis_decode( $redis->hget( $namerds, "customargs" ) );
 
             #Decode it to an array for proper use
-            if ($argsjson) {
-                @args = @{ decode_json($argsjson) };
+            if ($saved_config) {
+                @args = @{ decode_json($saved_config) };
             }
         }
         $args{customargs} = \@args;
 
     } elsif ( ref( $pluginfo{parameters} ) eq 'HASH' ) {
 
+        # Fill with default values first
+        %args = map { $_ => $pluginfo{parameters}{$_}{default_value} } keys %{ $pluginfo{parameters} };
+
         my %params = $redis->hgetall($namerds);
 
         # TODO: param conversion block, remove after deprecation period
         if ( $pluginfo{to_named_params} && exists $params{customargs} ) {
             %params = convert_to_named_params_and_persist( $redis, $namerds, $pluginfo{to_named_params}, %params );
+        } elsif ( exists $params{customargs} && !$pluginfo{to_named_params} ) {
+            my $logger = get_logger( "Plugin System", "lanraragi" );
+            $logger->warn( 'An old configuration for the plugin "'
+                  . $pluginfo{name}
+                  . '" was detected, but the plugin version you are using does not specify the conversion key "to_named_params"'
+                  . ' required for the upgrade. This will cause the old configuration to be lost when saving.' );
+            $logger->warn( 'customargs => ' . redis_decode( $redis->hget( $namerds, "customargs" ) ) );
         }
 
-        %args = map { $_ => redis_decode( $params{$_} ) } keys %params;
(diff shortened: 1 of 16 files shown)


tools/Documentation/plugin-docs/login.md
```diff
@@ -14,12 +14,11 @@ When executing your Plugin, LRR will call the `do_login` subroutine and give it
 ```perl
 sub do_login {
 
-    #First lines you should have in the subroutine
-    shift;
-    my ($param1, $param2, $param3) = @_; # Plugin parameters
+    #First line you should have in the subroutine
+    my (undef, $params) = @_; # Plugin parameters
```
Difegue, line 18: Same here, I think the `shift;` way is easier to understand and preps contributors if they ever fancy working on the full codebase which uses this pattern everywhere.  

(I guess one might argue that `shift` is a bit of a magic word either when you don't know that perl subs pass the module reference as the first argument and that shift just eats it, but it is how it is..)

tools/Documentation/plugin-docs/metadata.md
```diff
@@ -152,3 +154,37 @@ sub get_tags_from_somewhere {
 
 1;
 ```
+
+### Converting existing plugins to named parameters
+
```
Difegue, line 159: I would mention some history here, as in that plugins used to rely on an ordered array before 0.9.3 and that has changed to named parameters.  

For the time being, it's also perfectly fine to remain on ordered array parameters if plugin devs don't want to update -- User settings won't be lost.

tools/Documentation/plugin-docs/login.md
```diff
@@ -65,18 +64,17 @@ sub plugin_info {
...
     my $ua = Mojo::UserAgent->new;
 
-    if ($loginenabled) {
+    if ($params->{loginenabled}) {
 
         $ua->cookie_jar->add(
             Mojo::Cookie::Response->new(
                 name   => 'userID',
-                value  => $uID,
+                value  => $params->{uID},
```
Difegue, line 77: Are hash keys case-sensitive? I don't believe so but it might be better to show the same casing between parameter declaration and usage in the examples.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -114,25 +114,44 @@ sub get_plugin_parameters {
...
+        my %params = $redis->hgetall($namerds);
+
+        # TODO: param conversion block, remove after deprecation period
+        if ( $pluginfo{to_named_params} && exists $params{customargs} ) {
+            %params = convert_to_named_params_and_persist( $redis, $namerds, $pluginfo{to_named_params}, %params );
+        }
+
+        %args = map { $_ => redis_decode( $params{$_} ) } keys %params;
+
+    }    # else { # should we die here? }
```
Difegue, line 151: Don't see why you'd need to die here, parameter-less plugins can still exist.  

Although we might want to log a warning on the if above when `customargs` still exists yet the plugin doesn't have a `to_named_params` conversion method written, telling the user that the plugin is malformed and old settings might be lost.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -114,25 +114,44 @@ sub get_plugin_parameters {
...
-    foreach my $param ( @{ $pluginfo{parameters} } ) {
-        push( @args, $param->{default_value} );
-    }
+    if ( ref( $pluginfo{parameters} ) eq 'ARRAY' ) {
 
-    # Replace with saved values if they exist
-    if ( $redis->hexists( $namerds, "enabled" ) ) {
-        my $argsjson = $redis->hget( $namerds, "customargs" );
-        $argsjson = redis_decode($argsjson);
+        my @args;
```
Difegue, line 121: I would give this variable a different name, it's a bit confusing having a hash and an array with the same name in the scope of a single sub.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -114,25 +114,44 @@ sub get_plugin_parameters {
...
+    } elsif ( ref( $pluginfo{parameters} ) eq 'HASH' ) {
+
+        my %params = $redis->hgetall($namerds);
+
+        # TODO: param conversion block, remove after deprecation period
+        if ( $pluginfo{to_named_params} && exists $params{customargs} ) {
+            %params = convert_to_named_params_and_persist( $redis, $namerds, $pluginfo{to_named_params}, %params );
+        }
+
+        %args = map { $_ => redis_decode( $params{$_} ) } keys %params;
```
Difegue, line 149: I think this needs to fallback to `$param->{default_value}` if there wasn't a config key saved in redis for a parameter, like in the array version.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -164,13 +183,14 @@ sub use_plugin {
         %pluginfo = $plugin->plugin_info();
 
         # Get the plugin settings in Redis
-        my @settings = get_plugin_parameters($plugname);
+        my %settings = get_plugin_parameters($plugname);
+        $settings{oneshot} = $input;
```
Difegue, line 187: To reply to the question in the PR description:  

I wouldn't merge `lrr_info` and `params`, I think it's useful to keep a clear distinction between data that's dynamically provided by the server depending on the archive in play (name, thumbnail/id/tags, etc), and static-ish plugin configuration set by the user in the settings.  

The oneshot parameter in more of a one-time parameter set by the user for the current archive, so it can fit in both hashes -- I think it's fine having it in `params` as a result.  

(although since we're keeping backwards compatibility here, they'll stay duplicated for a while? I wouldn't really mind having it stay in lrr_info only either.)

Difegue (changes requested):
Sorry for the delay in reviewing this, I wanted to give it an in-depth look and needed to free up enough time for that 😅  

I think the change makes sense and helps with plugin dev -- I wouldn't have vetted it without backwards compatiblity but as it stands it's good for inclusion.  
I have a few comments regarding cleanup and formatting but nothing major. (except the `default_value` handling!)

IceBreeze:
> Sorry for the delay in reviewing this, I wanted to give it an in-depth look and needed to free up enough time for that 😅

This is not a job :smile: 

There would be one last change that I wanted to keep out of the PR.

I was thinking of moving the creation of the logger in the caller (`exec_*_plugin`) and passing it to the plugin in the `$lrr_info` parameter.
In this way you could use the same logger to write the exception message without having to do it in the plugin itself and you could also move the log line with the list of extracted tags (`Sending the following tags to LRR:...`) to the caller, making the log more consistent.
This is also due to the fact that the log files are separate.

On the other hand it will be necessary to pass the log as a parameter between the methods if it is needed.
What do you think about it?

Difegue:
I think it's more logical to have plugins create their own loggers so it's clear what's coming from the server and what's coming from external code.

Difegue:
This looks good to me otherwise, great work!

Guerra24:
Just FYI this has changed the API, it returns both an array or dictionary/object for the parameters.

image

IceBreeze:
> Just FYI this has changed the API, it returns both an array or dictionary/object for the parameters.

Thanks for letting me know!

I took a look at the docs and the API `GET .../api/plugins/:type` seems the only one involved. Moreover I don't see a way to use or save the parameters other than to posting a `multipart/form-data`.
So I don't think the parameters are useful as they are returned by the API right now, but I also don't like this inconsistency. Because of this i think the fix should be to stick with the array implementation of  the `parameters` property and move the parameter name in a property called `name` alongside `desc` and `type`.

Issue #1084, opened 2024-10-03
Filename Parsing Plug-in function request


hucurve:
![image](https://github.com/user-attachments/assets/a8939a8a-0e52-4692-809f-d792707a5045)
I'm using the built-in Plug-in for file metadata recognition, which I think is very useful and robust, but I still have two pain points that I hope to resolve one day：
1.Only the first can be recognized when multiple artists are present, while the others are in others
My file name =[黒ねずみいぬ, 市川和秀, 猪去バンセ, カサイこーめい, きしぐま, ＳＵＶ, 重丸しげる, ちんぱん☆Mk-Ⅱ, ばんじゃく, 英, ふぁい, 水樹 凱, やさごり]XXXX.zip
![image](https://github.com/user-attachments/assets/bd42cf48-4d64-4bc8-86e6-b9ad137b2d2d)

2.I often change the original file name, and after re-scanning, the changed file will not trigger the plug-in, but still use the original name. There is also no way to select the changed file in the batch tool. This often confuses me

I hope someone can see my request, anyway I still like LRR this beautiful software:)

Difegue:
This is a bit unconventional as far as using the naming standard goes, but I think it'd be pretty easy to check if the tags contain commas after regex parsing and assign the namespace to each one.

IceBreeze:
I was toying with this plugin a bit and this was an issue that I wasn't sure whether I wanted to handle or not.
This could also be applied to "groups", "series" and "languages" as well, but I knew for certain that at least the series can have a comma in their names.

I was going to break down just the language string where I can verify the generated tokens, but I can do the same for the rest and blindly accept the result.
It should be refined, but I can make a PR to dicuss.

Difegue:
I think it makes sense to have this behavior for language and groups.  
Series can have commas, but for those I think we should take them out as it'll break the tag storage.  

I do wonder if groups with commas in their names exist.. I don't think that'd be too common.

PR #1096, opened 2024-10-20
extends RegexParse plugin

IceBreeze:
This is my proposal for #1084.

I tried to do some permutations on the file name, but it would be better to have some real and more complex names to substitute in the tests.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/RegexParse.pm
+++ b/lib/LANraragi/Plugin/Metadata/RegexParse.pm
@@ -162,7 +162,7 @@ sub _classify_item {
     # if the namespace is specified, we are able to exclude some common words,
     # otherwise we are dealing with simple tags
     if ( $namespace && $COMMON_EXTRANEOUS_VALUES{ lc $item } || looks_like_number($item) ) {
-        return $PLUGIN_TAG_NS . lc $item;
+        return $PLUGIN_TAG_NS . $item;
     }
     return "${namespace}${item}";
 }
--- a/lib/LANraragi/Plugin/Metadata/RegexParse.pm
+++ b/lib/LANraragi/Plugin/Metadata/RegexParse.pm
@@ -15,45 +15,6 @@ use Scalar::Util               qw(looks_like_number);
 
 my $PLUGIN_TAG_NS = 'parsed:';
 
-# consider using Locale::Language / Locale::Script
-my %VALID_LANGUAGES = (
-    'chi'      => 'chinese',    # ?
-    'chinese'  => 'chinese',
-    'de'       => 'german',
-    'deu'      => 'german',
-    'en'       => 'english',
-    'eng'      => 'english',
-    'english'  => 'english',
-    'es'       => 'spanish',
-    'fr'       => 'french',
-    'fra'      => 'french',
-    'fre'      => 'french',
-    'french'   => 'french',
-    'ger'      => 'german',     # ?
-    'german'   => 'german',
-    'it'       => 'italian',
-    'ita'      => 'italian',
-    'italian'  => 'italian',
-    'ja'       => 'japanese',
-    'japanese' => 'japanese',
-    'jpn'      => 'japanese',
-    'ko'       => 'korean',
-    'kor'      => 'korean',
-    'korean'   => 'korean',
-    'pl'       => 'polish',
-    'pol'      => 'polish',
-    'polish'   => 'polish',
-    'ru'       => 'russian',
-    'rus'      => 'russian',
-    'russian'  => 'russian',
-    'spa'      => 'spanish',
-    'spanish'  => 'spanish',
-    'zh'       => 'chinese',
-    'zh'       => 'chinese',
-    'zho'      => 'chinese',
-    'textless' => 'textless'
-);
-
 my %COMMON_EXTRANEOUS_VALUES = (
     'uncensored' => 1,
     'decensored' => 1,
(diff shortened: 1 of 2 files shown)


lib/LANraragi/Plugin/Metadata/RegexParse.pm
```diff
@@ -115,7 +226,7 @@ sub get_tags {
 #(\(([^([)]+)\))? returns the content of (Series). Optional.
 #(\[([^]]+)\])? returns the content of [Language]. Optional.
 #\s* indicates zero or more whitespaces.
-my $regex = qr/(\(([^([]+)\))?\s*(\[([^]]+)\])?\s*([^([]+)\s*(\(([^([)]+)\))?\s*(\[([^]]+)\])?/;
+my $regex = qr/(\(([^([]+)\))?\s*(\[([^]]+)\])?\s*([^([]+)\s*(\(([^([)]+)\))?\s*(\[([^]]+)\])?(?<tail>.*)?/;
```
Difegue, line 229: Please document what this extra bit does in the mega-comment above, this regex is really unreadable without it so it needs to stay up to date😅  

From what I can gather this just keeps the remaining trailing part of the filename for the rest of the plugin to execute

lib/LANraragi/Plugin/Metadata/RegexParse.pm
```diff
@@ -23,34 +74,49 @@ sub plugin_info {
...
         author      => "Difegue",
-        version     => "1.0",
+        version     => "1.0.1",
         description =>
           "Derive tags from the filename of the given archive. <br>Follows the doujinshi naming standard (Release) [Artist] TITLE (Series) [Language].",
         icon =>
           "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAYAAACNiR0NAAAAAXNSR0IArs4c6QAAAL1JREFUOI1jZMABpNbH/sclx8DAwPAscDEjNnEMQUIGETIYhUOqYdgMhTPINQzdUEZqGIZsKBM1DEIGTOiuexqwCKdidDl0vtT62P9kuZCJEWuKYWBgYGBgRHbh04BFDNIb4jAUbbSrZTARUkURg6lD10OUC/0PNaMYgs1Skgwk1jCSDCQWoBg46dYmhite0+D8pwGLCMY6uotRDOy8toZBkI2HIhcO/pxCm8KBUkOxFl/kGoq3gCXFYFxVAACeoU/8xSNybwAAAABJRU5ErkJggg==",
-        parameters => [ ]
+        parameters => [
+            { type => "bool", desc => "Capture trailing tags in curly brackets" },
```
Difegue, line 83: I would give an example of the naming scheme caught by this either here or in the main plugin description, it's a bit confusing out of the box otherwise.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
```diff
@@ -23,34 +74,49 @@ sub plugin_info {
...
+            {   type => "bool",
+                desc => "Keep everything you catch as tags in the namespace \"${PLUGIN_TAG_NS}\"<BR />"
```
Difegue, line 85: It's unclear for end users what will or will not end up with `parsed:` prepended to it with or without this parameter enabled. 

From what I can gather:  
- specific "common extraneous values" will always get the `parsed:` namespace  
- languages we couldn't figure out will always get the `parsed:` namespace (I think this should be changed as per my comment above)  
- trailing tags in curly brackets will get the `parsed:` namespace if this option is enabled  

Is that correct? I'm not sure how to explain it better atm but I can try thinking about a better description.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
```diff
@@ -64,42 +130,87 @@ sub get_tags {
...
-        #Special case for circle/artist sets:
-        #If the string contains parenthesis, what's inside those is the artist name
-        #the rest is the circle.
-        if ( $artist =~ /(.*) \((.*)\)/ ) {
-            push @tags, "group:$1";
-            push @tags, "artist:$2";
+sub parse_language_value {
+    my ($language) = @_;
+    my @tags;
+    my @maybe_languages = map { trim( lc $_ ) } split( m/,/, $language );
```
Difegue, line 170: This feels too complex to have in a parsing plugin - I would just scrap the VALID_LANGUAGES part entirely and encourage the use of tag rules for users that want to coalesce their language tags.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
Difegue: I also think this one would make sense to default to `true`.

Difegue (changes requested):
huh, I certainly wasn't expecting that many changes for an issue that was just about splitting artist names. 😅 

I think most of the changes are OK but I would worry about making the parsing too complex - ideally users should still be relatively aware of what the plugin will spit out of their filenames.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
IceBreeze: I thought about leaving it like this to keep the old behavior, but if you are sure I'll change the default.

About the description I'll see what I can do. This is the hardest part :rofl:

lib/LANraragi/Plugin/Metadata/RegexParse.pm
IceBreeze: What I noticed is that there are often useful categorizations that are found beyond the standard part or even that if the standard part does not exist, especially for different media, there can still be a pattern traceable between similar media. So it could be convenient for the advanced user if we extract these labels making them manageable with the rules.

By default the plugin doesn't do anything different from before, but by enabling the advanced options the user can use the namespace `parsed:` to easily identify and rewrite the labels that can be of interest by creating some ad hoc tag rules.
What remains can be cleaned up by putting `-parsed:*` as the last rule.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
IceBreeze: You're right, but I thought that since they are quite easily identifiable, we could save a few lines of rules.

IceBreeze: On second thought, the language map is ugly to look at and replacing it with a library is excessive.
On the other hand how many languages ​​would one ever want to manage?
You've convinced me :smiley:

Difegue (approved):

PR #1098, opened 2024-10-23
Add Title parsing and updating to ComicInfo

shinji257:
Enables the ability to push the title found in ComicInfo files to LRR and update it accordingly.  I think this should be controlled via the global toggle for if plugins should update the title.


shinji257:
Looks like a unit test has to be made.  I don't really know unit tests but I can try to take a stab at it I guess.  I tested it against actual files over here.

Difegue:
Code looks accurate - I believe all you might have to do regarding testing is to modify the existing test to check that the plugin now returns the title as well. Pretty sure the existing ComicInfo sample already has a Title element set.

shinji257:
Yup.  That's what I'm trying to figure out.  Don't mind the extra commits.  I'll squash it all once I get the tests to pass. ;)

shinji257:
Ok.  Commit mess squashed into a single commit.  Unit tests added to original commit and they pass.  Took a page from another set of unit tests to sort out how I had to update things.

Difegue:
perfect, nice work.

PR #1105, opened 2024-10-30
Add Pixiv summary

psilabs-dev:
Add summary extraction for Pixiv's metadata plugin.


Difegue (changes requested):
Looks good - Would it be possible to amend the test sample so it checks and adds a summary?

psilabs-dev:
yep on its way

PR #1106, opened 2024-11-01
fix file upload permissions

psilabs-dev:
Currently when the LRR docker image is run, the contents folder is where archives are saved, and any file uploads will fail due to a lack of permission bc the contents folder in docker is not owned by koyomi. This commit rectifies that.

Checking 01-lrr-setup I saw that permfixes were applied but only to thumb, I'm not sure why only the thumb folder is checked. I could be missing something here

### DIFF
--- a/tools/build/docker/s6/cont-init.d/01-lrr-setup
+++ b/tools/build/docker/s6/cont-init.d/01-lrr-setup
@@ -24,10 +24,10 @@ chmod -R u+rwx /home/koyomi/lanraragi/database
 
 if [ "$FIX_PERMS" -eq 1 ]; then
   echo "Fixing permissions, hold on!"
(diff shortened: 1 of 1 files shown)


Difegue:
This was an intentional choice as some users don't trust the server to not tamper with their files, so the docker setup is made to technically work with a read-only content folder.  

Now this obviously makes file uploads, downloads and deletion from the server not work, but this was considered an OK tradeoff at the time.  

Auto-changing permissions on the thumbnail folder itself is fine, since that's not meant to contain user data. (And if we move on with completely separating it from the content folder instead of the weird double mount situation that's currently in place, this bit of script will make more sense I think)

psilabs-dev:
Ok, I'll close this, given this is intended behavior

PR #1107, opened 2024-11-01
Implement Duplicate Finder

dix0nym:
Implement a feature to detect and delete duplicate archives in the library based on the hamming distance between thumbnail hashes (https://github.com/Difegue/LANraragi/issues/338).

The minion job is parallelized and is way faster than the initial plugin, for my library (27k archives, 730GB) a full run took about 1h.
Additionally, I improved upon the script instead of only return duplicate pairs (e.g. we have 3 duplicate archives: id1 -> id2, id1 -> id3, id1 -> id3, now it returns one group [id1, id2, id3]).

view in `/duplicates`,  it is currently not connected from anywhere else. Should be considered if it should be linked to from navbar or settings. Furthermore, the CSS is defined inline in the HTML, maybe it should be separated into separate file?

view-duplicates

I could not figure out how to do the tooltip for the tags like it is on index page. Would be great to see the actual tags if you hover over the tag count. I would need some help there.

I implemented different ways to delete the duplicates:
- manual - one by one
- selecting based on attribute
  - tag count
  - file size
  - page count
  - date

select

A batch delete endpoint in the API would be great, the toasts are a bit annoying if you delete some archives in batch.

Tested on my library, detected and deleted 776 duplicate groups.

Would appreciate any feedback.

### DIFF
--- a/lib/LANraragi/Utils/Minion.pm
+++ b/lib/LANraragi/Utils/Minion.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Encode;
+use Mojo::JSON qw(encode_json);
 use Mojo::UserAgent;
 use Parallel::Loops;
 
@@ -137,6 +138,100 @@ sub add_tasks {
             };
 
             $job->finish( { errors => \@errors } );
+        }
+    );
+
+    $minion->add_task(
+        find_duplicates => sub {
+            my ( $job,      @args )  = @_;
+            my ( $threshold ) = @args;
+
+            my $logger = get_logger( "Minion", "minion" );
+            my $redis  = LANraragi::Model::Config->get_redis;
+            my @keys   = $redis->keys('????????????????????????????????????????');
+            $redis->quit();
+
+            $logger->info("Starting find duplicate job (threshold = $threshold)");
+
+            my $numCpus = Sys::CpuAffinity::getNumCpus();
+            my $pl      = Parallel::Loops->new($numCpus);
+
+            $logger->debug("Number of available cores for processing: $numCpus");
+
+            # Gather thumbhashes
+            my %thumbhashes;
+            foreach my $id (@keys) {
+                my $thumbhash = $redis->hget( $id, "thumbhash" );
+                $thumbhashes{$id} = $thumbhash if $thumbhash;
+            }
+
+            # Prepare to track visited nodes
+            my %visited :shared;
+            my @ids = keys %thumbhashes; # List of IDs to check
+
+            # Share the %visited hash across threads
+            $pl->share( \%visited );
+
+            my @sections = split_workload_by_cpu( $numCpus, @ids );
+            eval {
+                $pl->foreach(
+                    \@sections,
+                    sub {
+                        my $redis = LANraragi::Model::Config->get_redis;
+
+                        foreach my $id (@$_) {
+                            # Skip if this ID has already been processed in another thread
+                            next if $visited{$id};
+                            my @stack = ($id);
+                            my @group;
+
+                            while (@stack) {
+                                my $node = pop @stack;
+                                next if $visited{$node};
+
+                                # Mark the node as visited
+                                {
+                                    lock(%visited);
+                                    $visited{$node} = 1;
+                                }
+                                push @group, $node;
+
+                                # Find all potential duplicates for this node
+                                foreach my $other_id (keys %thumbhashes) {
+                                    next if $node eq $other_id || $visited{$other_id};
+
+                                    # Calculate Hamming distance
+                                    my $distance = 0;
+                                    for (my $i = 0; $i < length($thumbhashes{$node}); $i++) {
+                                        $distance++ if substr($thumbhashes{$node}, $i, 1) ne substr($thumbhashes{$other_id}, $i, 1);
+                                        last if $distance > $threshold;  # Early exit if threshold exceeded
+                                    }
+
+                                    # If within threshold, add to stack for further exploration
+                                    if ($distance <= $threshold) {
+                                        $logger->debug("Found potential duplicate: $node and $other_id with distance $distance");
+                                        push @stack, $other_id;
+                                    }
+                                }
+                            }
+
+                            # Add the discovered group to redis
+                            # to avoid redudnant groups in different orders - sort and composite key
+                            if (@group && scalar @group >= 2) {
+                                @group = sort @group;
+                                my $composite_key = join '', map { substr($_, 0, 10) } @group;
+                                my $group_json = encode_json(\@group);
+                                $logger->debug("duplicate group '$composite_key': $group_json");
+                                $redis->hset("duplicate_groups", "dupgp_$composite_key", $group_json);
+                            }
+                        }
+
+                        $redis->quit();
+                    }
+                );
+            };
+
+            $job->finish( { } );
         }
     );
 
--- a/public/js/duplicates.js
+++ b/public/js/duplicates.js
@@ -1,0 +1,191 @@
+/**
+ * Duplicate Operations.
+ */
+const Duplicates = {};
+
+Duplicates.dt = {};
+
+
+Duplicates.initializeAll = function () {
+    // bind events to DOM
+    $(document).on("click.goback", "#goback", () => { window.location.replace("./"); });
+    $(document).on("mouseenter.thumbnail-wrapper", ".thumbnail-wrapper", (e) => $(e.currentTarget).find(".thumbnail-popover").show());
+    $(document).on("mouseleave.thumbnail-wrapper", ".thumbnail-wrapper", (e) => $(e.currentTarget).find(".thumbnail-popover").hide());
+
+    $(document).on("click.find-duplicates", ".find-duplicates", Server.findDuplicates);
+    $(document).on("click.clear-duplicates", ".clear-duplicates", () => { window.location.href = new LRR.apiURL("/duplicates?delete=1"); });
+    $(document).on("click.delete-archive", ".delete-archive", Duplicates.deleteArchive);
+    $(document).on("click.delete-selected", ".delete-selected", Duplicates.deleteArchives);
+
+    $(document).on("change.duplicate-select-condition", ".duplicate-select-condition", Duplicates.conditionChange);
+    Duplicates.inizializeDataTable();
+}
+
+Duplicates.drawCallbackDataTable = function (settings) {
+    var groupColumn = 0;
+    var api = this.api();
+    var rows = api.rows({ page: 'current' }).nodes();
+    var lastGroup = null;
+
+    // Iterate over the data once to insert group rows at end of each group
+    api.column(groupColumn, { page: 'current' })
+        .data()
+        .each(function (group, i) {
+            if (lastGroup && lastGroup !== group) {
+                $(rows).eq(i).before(
+                    '<tr class="separator"><td colspan="10" style="padding: 0px;"></td></tr>'
+                );
+            }
+            lastGroup = group;
+        });
+}
+
+Duplicates.inizializeDataTable = function () {
+    Duplicates.dt = $('#ds').DataTable({
+        dom: '<"table-control-wrapper" <"search-box" f><"length-box" l>><t><p>',
+        // avoid sorting columns as it messes with the grouping
+        columns: [
+            { title: 'Group-Key', visible: false },
+            { title: '', orderable: false },
+            { title: '', orderable: false },
+            { title: 'Title', orderable: false },
+            { title: 'Pages', orderable: false },
+            { title: 'Filename', orderable: false },
+            { title: 'Filesize', orderable: false },
+            { title: 'Date', orderable: false },
+            { title: 'Tags', orderable: false },
+            { title: 'Action', orderable: false }
+        ],
+        order: [[0, 'asc']],
+        autoWidth: false,
+        pageLength: 10,
+        deferRender: true,
+        drawCallback: Duplicates.drawCallbackDataTable
+    });
+};
+
+Duplicates.compareDuplicates = function (rows, field, fieldType, order = 'desc') {
+    var values = [];
+    var rowToExclude = null;
+
+    // Determine comparator and starting value based on order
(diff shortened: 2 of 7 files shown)


Difegue (changes requested):
Thanks a lot for this! Looks delightful.  

I'll give this a more indepth review later, but I don't think the API should have any "batch" kind of endpoints if possible - Single/Atomic operations are easier to code on the server-side and clients can just loop a bunch of calls..  

If the toasts are annoying I believe it'd be useful to add an optional parameter to the JS function for API calls to disable showing them on a successful operation.

chu-shen:
Please add ignore option to the archive. For example, if you want to preserve duplicate archives with different translation styles.

public/js/server.js
```diff
@@ -326,6 +326,42 @@ Server.deleteArchive = function (arcId, callback) {
...
+ * Sends a POST request to queue a find_duplicates job,
+ * detecting archive duplicates based on their thumbnail hashes.
+ */
+Server.findDuplicates = function () {
+    Server.callAPI(`/api/find_dupes`, "POST",
+        "Queued up a job to find duplicates! Stay tuned for updates or check the Minion console.",
+        "Error while sending job to Minion:",
+        (data) => {
+            // Disable the buttons to avoid accidental double-clicks.
+            $(".find-duplicates").prop("disabled", true);
```
Difegue, line 339: I think it'd also be good to feature a spinner of some form in the UI to indicate the job is in progress.  

Now of course that all kinda goes away if the user reloads the page while the job is running -- I wonder if we shouldn't store the minion job ID somewhere so it can be detected on a reload. (Either on the controller side with the database, or just with localstorage? Controller is probably better but that'd be a bit annoying with the generic API call)

public/js/server.js
```diff
@@ -326,6 +326,42 @@ Server.deleteArchive = function (arcId, callback) {
...
+
+            // Check minion job state periodically while we're on this page
+            Server.checkJobStatus(
+                data.job,
+                true,
+                (d) => {
+                    $(".find-duplicates").prop("disabled", false);
+                    LRR.toast({
+                        heading: "All duplicates found! Encountered the following errors:",
+                        text: d.result.errors,
```
Difegue, line 349: I don't think the duplicate job can return errors, right?

public/js/duplicates.js
```diff
@@ -0,0 +1,191 @@
...
+        confirmButtonColor: "#d33",
+    }).then((result) => {
+        if (result.isConfirmed) {
+            $("table tbody tr").each(function () {
+                const row = $(this);
+                const isChecked = row.find(".form-check-input").is(":checked");
+                const dataId = row.find(".delete-archive").attr("data-id");
+
+                if (isChecked && dataId) {
+                    Server.deleteArchive(dataId, () => { Duplicates.dt.row(row).remove().draw() });
```
Difegue, line 182: As mentioned in the previous comments, it'd be fine to add an optional boolean to `deleteArchive` to make it not emit the success toast.  

(The success toast doesn't really makes sense here as well since it doesn't actually reload the page, too)

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+                            <td>[% archive.name %]</td>
+                            <td class="file-size">
+                                [% arcsize_mb = archive.arcsize / (1024 * 1024) %]
+                                [% arcsize_mb FILTER format('%.2f') %] MB
+                            </td>
+                            <td class="date-added">[% archive.date_added %]</td>
+                            <td class="tag-count">
+                                <i class="fa-solid fa-tag"></i>
+                                [% archive.tags.split(",").size %]
+                            </td>
```
Difegue, line 225: If you want the tags tooltip to work, I think you could do something like this:  
```suggestion
                            <td class="tag-count tag-tooltip" onmouseover="IndexTable.buildTagTooltip(this)" >
                                <i class="fa-solid fa-tag"></i>
                                [% archive.tags.split(",").size %]
                            </td>
                            <div class="caption caption-tags" style="display: none;" >
                                    ${LRR.buildTagsDiv("[% archive.tags %]")}
                            </div>
```

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+
+	<script src="[% c.url_for("/js/vendor/jquery.min.js") %]" type="text/JAVASCRIPT"></script>
+    <script src="[% c.url_for("/js/vendor/jquery.dataTables.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/preact.umd.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/hooks.umd.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/compat.umd.js") %]" type="text/JAVASCRIPT"></script>
+	<script>window.React = window.preactCompat; window.react = window.preactCompat;</script>
+	<script src="[% c.url_for("/js/vendor/clsx.min.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/react-toastify.umd.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/jqcloud.min.js") %]"></script>
```
Difegue, line 31: Don't think jqcloud and allcollapsible are used here

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+        .duplicate-control-btn-group {
+            margin-bottom: 15px;
+        }
+    </style>
+</head>
+
+<body>
+	<div class='ido' style='text-align:center; overflow-x:auto;'>
+		<h2 class="ih" style="text-align:center">Duplicates</h2>
+        <p>Found [% duplicates.size %] duplicate groups</p>
```
Difegue, line 155: There should be some explanatory text here that details how the feature works. 

("Found X duplicate groups" might be good to hide if there's no duplicate results? Not necessary though imo)

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+
+        [% IF userlogged %]
+        <div class="control-btn-group">
+            <div class="duplicate-control-btn-group">
+                <button type="button" class="stdbtn find-duplicates">Find Duplicates</button>
+                <button type="button" class="stdbtn clear-duplicates">Clear Duplicates</button>
+            </div>
+            <div class="select-btn-group">
```
Difegue, line 163: I'd hide this div if there are no duplicate groups found/available.

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+                            </td>
+                            <td>
+                                <div class="thumbnail-wrapper">
+                                    <a href="/reader?id=[% archive.arcid %]" title="[% archive.title %]">
+                                        <img class="thumbnail" src="/api/archives/[% archive.arcid %]/thumbnail" width="100"/>
+                                    </a>
+                                    <div class="thumbnail-popover">
+                                        <img src="/api/archives/[% archive.arcid %]/thumbnail" />
+                                    </div>
+                                </div>
```
Difegue, line 212: I'd use the same tooltip mechanism as the main index here instead:  

```suggestion
                                <div class="thumbnail-wrapper">
                                    <a onmouseover="IndexTable.buildImageTooltip(this)" href="${new LRR.apiURL('/reader?id=[% archive.arcid %]')}" title="[% archive.title %]">
                                        <img class="thumbnail" src="${new LRR.apiURL(`/api/archives/[% archive.arcid %]/thumbnail`)}" width="100"/>
                                    </a>
                                    <div class="caption" style="display: none;">
                                        <img style="height:300px" src="${new LRR.apiURL('/api/archives/${data.arcid}/thumbnail')}" 
                                                  onerror="this.src='${new LRR.apiURL('/img/noThumb.png')}'">
                                    </div>
                                </div>
```

lib/LANraragi/Utils/Minion.pm
```diff
@@ -140,6 +141,100 @@ sub add_tasks {
...
+            my @sections = split_workload_by_cpu( $numCpus, @ids );
+            eval {
+                $pl->foreach(
+                    \@sections,
+                    sub {
+                        my $redis = LANraragi::Model::Config->get_redis;
+
+                        foreach my $id (@$_) {
+                            # Skip if this ID has already been processed in another thread
+                            next if $visited{$id};
```
Difegue, line 184: Since you've used `split_workload_by_cpu`, normally each process/thread should have its own unique set of IDs.  

I'm not sure you need visited as a result?

templates/duplicates.html.tt2
```diff
@@ -0,0 +1,243 @@
...
+	<script src="[% c.url_for("/js/vendor/clsx.min.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/react-toastify.umd.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/jqcloud.min.js") %]"></script>
+	<script src="[% c.url_for("/js/vendor/allcollapsible.min.js") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/vendor/sweetalert2.min.js") %]" type="text/JAVASCRIPT"></script>
+
+	<script src="[% c.url_for("/js/common.js?$version") %]" type="text/JAVASCRIPT"></script>
+	<script src="[% c.url_for("/js/server.js?$version") %]" type="text/JAVASCRIPT"></script>
+    <script src="[% c.url_for("/js/duplicates.js?$version") %]" type="text/JAVASCRIPT"></script>
+    <style>
```
Difegue, line 38: Yeah I'm not a fan of having special CSS here, this will mess up custom themes.  
Ideally I think everything here is doable with the base lrr.css and re-using Index bits for the tags and thumbnail popups.  

It's fine to add a few extra classes to lrr.css if you'd need specific sizing.

psilabs-dev:
Thanks for working on this problem!

I notice there's one new post API in routing. Is this fire and forget/browserless? I'd like to call the job via curl/API client, let it run on a handful of cpus, then collect the results after a couple days.

In my case I'm hesitant to delete based on thumbnail similarity alone, but this feature would be a great starting point for further duplicate analysis based on custom user logic and code. E.g., merge tags among dupes, adding dupe sources to kept archives, prioritizing archives with most likes/bookmarks (if supported) or based on existing sort namespace, keeping different translations, setting do-not-downloads, etc.

This is also a comment long past implementation, so I understand if it's out of scope now.

Difegue:
The dupe scan job is just a minion task, so you can just fire and forget it, yeah.  
I'll wrap up the PR myself eventually if the original writer doesn't come back - that's what I get for having taken 3 months to review it...

Difegue:
Oh, thanks! Feel free to take your time, I'm in no rush.

psilabs-dev:
> The current approach is the simplest and most lightweight beside simple hash comparison, but I’m curious - how would you envision integrating _custom_ user logic and code? Would this be through additional filters and sorting options, or are you thinking of a deeper integration, some kind of plugin support?

It depends on the data source/plugin. For a site like Pixiv, the idea of "duplicate" doesn't *really* exist. Nhentai, on the other hand, has tons of true duplicates. Same manga, different translations; same translation, different translators; same series continuation, different to-be-continued chapters... Perhaps metadata plugins may be extended, e.g. nhentai-deduplication, e-hentai-deduplication, etc.

It would be difficult to incorporate all custom user logic, because different users have different definitions of what it means for two archives to be similar. It's better to keep two archives that may be duplicates than it is to delete a unique archive. So hands-off archive deletion of potentially tens of thousands of archives is not an easy ask.

The straighforward option is to offer an API interface, and let the users write their own duplicate detection in their language of choice. One API to trigger duplicate archive scanning, another to collect the results.

For example, trigger a scan on some endpoint:
```sh
curl -X POST http://localhost:3000/api/duplicates?cpus=4
# {"job_id": 123, "status": "success", "message": "Thumbnail duplicate scanning minion job queued!"}
```
then collect scan results:
```sh
curl -X GET http://localhost:3000/api/duplicates/123
```
return payload as list of similar archives by ID:
```json
{
    "status": "success/in-progress/failed",
    "data/duplicate archive IDs": [
        ["abcd", "efgh"],
        ["ijkl", "mnop", "qrst"]
    ]
}
```
I also wonder if we can extract the "image similarity" method so we can choose methods besides levenshtein distance in the future...

To give a practical example: I have a number of tagged archives downloaded via nhentai-archivist and cleaned with an internal service. LRR has RO access to contents downloaded by archivist so it cannot delete archives. This is what I'd be eventually doing (on a privileged service).

1. collect results of the duplicate scans
3. categorize archives within each cluster by "language:\*" tag
4. within a specific bucket, perform page-by-page image similarity checks: if an archive is a subset of another archive to a large extent (e.g. 80%-100% contained): the smaller archive can be deleted without much loss of information. Otherwise, they are not sufficiently similar. Then break these buckets down to page content-specific buckets
5. rank archives in these buckets by number of favorites, whether they have the "uncensored" tag or another indication of quality tag, whether they are in-progress, or whether they are in a static category. Basically find the "best quality" archive. Keep the archive with the highest ranking, queue deletion of the rest, and add some kind of "duplicate_source" tag for this kept archive pointing to the sources of deleted archives.
6. add the deleted archives to the nhentai-archivist do-not-delete catalog, then validate/execute deletion of archives

then add some basic logging, notifications, and run this once a month or something with a cron-job. Of course what I'm doing is subjective to my setup.

psilabs-dev:
Just an update on this; I decided to go the route of deduplication by treating archives as sequences of images, and solving the problem by finding all sequences which are proper subsequences of another sequence. In this case, equality between images is replaced with sufficient similarity between resnet embeddings. Equal-length duplicate sequences are handled by an nhentai-specific metadata comparison algorithm to determine which archive to keep.

The job is taking a super-long time (attempt to multiproc caused race conditions I was too lazy/dumb to fix :p), but around 16-17% of archives are projected to be duplicates (which after checking, indeed they are) which is good enough for me since that would shave off a considerable number of gbs from my disk, so I'm just letting it run while I go on with my day

PR #1109, opened 2024-11-07
Initial version of the GalleryDL Metadata plugin

okaros:
This PR implements a new metadata plugin, GalleryDL.  Its purpose is to import tagging metadata generated by the gallery-dl tool (https://github.com/mikf/gallery-dl) from json files stored either alongside the archive file (same name as the archive but ending in .json) or from an embedded info.json file.

I've based this plugin on the existing Eze plugin, as it's similar in concept. I opted to make it a separate plugin due to gallery-dl's significant variety of tagging formats and different metadata file structure; Hypothetically this plugin could be merged with the Eze plugin to create a single plugin that generically handles both, but I think that would just make both of them more confusing to use and the code/tests more difficult to manage.  Unlike the Eze plugin there are no parameters; My expectation is that gallery-dl users would handle any tag/title/etc... reformatting or management using gallery-dl's configuration rather than this plugin.

In theory this plugin should work with all the gallery-dl tagging metadata formats (there seem to be three variations), but it's possible I've missed some as there are ~200 different extractors in gallery-dl currently and I did not explicitly test every single one of them.  If needed, adding additional formats should be relatively easy.

I've included some test-cases for the different metadata variations; They pass cleanly on my dev system (and testing by hand/batch/automated import with this plugin in a running environment also works for me for all of the variations).


Difegue:
Looks great, I have no qualms merging this as-is. Agreed that Eze should probably remain separate as it's its own product. 

The hashtags format seems to store its tags with underscores rather than spaces (ie `purple_fire` instead of `purple fire`) -- The server supports both so it doesn't really matter, and I imagine this can be fixed with gallery-dl's preprocessor configuration anyway.  

Thanks for your work!

PR #1110, opened 2024-11-07
Test for file existance before removing mojo/minion/shinobu pid files

okaros:
Just a quick PR to remove a handful of common startup errors from the logs. This PR updates the pid file removals in 01-lrr-setup to only happen if the files actually exist.

Tested on my container-based deployment.


Difegue:
Sure, this works for me. Thanks!

PR #1113, opened 2024-11-08
Fixed Fakku summary (again) and added "thumb" folder to gitignore.

Nixis198:
Fakku been updating their website again and broke the summary part.
Looking at the HTML it seems that they put the summary in a div that shouldn't change so this fix should last for a bit.
I updated the test HTML to simulate the new format, so all tests pass for it.

I also added the "thumb" folder to the gitignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 content
+thumb
 .minion.db*
 npm-debug.log
 dump.rdb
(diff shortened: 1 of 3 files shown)


Difegue:
Oh they actually use the meta tag correctly now? Very nice yeah, this should hold for a bit. Thanks!

PR #1116, opened 2024-11-16
fixes inconsistencies in plugin parameters returned by the API

IceBreeze:
As pointed out by @Guerra24 in [this comment](https://github.com/Difegue/LANraragi/pull/1091#issuecomment-2480458807), there were some inconsistencies in the way plugin parameters were returned by the API `/api/plugins/:type`.

This should fix them.


Difegue:
Looks good, thanks. I missed that issue with the API, as mentioned in the old PR thread this isn't a very used call at the moment so it slipped my mind 🫠

PR #1117, opened 2024-11-17
Explicitly use ipv4 in Docker healthcheck

psilabs-dev:
When running the current Docker image, the healthcheck fails. This is because (by inspecting `/etc/hosts`), localhost is resolved to an ipv6 address. By default, Docker does not use/support ipv6. (https://docs.docker.com/engine/daemon/ipv6/, https://www.reddit.com/r/ipv6/comments/1alpzmb/docker_deployments_and_ipv6_how_do_you_do_it/) The solution is to explicitly use an ipv4 address rather than resolve it using localhost (it will take ~1min for the health status to be reflected).


TioaChan:
please revert this commit

CleanShot 2025-05-13 at 11 38 32@2x

PR #1125, opened 2024-12-06
Fix noisy logs

psilabs-dev:
Small fixes to code that cause lots of "use of uninitialized value" and redeclaration warnings. In particular, the "start" var is often undefined and throws a warning whenever search is called. Hopefully this commit will lead to less noisy logs.


Difegue:
looks good to me.

Issue #617, opened 2022-04-28
Add a i18n library to the frontend to allow translations


Difegue:
There's been demands for TL'd versions over the years, and while I'm not sure translators would step up to the plate to help maintain the translated strings, it'd be nice for the infrastructure to be in place. (And additionally to stop using hardcoded english everywhere)

Potentially quite a bit of work since both the JS and the html templates need to be internationalized -- Not sure what's the best course of action for this.  
i18next is the standard for this on the JS side at least:  
https://dev.to/adrai/the-progressive-guide-to-jquery-internationalization-i18n-using-i18next-3dc3

wc7086:
Any progress?

Difegue:
Unfortunately no.

PR #1131, opened 2024-12-23
Add i18n

Haibersut:
Having been preoccupied with my graduation project, I've been delayed for quite a while. As I revisited the project in the past few days, I realized that JavaScript's i18n isn't very effective for template replacements. Therefore, I discovered `Locale::Maketext` and made a few modifications based on examples I've found. Now, it automatically switches to the corresponding translation based on the browser's language setting.

Although `Locale::Maketext` isn't perfect and often presents peculiar encoding and escape character issues (the developer seemed to think that the `~` symbol wouldn't be commonly used, yet we do use it, XD), it's been a useful tool.

The translation and extraction of text from **the templates have now been completed**.

Next up is translating the JavaScript part. For JavaScript, I think sticking with i18n should suffice, though this will likely require a bit more time.

#617


Difegue:
This looks great at a glance - I need to look a bit closer at the initialization method to understand how it works, but I actually really like how this just uses the original english text as the keys -- It keeps the templates readable and makes it easy to update.  

Appreciate how it's just using .po files now as well.

Haibersut:
> This looks great at a glance - I need to look a bit closer at the initialization method to understand how it works, but I actually really like how this just uses the original english text as the keys -- It keeps the templates readable and makes it easy to update.
> 
> Appreciate how it's just using .po files now as well. Will likely merge soon, not seeing anything wrong with that approach 🤔

I see. So far, I've only extracted and translated text from the templates, but haven't yet addressed the text within the JavaScript.

Of course, you can proceed with merging. If you merge, would I be possible to include methods for contributing to translations in the documentation?

Difegue:
Yeah, it'd be fine to have the JS portion in a separate PR later; No problem keeping it all in this one if it makes things easiee for you though.

As for docs, I can pen something if needed yeah; Although it'd probably be better if you wrote it since you have better knowledge of how the localization system works than I do.

Haibersut:
> Yeah, it'd be fine to have the JS portion in a separate PR later; No problem keeping it all in this one if it makes things easiee for you though.
> 
> As for docs, I can pen something if needed yeah; Although it'd probably be better if you wrote it since you have better knowledge of how the localization system works than I do.

Got it. I think it's better to separate them into different PRs. Extracting and organizing text in JavaScript is really giving me a headache.

I've written a document on translating text within templates. Feel free to take a look.
[doc.md](https://github.com/user-attachments/files/18242823/doc.md)

Love-Megumin:
hero ! i really need this. ლ(＾ω＾ლ)

lib/LANraragi/Utils/I18NInitializer.pm
```diff
@@ -0,0 +1,58 @@
...
+                ($lang) = split /,/, $accept_language;
+                $lang =~ s/-.*//;
+                $c->log->debug("Detected language: $lang");
+            }
+
+            my $handle = LANraragi::Utils::I18N->get_handle($lang);
+
+            unless ($handle) {
+                $c->log->debug("No handle for language: $lang");
+                return $key;
```
Difegue, line 29: This means that if I'm using an accept_language header that's not in the lexicons, it'll return the localisation key as-is, right?  

Wouldn't this break the few keys we have that require to go through `maketext` for argument processing? (Like `"Version %1 %2"`)  

It might be better to just set the handle to `LANraragi::Utils::I18N->get_handle("en")` here and let the rest of the sub proceed as normal.

Difegue (changes requested):
One small potential bug I noticed.  

Apart from that I'm happy, this makes good use of Mojo's helpers and the result in templates is overall quite elegant. Really great work!  

The documentation looks good as well, I've integrated that into Gitbook.

lib/LANraragi/Utils/I18NInitializer.pm
Haibersut: This won't lead to an issue since `Locale::Maketext`, when it doesn't find a corresponding translation file, will automatically default to the English language file. It only returns the key if the corresponding key is not found in the currently matched language file.

However, I believe this is an area that can be improved. I've made some modifications: essentially, if a translator makes a typo in the key value or encounters other problems, it will automatically use the corresponding `msgstr` from the English file.

PR #1132, opened 2024-12-23
Fix the issue where "Pin this Category" doesn't work in Chrome

### LINKED ISSUE #1038: "Pin this Category" in wrong
**LRR Version and OS**
0.9.21 docker

**Bug Details**
https://github.com/Difegue/LANraragi/issues/1006#issuecomment-2198808208, Sorting for other languages ​​has been fixed，But "Pin this Category" is still wrong, and seems to have nothing to do with the language，I created a new docker to try and use English for all categories，The pin's category is not topped, only a red pin icon appears next to it

Reiyy:
This should fix #1038
Fix the issue where "Pin this Category" doesn't work in Chrome.

According to the documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort The sorting function should return a positive number, negative number, or 0. However, a boolean value is currently being used, which leads to unstable sorting results and unexpected behavior.

I’m not sure why everything works fine in Firefox, but this patch will fix the issue in Chrome and also return the correct results in Firefox.

修复在Chrome浏览器中“Pin this Category”不起作用

根据文档：https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort 排序函数需要返回正数、负数或0，而当前使用的是布尔值，这会导致排序结果不稳定而产生意外结果。

我尚不清楚为何在Firefox中一切正常，不过这个补丁可以修复在Chrome中的问题，同时在Firefox中也能返回正确结果。

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -663,7 +663,7 @@ Index.loadCategories = function () {
             // Sort by pinned + alpha
             // Pinned categories are shown at the beginning
             data.sort((b, a) => b.name.localeCompare(a.name));
-            data.sort((a, b) => a.pinned < b.pinned);
+            data.sort((a, b) => b.pinned - a.pinned);
             let html = "";
 
             const iteration = (data.length > 10 ? 10 : data.length);


Difegue:
Makes sense to me. Thanks for looking into this!

PR #1134, opened 2024-12-27
Copy locales to container images

Guerra24:
👀


Difegue (approved):
Real. Thanks for catching that.

PR #1137, opened 2024-12-29
feat: add DuplicateFinder plugin

chu-shen:
Find duplicate archives by title and add them to the DuplicateArchives category.

If you want to find duplicate archives, you need to manually trigger it. Check  the DuplicateArchives category later.
chrome_zPg5hfg8sV

---

Thanks to the standardized naming of doujinshi, different versions of doujinshi can be filtered by comparing the titles.

For titles that are different or have differences, they cannot be filtered out. The original design intention of this plugin is to filter out obvious duplicate archives as quickly as possible, rather than completely rejecting duplicates.

### DIFF
--- a/lib/LANraragi/Plugin/Scripts/DuplicateFinder.pm
+++ b/lib/LANraragi/Plugin/Scripts/DuplicateFinder.pm
@@ -1,0 +1,77 @@
+package LANraragi::Plugin::Scripts::DuplicateFinder;
+
+use strict;
+use warnings;
+no warnings 'uninitialized';
+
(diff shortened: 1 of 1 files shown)


Difegue:
Hmm, this duplicates a bit with https://github.com/Difegue/LANraragi/pull/1107 - I still need to review that one properly (and it'll need some upkeep now that we've implemented translations...) but I think it would solve your issue here.

Difegue:
Closing in favor of https://github.com/Difegue/LANraragi/pull/1224.

Issue #676, opened 2022-08-08
Resizeable columns in list view


soarpasser:
Usually some archives have pretty long titles, and the archive title column is not wide enough in list view to display all of them. Would be neat to be able to resize these columns.
![image](https://user-images.githubusercontent.com/74951691/183349232-cada0cf6-c628-4b65-bce0-5f670671fbad.png)

Difegue:
This seems [doable](https://stackoverflow.com/questions/69907376/columns-of-datatables-resizable-jquery), although one'd have to make sure the current proportion of column sizes is respected for first loads and when resizing the window.

Issue #669, opened 2022-07-28
[Feature Request] More custom columns for main page


hhz1213Bunches:
Please describe your suggestion, and the problem it'd solve.
I have checked the GH Discussions and it looks like no one talk about this suggestion. 
I think it would be better to have more columns to show in main page, so we can split some tags into different columns. (tags like ratings, doujin groups, publication date etc.)

<img width="1185" alt="archive_list" src="https://user-images.githubusercontent.com/63534543/181440895-381baccf-6b9e-4973-b8fb-03d7575833c1.png">

Difegue:
Simply adding more columns wouldn't be [that hard](https://github.com/Difegue/LANraragi/blob/258e500673f19ee4b8201f2f26270f7eb3a80279/public/js/index_datatables.js#L70), but I'm not sure people actually want 4 customizable columns at all time...  

It'd be better to make the columns fully customizable (add however many you want with a minimum of 1), but that is magnitudes harder. 🤔

PR #1144, opened 2024-12-30
Dynamic column

chu-shen:
Support dynamic columns and adjust column width

**default:**
image

**dynamic columns:**
image

**adjust column width**:
image
cursor:
image
After changing the column width, the page needs to be manually refreshed

#669 #676

### DIFF
--- a/locales/template/en.po
+++ b/locales/template/en.po
@@ -658,8 +658,8 @@ msgstr "Crop thumbnails"
 msgid "Go to Page:"
 msgstr "Go to Page:"
 
-msgid "Select Columns:"
-msgstr "Select Columns:"
+msgid "Columns:"
+msgstr "Columns:"
 
 msgid "Switch to Compact Mode"
 msgstr "Switch to Compact Mode"
--- a/locales/template/en.po
+++ b/locales/template/en.po
@@ -658,6 +658,9 @@ msgstr "Crop thumbnails"
 msgid "Go to Page:"
 msgstr "Go to Page:"
 
+msgid "Select Columns:"
+msgstr "Select Columns:"
+
 msgid "Switch to Compact Mode"
 msgstr "Switch to Compact Mode"
 
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -19,13 +19,8 @@ td.title {
   max-width: 400px;
 }
 
-#customheader1,
-td.customheader1 {
-  max-width: 200px;
-}
-
 [id^="customheader"],
-td[id^="customheader"] {
+td[class^="customheader"] {
   max-width: 100px;
 }
 
@@ -567,7 +562,7 @@ body.swal2-shown>[aria-hidden="true"] {
 
   #tagsheader,
   [id^="customheader"],
-  td[id^="customheader"],
+  td[class^="customheader"],
   td.tags {
     max-width: 50px;
   }
@@ -638,7 +633,7 @@ body.swal2-shown>[aria-hidden="true"] {
   #i2,
   [id^="customheader"],
   #tagsheader,
-  td[id^="customheader"],
+  td[class^=" customheader"],
   td.tags {
     display: none;
   }
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -20,12 +20,12 @@ td.title {
 }
 
 #customheader1,
-td.custom1 {
-  max-width: 150px;
+td.customheader1 {
+  max-width: 200px;
 }
 
-#customheader2,
-td.custom2 {
+[id^="customheader"],
+td[id^="customheader"] {
   max-width: 100px;
 }
 
@@ -565,11 +565,9 @@ body.swal2-shown>[aria-hidden="true"] {
     padding: 2px;
   }
 
-  #customheader1,
-  #customheader2,
   #tagsheader,
-  td.custom1,
-  td.custom2,
+  [id^="customheader"],
+  td[id^="customheader"],
   td.tags {
     max-width: 50px;
   }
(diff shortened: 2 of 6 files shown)


locales/template/en.po
```diff
@@ -658,6 +658,9 @@ msgstr "Crop thumbnails"
 msgid "Go to Page:"
 msgstr "Go to Page:"
 
+msgid "Select Columns:"
```
Difegue, line 661: I think it's fine for this to just be `Columns:` in English.

public/js/index.js
```diff
@@ -758,6 +799,64 @@ Index.migrateProgress = function () {
...
+        }
+    }
+
+    function stopResize() {
+        if (currentHeader) {
+            const index = Array.from(headers).indexOf(currentHeader);
+            localStorage.setItem(`resizeColumn${index}`, currentHeader.style.width);
+            currentHeader = null;
+        }
+        document.removeEventListener('mousemove', resizeColumn);
```
Difegue, line 854: I feel like this function needs a way to block the sort-on-click behavior that datatables has;  

Currently if I drag to resize a column and release the mouse while in the column, it'll trigger a sort, which is very disorienting and resets the column width.  

https://github.com/user-attachments/assets/f11583b7-e663-4e78-950d-5631c832d556

public/js/index.js
```diff
@@ -758,6 +799,64 @@ Index.migrateProgress = function () {
...
+                header.style.cursor = 'col-resize';
+            } else {
+                header.style.cursor = 'default';
+            }
+        });
+    });
+
+    function resizeColumn(event) {
+        if (currentHeader) {
+            const newWidth = startWidth + (event.clientX - startX);
```
Difegue, line 841: This calculation is very inaccurate to me some times - I have to drag the mouse pretty far before the column width starts reacting (as seen in the screen recording in the other comment)  

Some other times, it will violently snap the column width to a very different value as soon as I start sliding.

public/css/lrr.css
```diff
@@ -638,11 +636,9 @@ body.swal2-shown>[aria-hidden="true"] {
   }
 
   #i2,
-  #customheader1,
-  #customheader2,
+  [id^="customheader"],
```
Difegue, line 639: These changes seem to only collapse the headers, not the columns.  
image

templates/index.html.tt2
```diff
@@ -143,6 +143,14 @@
 				<div style="margin-left:auto">[% c.lh("Go to Page:") %]
 					<select class="favtag-btn" id="page-select">
 					</select>
+					<label for="columnCount">[% c.lh("Select Columns:") %]</label>
```
Difegue, line 146: I think this should only be visible in compact mode, it's confusing seeing it in the thumbnail view where it effectively does nothing.

public/js/index_datatables.js
```diff
@@ -46,6 +46,20 @@ IndexTable.initializeAll = function () {
...
+    let columnCount = localStorage.columnCount ? parseInt(localStorage.columnCount) : 2;
+    for (let i = 1; i <= columnCount; i++) {
+        columns.push({
+            data: "tags",
+            className: `customheader${i} itd`,
+            name: localStorage[`customColumn${i}`] || `defaultCol${i}`,
+            render: (data, type) => IndexTable.renderColumn(localStorage[`customColumn${i}`], type, data)
+        });
+    }
+    columns.unshift({ data: null, className: "title itd", name: "title", render: IndexTable.renderTitle });
```
Difegue, line 60: I'd just initialize `columns` with this line already in instead of unshifting it in after adding all the custom headers.

templates/index.html.tt2
Difegue: (I also wonder if a tooltip wouldn't be useful here -- This is technically only the number of custom columns and doesn't include title/tags, but it'd be way too long to write that in the main UI. Would also be a good place to note that changing this reloads the page)

Difegue (changes requested):
Wow, I never thought anyone would tackle this!  
I have some remarks on the dynamic column count feature but overall it's pretty good, it follows on the original logic which was already (almost) expandable.  

The resizable columns are a bit rough but hopefully it can be smoothed out. (Otherwise, I'd be happy just merging the dynamic columns first! We'll see)

chu-shen:
maybe fix:
- block the sort-on-click
- resizeColumn

https://github.com/user-attachments/assets/6daaccb3-f2c6-4386-851d-3878dfea9295

public/js/index.js
```diff
@@ -819,6 +819,7 @@ Index.resizableColumns = function () {
         // init
         header.addEventListener('mousedown', function (event) {
             if (event.offsetX > header.offsetWidth - 10) {
+                event.preventDefault();
```
Difegue, line 822: This change doesn't seem to fix anything for me on Firefox unfortunately.

public/js/index.js
```diff
@@ -856,6 +858,8 @@ Index.resizableColumns = function () {
         document.removeEventListener('mousemove', resizeColumn);
         document.removeEventListener('mouseup', stopResize);
         document.body.style.cursor = 'default';
+
+        document.location.reload(true);
```
Difegue, line 862: This makes the behavior worse it seems, the columns seemed to resize and remember their size fine without needing a page reload here.  

In fact it's worse now, since there appears to be some kind of race condition that prevents the column size from being saved in localStorage before the page reloads.

Difegue (changes requested):

chu-shen:
There seems to be no problem on Firefox v125.0 Portable

https://github.com/user-attachments/assets/78aa943c-12a3-4507-87cf-41c9c14ae5a3

image

Difegue:
Turns out using `offsetWidth` doesn't really work depending on DPI scaling on the host browser side -- I fixed that by using localStorage and `width` whenever possible.  

I also cleaned up some more logic to avoid the resizes weirdly jumping around and clamped the resizes by min/maxwidth.  

I took a sidetrip to see if we couldn't use a datatables plugin for the resizes, but the [one I found](https://github.com/dhobi/datatables.colResize) wasn't really better than what you'd already implemented. 🤝  

Thanks for the help!

PR #1152, opened 2025-01-03
Fix decode_json error when setting non-ASCII characters in plugin configuration

natanane:
Setting non-ASCII characters, such as Japanese, in the plugin configuration causes a decode_json error. To fix this, the data should be encoded before performing JSON decoding.


lib/LANraragi/Utils/Plugins.pm
```diff
@@ -131,6 +131,8 @@ sub get_plugin_parameters {
 
             #Decode it to an array for proper use
             if ($saved_config) {
+                # encode before json-decoding it in case it has UTF8 characters in plugin config
+                $saved_config = redis_encode($saved_config);
```
Difegue, line 135: Since we're doing a `redis_decode` 3 lines prior, you might as well remove it up there, so that `decode_json` uses the DB-encoded version of the string.

Difegue (changes requested):

Issue #1080, opened 2024-09-27
[API] Statistics API end point like those displayed on /stats


ForsakenRei:
Suggestion

For a homelab environment where poeple usually has a dashboard, maybe extend the current database statistics API end point to add something like total archives, tags etc. display on the https://yourinstance.com/stats?

Additional Context

The dashboard I'm using now is https://github.com/gethomepage/homepage which support custom API so I can display the statics there. I have a Python script to grab the numbers from the /stats page the expose my homemade API but I felt it's not the most elegant way to do it lol.

Difegue:
I presume you know of `api/database/stats` for the tag cloud data? This is what the "different tags existing" stat relies on.  
For the others, I've added `total_archives` to /api/info since it was straightforward; Content folder size is a bit more work to API-fy since it currently hard-formats the number to GBs and it's probably be better to return the raw number.  

Let me know if this enough for your needs.

ForsakenRei:
Thanks! And yeah I'm aware of the tag cloud data, though I didn't find the total count (or maybe I missed it somewhere?) so as lazy as I am I just use Python to scrap the statistics page lol. I guess if we already have a count for the statistics page it shouldn't be too complicated to add another end point...?

As for the total size, I didn't realized it's hard coded to GB so there is no need to rush, just when you have free time.

nonamethanks:
The correct way to do this would be to use prometheus metrics, which are the standard for monitoring applications nowadays. That way then it can be pulled into grafana/any other system out there.

ForsakenRei:
> The correct way to do this would be to use [prometheus metrics](https://eytanmanor.medium.com/an-introduction-to-prometheus-a-tool-for-collecting-metrics-and-monitoring-services-12fcc3bdb5d6), which are the standard for monitoring applications nowadays. That way then it can be pulled into grafana/any other system out there without worrying about supporting different custom apps.

Yes, my idea is we have some general API end points so people can decide what to do with them, I have prometheus and homepage supports custom API so there is no need for LANraragi to do extra development about dashboard support.

psilabs-dev:
I'll be playing around with prometheus for a bit.

psilabs-dev:
I've done some testing using the Net::Prometheus library, which is a common perl lib for implementing an exporter and have some thoughts. There’s also prometheus tiny, and an exporter specific to mojo which relies on net prom.

[...]

PR #1165, opened 2025-01-14
Python test environment

EfronC:
This is just a proof of concept for a possible test environment using Python, for a little more robust testing of the API endpoints. Feel free to close if you think this is out of scope for the platform.

**WHY?**
Current testing environment in Perl is OK for most scenarios, and probably is more than enough for the scope, but at the end of the day is very limited when it means to really interacting with the Redis DB, we have to mock up the actual Redis's commands,which at the end of the day means that tests might not end up being very reliable. With this tiny Python test environment, we actually mount the API, and Python simply requests the API, which allows it to execute the actual Redis commands to an actual Redis DB, and hence allows us to center on checking the actual result is what we should be expecting from an input data.

**Is it a better approach?**
TBH No, since this only allows us to test the API endpoints, and not the inner methods, TDD usually requires that every single piece of code is tested, and here we have that limitation, so this is more of a way to create superficial tests to determine that LRR is still working as expected after a whole set of work, than to verify that the platform as a whole is stable.

**How does it work?**
I added a new docker compose file, and a new Dockerfile, those are the same than the base platform, but they add a Python instance with pytest and a few other libraries. Inside the tests folder there is a Python folder with all the python files for the testing(And a requirements.txt in the root). You call pytest on the Python instance to run the tests and verify that everything is passing. I usually use these 2 commands on a Makefile to execute them:

```
build-test:
	docker compose -f docker-compose-testing.yml build
test:
	-docker compose -f docker-compose-testing.yml run --rm python-tests python -m pytest
	docker compose -f docker-compose-testing.yml down --volumes --remove-orphans
```

I'm only providing one single test for this concept, since working on everything is a little lot of work, and I'm not sure if this might be a good feature, or just something to be rejected, especially because testing the log in required endpoint will probably need a little more of work to find a way to generate an API key on start up. I'm mostly creating this PR to showcase and know if this is something worth working on, or if should be discarded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ config.log
 autobackup.json
 Makefile
 oshino
+
+__pycache__/
+.pytest_cache/
+*.py[cod]
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+pytest
+pytest-redis
+requests
+factory_boy
--- a/tests/python/factories/archivefactory.py
+++ b/tests/python/factories/archivefactory.py
@@ -1,0 +1,53 @@
+import factory
+import os
+import zipfile
+
+from factories.models.archives import Archive
+
+def generate_file(obj):
+    logo_path = "/lanraragi/public/img/logo.png"
+    content_path = "/lanraragi/content/test"
+    
+    if not os.path.exists(content_path):
+        os.makedirs(content_path)
+
+    zip_path = os.path.join(content_path, f"{obj.name}.zip")
+
+    with zipfile.ZipFile(zip_path, 'w', zipfile.ZIP_DEFLATED) as zipf:
+        file_name = os.path.basename(logo_path)
+        zipf.write(logo_path, file_name)
+
+    return "/home/koyomi"+zip_path
+
+class ArchiveFactory(factory.Factory):
+    class Meta:
+        model = Archive
+
+    name = factory.Faker("first_name")
+    tags = "date_added:1736124197"
+    summary = ""
+    arcsize = "16532135"
+    file = factory.LazyAttribute(lambda p: '/home/koyomi/lanraragi/content/test/{}.zip'.format(p.name))
+    title = factory.LazyAttribute(lambda p: p.name)
+    isnew = "true"
+    pagecount = "30"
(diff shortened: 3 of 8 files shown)


psilabs-dev:
Regarding how to inject API keys to LRR, I have a method you can reference [here](https://github.com/psilabs-dev/lanraragi-satellite/blob/a91a003f891bde0a772a13210b9d2589e2ac9e6b/integration/setup.sh#L10).

Integration/e2e tests in Python are something I'm also working on through https://github.com/psilabs-dev/lanraragi-satellite, but right now it's not mature. It provides an async API [client](https://github.com/psilabs-dev/lanraragi-satellite?tab=readme-ov-file#lanraragi-python-sdk) and mock archive generation tools ([manycbz](https://github.com/psilabs-dev/lanraragi-satellite?tab=readme-ov-file#manycbz)), which are used mostly by my microservice but also for future testing.

The issue with testing a data-oriented web server is reproducible data generation at scale. Some problems only express themselves when a large number of archives are inserted, the same goes for making many simultaneous requests to trigger race conditions and locking issues. To find these bugs we need the infrastructure to test them, hence the need for an async client and mock data generator.

EfronC:
Thanks, I just added a simple example using an api key for a test to an auth required endpoint, for a more complete showcase.

Difegue:
Thanks for the work put into this, I appreciate the effort.  
I do think a separate Python test environment with its own set of Dockerfiles/libraries/samples is a bit out of scope to put in the main repo...   
Having that in a separate project/microservice like Satellite makes more sense to me.  

I'm definitely interested in integration tests and further instrumentation of the server, as I think it'd be key to fixing some of the speed/processes issues we have at scale/load as mentioned.   

As far as the core server can help, implementing better support for metrics that can be read by external services as in https://github.com/Difegue/LANraragi/issues/1080 would probably be helpful?  
Stuff like CPU time per request and current amount of processes/workers, etc.

EfronC:
Ok, thanks for the answer.

Yeah, that's why I created this showcase, to know if it was worth to work on before doing so. I'll close the PR, and leave the space open for another implementation of this same idea 👍

Issue #1129, opened 2024-12-22
UX feature request: Like Button


psilabs-dev:
Add an option to have a static category with the "Like" (or "Favorites") property, and express this property as a toggleable like icon on archive thumbnails. Users can assign any single category with this property, and of course assign any meaning to "liking" an archive.

In many social media platforms, this functionality is supported but is named differently; for example in Pixiv, it's called "Bookmark", in nhentai it's "Favorite", in Twitter/Facebook it's "Like". In LRR, User can change which category is their "Like" category.

The following are two examples of the like button on Pixiv and Twitter:
<img width="1274" alt="Screenshot 2024-12-21 at 7 23 09 PM" src="https://github.com/user-attachments/assets/375da183-db0b-46d4-9e92-6d1293224851" />
<img width="591" alt="Screenshot 2024-12-21 at 7 54 21 PM" src="https://github.com/user-attachments/assets/a78e8609-c328-48ca-b07f-99f957d01c1f" />

If a "Like" category is configured, the UI will add a "like" button on every archive thumbnail which toggles the archive's membership in the category. Additionally, when reading an Archive, the "like" button will also be present. The "like" button serves two purposes: 1) to quickly show whether or not the archive is in the "Like" category, 2) to conveniently toggle the archive's membership in said category.

### Why have this feature? 

Often I find that I want to quickly add images/archives in my server to my Favorites. The fastest way to do this is on a laptop, even then I need to right click, then add to categories, then select my Favorites category. In a situation where you have tens of thousands of archives, this can become very time consuming. It is also nice to easily notice archives that you've liked.

On a tablet like ipad/phone, it's much slower: there is no right click functionality that supports putting an archive to category easily, I have to go to the archive metadata page and add a category there. 

While I can make frontend customizations to my LRR instances (e.g. add a button that automatically toggles my archive's membership for a *specific* category ID, I would like to look for more general ways to approach this.

###  One way to implement this
Add a hash key LRR_LIKE_CATEGORY_ID and point it to null or a static category ID. Implement corresponding CRUD APIs for this value.

When rendering archives in the welcome page or while reading an archive, check if each archive is a member of the category fetched by LRR_LIKE_CATEGORY_ID if exists, and display the appropriate icon button; otherwise, skip the button rendering stage.

And a (poorly) drawn example for LRR.
[...]

PR #1173, opened 2025-01-25
Implement bookmark button

psilabs-dev:
Issue #1129

Just gonna put an AI explanation, since it's more clear than whatever I had

---

The bookmark link feature allows users to quickly add/remove archives to a specific static category using a convenient bookmark icon. This is essentially a one-click way to manage your favorite or frequently accessed content.

1. **Bookmark Icon**: When enabled, a bookmark icon appears on archive thumbnails in the index page and in the reader interface. Clicking this icon toggles whether the archive belongs to the designated category.

2. **Linked Category**: The bookmark icon is linked to a single static category in your collection. By default, a "Favorites" static category is created for new installations and linked to the bookmark feature.

3. **Static Categories Only**: Only static categories (those that contain a manually curated list of archives) can be linked to the bookmark feature. Dynamic categories (those based on search predicates) cannot be used with this feature.

4. **Configuration**: You can change which static category is linked to the bookmark button through the Categories management page. Select a static category and enable the "Link Bookmark Icon to this Category" toggle.

5. **Disabling**: You can disable the bookmark feature entirely by toggling off the "Link Bookmark Icon to this Category" option. The feature is also automatically disabled if you delete the category it was linked to.

6. **User Authentication**: The bookmark functionality requires user authentication - only logged-in users can toggle bookmarks.

7. **State Persistence**: The bookmark state (whether an archive is in the linked category) is visually shown with a filled bookmark icon (added) or an outline bookmark icon (not added).


lib/LANraragi.pm
```diff
@@ -184,6 +185,17 @@ sub startup {
     shutdown_from_pid( get_temp . "/shinobu.pid" );
     start_shinobu($self);
 
+    # First-install actions: check if redis LRR_CONFIG is updated.
```
Difegue, line 188: I think this should go in a dedicated sub in Model::Config IMO

lib/LANraragi/Model/Category.pm
```diff
@@ -169,6 +170,10 @@ sub delete_category {
     }
 
     if ( $redis->exists($cat_id) ) {
+        if ( $redis->hget('LRR_CONFIG', 'highlight') eq $cat_id ) {
+            $logger->info("$cat_id is a highlight, removing.");
```
Difegue, line 174:
```suggestion
            $logger->info("$cat_id is the highlighted category, removing it.");
```

lib/LANraragi.pm
```diff
@@ -184,6 +185,17 @@ sub startup {
...
+    unless ( $self->LRR_CONF->get_redis_config->hexists('LRR_CONFIG', 'htmltitle') ) {
```
Difegue, line 189: This will consider any update a first-time installation, except for users who changed their server title, right? I suppose that's fine by me.

lib/LANraragi/Controller/Api/Category.pm
```diff
@@ -138,5 +139,61 @@ sub remove_from_category {
     }
 }
 
+sub get_highlight_category {
```
Difegue, line 142:
```suggestion
sub get_highlighted_category {
```  
I think we should update all the names to refer to a "highlighted category", instead of a highlight. It's a little thing, but it makes it more explicit that is is a specific flag on top of a given category ID.

lib/LANraragi/Utils/Routing.pm
```diff
@@ -148,9 +148,12 @@ sub apply_routes {
 
     # Category API
     $public_api->get('/api/categories')->to('api-category#get_category_list');
+    $public_api->get('/api/categories/highlight')->to('api-category#get_highlight_category');
```
Difegue, line 151: Endpoints look good too, I'd just use highlighted here too.
```suggestion
    $public_api->get('/api/categories/highlighted')->to('api-category#get_highlighted_category');
```

Difegue (changes requested):
This looks overall good, sweet and simple. I don't have much to complain about.  

I'm a bit iffy on the "highlighted" term, but I don't really have a better one coming to mind at the moment.

Difegue:
Regarding UI, I'd just add a checkmark below "pin this category" that makes the selected cat the highlighted one. Ofc, you can only do this with static categories.  

In the index, i think you can add the "like" button to the javascript-created thumbnails fairly easily, that's in a dedicated function. 
For compact mode, I would add a button inline to the left of the title, like how we used to do for the download/edit metadata buttons before the context menu came around. (that was a while ago...)  

You need however to know the "initial" state of the like button when the index loads (aka whether an ID is in the highlighted category or not).  
We could either do this purely with API calls and JS, or do some search engine work to easily surface that a given archive belongs or not to the highlighted category. (Get the ID list of the highlighted category, then when iterating over the final search results, check whether or not your IDs are in there)

lib/LANraragi.pm
psilabs-dev: Yes; in fact any user who has ever updated their settings will be exempt from the first-time installation update, as a settings update will add the server title to redis.

psilabs-dev:
Yeah I'm not too sure about the naming either, we can sit on that for a while, see if anyone has thoughts, or ask an AI for ideas. I wouldn't mind if we just gave it a Japanese phrase either.

I'm sure the front end work isn't difficult, I just need to spend some good time learning the basic principles and HTML/CSS/JS structure/DSL and speaking the frontend language. Right now I don't know what I don't know or what I should know, or if a term is used in a informal or technical setting.

lib/LANraragi.pm
psilabs-dev: Will do

lib/LANraragi.pm
Difegue: Makes sense, that's even better 👍

Difegue:
Fair enough - Maybe someone will come up with a better name by the time we wrap up the UI!  
I don't know if there's an equivalent japanese term in Pixiv or such people would be familar with.

psilabs-dev:
Ok I have a first draft down. I'm going to test this on my server for a bit and clean up the code (also I really want to start using it lol).

A lot of sync-related bugs were squashed, but I'm sure there are still more.. I'll see how it works in prod-like environments. Also added some python API [integration tests](https://github.com/psilabs-dev/lanraragi-satellite/blob/a731de1db93119f3ce48dc4cdaa54adf0b7df8bb/integration-tests/test_simple.py#L121) on my side. Didn't do much code standardization, I figure more changes will be made.

I've decided to go with the name "bookmark" for the frontend, and "bookmark link" for the backend. The word highlight is too confusing, and this backend functionality is a conceptually abstract; most websites don't allow you to flexibly change where the bookmark points to. It's like youtube letting you choose which playlist the like button saves videos to. A readonly user would never see the "bookmark link" terminology; they would only interact with the bookmark button.

(We could also go for "like", "star", "love", "save", and replace "link" with "reference" or "connect". I realized midway the repo uses fontawesome which supplies icons, so we don't need to cook up in-house icons. However, "save" and "star" are overloaded terms, and "like/love" generally implies the notion of a "dislike")

Now, for functionality: the bookmark icon/button is implemented in three locations mainly: categories, reader, and index, with helpers in LRR and server. Generally, the bookmark category ID will be fetched on initialization and put in local storage.

Categories page: Enables changing which category the bookmark toggle points to, and whether the bookmark toggle should exist.

Reader: adds a bookmark icon to the `absolute-left` divs. This icon appears only when the bookmark is configured. The icon toggles the membership of an archive to that category, and is synced with the archive metadata settings, i.e. if you unbookmark the archive via the icon, it will remove the bookmark category from the metadata page, and vice versa.

Index page: adds a bookmark icon to the bottom left of each image thumbnail. For every archive, their icons are synced; e.g., when the same archive appears in carousel and in datatables. Also, these icons are synced with the context menu categories section. If you unbookmark/bookmark an archive through any channel, all other indications will be updated. Also, this bookmark icon will show only when the bookmark is configured.

Index page (compact): prepends a checkbox col before the title col. Likewise this checkbox column is synced with all other channels. When the bookmark is not configured, this column will display off and be disabled from toggling. Maybe we'd want to disable this column instead, but I haven't gotten that to work. I still don't really understand the datatables-related code.

templates/index.html.tt2
```diff
@@ -48,7 +48,7 @@
 
 </head>
 
-<body>
+<body data-user-logged="[% userlogged %]">
```
psilabs-dev, line 51: Since userlogged is a server-side variable, it's not exposed to the client so the client doesn't really know whether the user is logged in.

Toggleability of bookmark icon should be enabled only if the user is logged in (i.e. user has "write" permissions). In readonly mode (i.e. user viewing site with no-fun-mode disabled), bookmark icons and checkboxes should not be toggleable. This decision requires exposing the logged-in status to the client somewhat, hence the use of html data attributes here

templates/footer.html.tt2
```diff
@@ -6,4 +6,11 @@
     <br>
     [% END %]
     Powered by <a href="https://github.com/Difegue/LANraragi"> LANraragi.</a>
-</p>
\ No newline at end of file
+</p>
+
+<!-- i18n client-side access -->
```
psilabs-dev, line 11: Decided to put i18n translations that are dynamically required by clients in this central location and made accessible by common.js, to avoid making script initializations in every tt2 file that corresponds to a js file that dynamically generates elts with translatable text

locales/template/zh.po
```diff
@@ -1028,6 +1031,8 @@ msgstr "档案概览"
 msgid "FullScreen"
 msgstr "全屏"
 
+msgid "Toggle Bookmark"
+msgstr "切换书签"
 # ------End of Reader.html.tt2------
```
psilabs-dev, line 1036: Kept the Chinese translations minimal, but alternative suggestions are welcome.

templates/footer.html.tt2
Difegue: > avoid making script initializations in every tt2 file that corresponds to a js file that dynamically generates elts with translatable text

How did you figure out that was what I was working on for frontend translations? 🫠  
I've pushed my WIP here for reference -- https://github.com/Difegue/LANraragi/commit/2eefd23a58c0c3f18b1b2610e0463e8534a1b6a0  

I'm not sure if I prefer putting it all in the footer -- It feels less clear when working on the codebase that this is where the glue for frontend TLs would end up.   

Also, this requires double-declaring every string in common.js, which is going to add a lot of heft to that file.

Difegue: (I'm also still undecided on how to work with placeholders in localized text.. The backend didn't have this issue for obvious reason, but I need to decide on a format for string interpolation here)

psilabs-dev: Ah I didn't know how the i18n system worked initially, when I found it was generated by the server I figured this problem would come up sooner or later 😅

Yeah I wasn't too excited about this approach either, If you're working on this part then I'll just use yours when it's implemented, then I can do away with this awkward workaround

public/js/index.js
```diff
@@ -119,20 +125,83 @@ Index.initializeAll = function () {
...
+        Server.removeArchiveFromCategory(id, localStorage.getItem("bookmarkCategoryId"));
+        Index.bookmarkIconOff(id);
+    }
+}
+
+Index.toggleBookmarkStatusByIcon = function (e) {
+    const icon = e.currentTarget;
+    const id = icon.id;
+
+    // TODO: Considering removing this toast
```
psilabs-dev, line 185: These types of toasts might be too noisy, so I wouldn't mind if we remove them, or replace with something else

psilabs-dev:
okiee putting this up for review since no further code changes are expected (surely I will not find a critical bug right after marking this as review, **right**??). Some smaller opinion/design based issues (e.g. naming) still are up for debate in comments. Dogfed this for the past couple months on production LRR servers totalling 200-300k archives, ofc the API is so restrictive I never expected anything bad to happen.

Difegue:
Thanks - It's a big one but I'll try to get through it in a reasonable time.

public/js/index_datatables.js
```diff
@@ -158,6 +167,28 @@ IndexTable.renderColumn = function (namespace, type, data) {
...
+        const isBookmarked = JSON.parse(localStorage.getItem("bookmarkedArchives") || "[]").includes(data.arcid);
+        if (type === "display") {
+            const disabledAttr = LRR.isUserLogged() ? '' : 'disabled';
+            const disabledStyle = LRR.isUserLogged() ? '' : 'style="opacity: 0.5; cursor: not-allowed;"';
+            return `<input class="bookmark-checkbox" id="${data.arcid}" type="checkbox" title="Toggle Bookmark" ${isBookmarked ? 'checked' : ''} ${disabledAttr} ${disabledStyle}>`;
+        }
+        return isBookmarked;
+    }
+    if (type === "display") {
+        return `<input class="bookmark-checkbox" id="${data.arcid}" type="checkbox" disabled title="Link bookmark to category to use the checkbox.">`;
```
Difegue, line 187: Missing a loc key if we end up keeping this

public/js/index_datatables.js
```diff
@@ -394,3 +420,11 @@ IndexTable.buildTagTooltip = function (target) {
 
     $(target).attr("onmouseover", "");
 };
+
+/**
+ * When bookmark is configured, the title column is the second column, otherwise it's the first column
+ * @returns index of title column in compact mode
+ */
+IndexTable.getTitleIndex = function () {
```
Difegue, line 428: Not a fan of this -- It makes the table clunkier to read as the bookmark column takes a lot of space just for a checkbox. 
image  

I also think it breaks the dynamic column resizing we've implemented recently..

I'd either move bookmarking to the context menu for table mode, or add it as an overlay icon in the title column like the "New" indicator.

lib/LANraragi/Model/Config.pm
```diff
@@ -168,6 +171,28 @@ sub get_tagrules {
...
+# button. Returns 1 if is first-time installation, else 0.
+sub first_install_actions {
+    my $redis = get_redis_config();
+    my $logger = get_logger( "Config", "lanraragi" );
+    unless ( $redis->hexists('LRR_CONFIG', 'htmltitle') ) {
+        $logger->info("First-time installation detected!");
+        $redis->hset('LRR_CONFIG', 'htmltitle', 'LANraragi');
+
+        $logger->debug("Creating first category...");
+        my $default_category_id = LANraragi::Model::Category::create_category("Favorites", "", 0, "");
```
Difegue, line 186:
```suggestion
        my $default_category_id = LANraragi::Model::Category::create_category("🔖 Favorites", "", 0, "");
```  

Let's add an emoji here so it fits with the other default selectors. 
image
Hopefully it makes it intuitive for new users that the bookmark function is linked to this category by default.

public/css/lrr.css
```diff
@@ -771,6 +782,25 @@ div.fullscreen img {
   overflow-y: scroll;
 }
 
+.thumbnail-bookmark-icon {
+  cursor: pointer;
+  position: absolute;
+  left: 10px;
```
Difegue, line 788: I don't think the default placement of the icon should intersect thumbnails -- Maybe we should give it a bit more padding? I'd prefer having it on the right side instead of the left one too but that could just be personal preference.  

image

```suggestion
  right: 24px;
```  

image

public/css/lrr.css
```diff
@@ -771,6 +782,25 @@ div.fullscreen img {
...
+  bottom: 20px;
+  z-index: 2;
+  font-size: 24px;
+  color: white;
```
Difegue, line 792:
```suggestion
```
I'd remove this attribute outright so the bookmark button uses the text color for the current theme instead.  

image
image
image

(Using white has a bit of a readability issue against B/W manga pages, but the text shadow is enough to handle that IMO)

public/js/category.js
```diff
@@ -139,11 +148,44 @@ Category.saveCurrentCategoryDetails = function () {
     Category.indicateSaving();
 
     // PUT update with name and search (search is empty if this is a static category)
+    // Indicate saved and load categories are placed inside the API call to avoid race conditions.
     Server.callAPI(`/api/categories/${categoryID}?name=${catName}&search=${searchtag}&pinned=${pinned}`, "PUT", null, "Error updating category:",
         (data) => {
-            // Reload categories and select the newly created ID
-            Category.indicateSaved();
-            Category.loadCategories(data.category_id);
+            if (searchtag === "") {
```
Difegue, line 154: I'm not sure it makes sense to do the bookmark_link API calls after having done a base save (which effectively does nothing if all the user did was toggle the checkbox.)  

Wouldn't it be better to move this entire block into a dedicated `Category.updateBookmarkLink` method and attach that to `change.bookmark-link`?

public/js/common.js
```diff
@@ -260,6 +277,14 @@ LRR.buildThumbnailDiv = function (data, tagTooltip = true) {
     const id = data.arcid || data.id;
     let reader_url = new LRR.apiURL(`/reader?id=${id}`);
 
+    let bookmarkIcon = ``;
+    if ( LRR.bookmarkLinkConfigured() ) {
+        const isBookmarked = JSON.parse(localStorage.getItem("bookmarkedArchives") || "[]").includes(id);
+        const bookmarkClass = isBookmarked ? "fas fa-bookmark" : "far fa-bookmark";
+        const disabledClass = LRR.isUserLogged() ? "" : " disabled";
+        bookmarkIcon = `<i id="${id}" class="${bookmarkClass} thumbnail-bookmark-icon${disabledClass}" title="Toggle Bookmark" ${!LRR.isUserLogged() ? 'style="opacity: 0.5; cursor: not-allowed;"' : ''}></i>`;
```
Difegue, line 285: I'd put the style as a separate const as well since you did it for disabledClass.

public/js/index_datatables.js
Difegue: image
```
<td class="title itd">
    <i id="{...}" class="fa-bookmark thumbnail-bookmark-icon-inline fas" title="Toggle Bookmark"></i>
    <div class="isnew"><sup>2/10</sup></div> 
                <a class="context-menu" [...]
```  

```
.thumbnail-bookmark-icon-inline {
  cursor: pointer;
  font-size: 12px;
  text-shadow: 0 0 3px rgba(0, 50, 116, 0.9);
}
```

public/js/reader.js
```diff
@@ -144,8 +152,30 @@ Reader.initializeAll = function () {
...
     );
+
+    // Fetch "bookmark" category ID and setup icon
+    Reader.loadBookmarkStatus();
 };
 
+/**
+ * Adds a removable category flag to the categories section within archive overview.
+ */
+Reader.addCategoryFlag = function ( categoryId ) {
```
Difegue, line 163: "Flag" made little sense to me when first reading the changes and then I understood it was just refactoring the existing subitems that go in archive overview.  

Can we rename this to  
```suggestion
Reader.addCategoryBadge = function ( categoryId ) {
```

public/js/reader.js
```diff
@@ -144,8 +152,30 @@ Reader.initializeAll = function () {
...
+    const html = `<div class="gt" style="font-size:14px; padding:4px">
+        <a href="${url}">
+        <span class="label">${categoryName}</span>
+        <a href="#" class="remove-category" data-id="${categoryId}"
+            style="margin-left:4px; margin-right:2px">×</a>
+    </a>`;
+    $("#archive-categories").append(html);
+}
+
+Reader.removeCategoryFlag = function ( categoryId ) {
```
Difegue, line 175: Same for this one.

public/js/reader.js
```diff
@@ -435,6 +472,66 @@ Reader.toggleHelp = function () {
     // all toggable panes need to return false to avoid scrolling to top
 };
 
+Reader.toggleBookmark = function(e) {
+    e.preventDefault();
+    if ( !localStorage.getItem("bookmarkCategoryId") ) {
+        return;
```
Difegue, line 478: This should probably at least log something to the console? Even though it should normally never happen since we do the api call on load.

public/js/reader.js
```diff
@@ -435,6 +472,66 @@ Reader.toggleHelp = function () {
...
+    };
+
+    if (!LRR.isUserLogged()) {
+        LRR.toast({
+            text: I18N.LoginRequired(new LRR.apiURL("/login")),
```
Difegue, line 483:
```suggestion
            heading: I18N.LoginRequired(new LRR.apiURL("/login")),
```

public/js/index.js
```diff
@@ -119,20 +121,81 @@ Index.initializeAll = function () {
...
+    }
+}
+
+Index.toggleBookmarkStatusByIcon = function (e) {
+    const icon = e.currentTarget;
+    const id = icon.id;
+
+    if (!LRR.isUserLogged()) {
+        LRR.toast({
+            text: I18N.LoginRequired(new LRR.apiURL("/login")),
```
Difegue, line 183:
```suggestion
            heading: I18N.LoginRequired(new LRR.apiURL("/login")),
```

public/js/reader.js
Difegue: Although if the api call failed, we won't even show the bookmark icons... nvm

Difegue (changes requested):
This looks good overall - I think the changes made to compact mode/datatables to add an extra column end up bringing a lot of unneeded complexity we could do without.  

Thanks for being so exhaustive with the doc and screenshots!

public/js/index.js
Difegue: This one went away, but I'm ok with the add/remove category toasts when interacting with bookmarks as they stand - it's not massively intrusive.

locales/template/en.po
```diff
@@ -172,6 +172,18 @@ msgstr "If you select a Static Category, your archives will appear here so you c
 msgid "Return to Library"
 msgstr "Return to Library"
 
+msgid "Link Bookmark Icon to this Category"
```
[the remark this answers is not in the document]
Difegue: I'm personally happy with "bookmark" as the overall term for the feature and bookmark-link for the API BTW.  

Hopefully people won't conflate it with bookmarking a specific page in a given archive?

public/js/index_datatables.js
psilabs-dev: > I also think it breaks the dynamic column resizing we've implemented recently..

Oh oiops, that is a bug I can fix. Granted I don't use compact mode all that often, so it slipped my checks.

public/js/index.js
psilabs-dev: Yeah I think the average user probably isn't spamming the bookmark icon for a bunch of toasts showing up to be an issue.

public/js/index_datatables.js
psilabs-dev: > I'd either move bookmarking to the context menu for table mode

Putting it in the context menu would somewhat defeat the purpose if the feature I think, since if you open the context menu to toggle a bookmark, you might as well add to category instead; the savings don't really justify it

public/js/reader.js
psilabs-dev: I added a log here anyways, though gta6 would probably come out before this log is triggered

public/js/category.js
psilabs-dev: Ah this was a while back, I was just mimicking the the methods above me. Fixed

templates/index.html.tt2
psilabs-dev: Yeah we would be just be passing abstractions around.

It might help if we add a role API `getRole`, to show what permissions we have? Right now we only have 2 roles ("read all archives" and "admin") so it's really basic, but if in the future we entertain the idea of having more roles (e.g. multi-user management), this might be worth looking into.

public/js/index_datatables.js
psilabs-dev: How about this? The bookmark icon will display to left of title when bookmark is configured, and reflect the archive's bookmarked status. Behavior will be the same as that of the thumbnail.

demo

Difegue (changes requested):
one small nit left but it's all looking good now.

public/js/index_datatables.js
Difegue: Yup, that's exactly what I wanted. Nice job!

public/css/lrr.css
```diff
@@ -771,6 +782,35 @@ div.fullscreen img {
...
+  position: relative;
+  right: auto;
+  bottom: auto;
+  font-size: inherit;
+  text-shadow: none;
+  vertical-align: middle;
+  margin-right: 5px;
+}
+
+.thumbnail-bookmark-icon:hover {
```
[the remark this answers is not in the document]
psilabs-dev: done

psilabs-dev:
> Thanks for being so exhaustive with the doc and screenshots!

Well I am the one pushing for this feature, so I better make sure it's good. 😄

Appreciate the speedy review!

Difegue:
looks all good, I think we can merge this!

I might make a release relatively™️ soon with this + the CHI changes and the dupefinder if all goes well.

Difegue (approved):

PR #1189, opened 2025-02-16
Fix incorrect sort display

### LINKED ISSUE #1183: Index.handleCustomSort doesn't render the correct sortby namespace
OS: docker on ubuntu 22.04

The latest dev builds of 0.9.40 have introduced a regression on UI. Changing the sorting namespace does not change the displayed namespace.

Image

The namespace will always be displayed as "title". This problem is traced to `Index.handleCustomSort`. Executing `IndexTable.dataTable.draw();` renders the namespace correctly in a stable build of 0.9.30 (3a78ac3f6d04a33a6c2e531213fe67d43401e051), but doesn't in 0.9.40. I'm suspecting a change in the datatables caused this, though I don't know what.

Update: narrowed the issue down to commit 0a22e12d1a8678002d4488fa899814ad31bca0ab.

psilabs-dev:
Closes #1183 

`columnCount` isn't a number, it's some html object and cannot be properly compared with a number. To get the value of column count you would need to access the value attribute.

Making the change results in a correct comparison, but please double-check if the logic is correct

### DIFF
--- a/public/js/index_datatables.js
+++ b/public/js/index_datatables.js
@@ -283,7 +283,7 @@ IndexTable.drawCallback = function () {
         if (currentSort > currentCustomColumnCount) {
             localStorage.indexSort = 0;
         }
-        if (currentSort >= 1 && currentSort <= columnCount) {
+        if (currentSort >= 1 && currentSort <= columnCount.value) {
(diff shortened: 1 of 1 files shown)


Difegue:
lgtm

PR #1195, opened 2025-02-20
docs(search-api): fix search api tank parameter name is not correct

Reiyy:
The name defined in the code is `groupby_tanks`, but the name in the documentation is `group_tanks`.

In `lib\LANraragi\Controller\Api\Search.pm` :
`my $grouptanks = $req->param('groupby_tanks') || "false";`


Difegue:
Correct, thanks!

Issue #1140, opened 2024-12-29
[Pixiv] Key constraint for metadata plugin on "date_uploaded" and "artist"


psilabs-dev:
It came to mind that the metadata plugin for pixiv doesn't enforce uniqueness on certain tags, which can result in data integrity issues. In particular, the values for "date_uploaded" and "artist" should be unique when called by the pixiv metadata plugin on an artwork ID.

Case for "date_uploaded": If an artist privates their artwork and unprivates it, the date of upload will change to the more recent unprivated date. The metadata plugin should overwrite the date_uploaded field so that any archive should only reflect the most recent uploaded date.

Case for "artist": Pixiv archives always have exactly one artist. If the Pixiv artist changes their name, the name of the artist corresponding to the artwork should be changed to the new name, rather than having both new and old names.

Other tags like "pixiv_user_id" or "source" or "date_created" are assumed to never change, so they're OK.

This is probably a theoretical bug I don't think many people have run into; it requires you to call the plugin on an already-tagged archive in which the source upload date or artist name have changed, but I should fix this sometime in the future.

psilabs-dev:
resolved with API jobs

PR #1197, opened 2025-02-24
Fetch engagement metrics from Nhentai/Pixiv

psilabs-dev:
This PR adds the option to fetch favorite count from nhentai (tag: "nhentai_favorites:") and bookmark count from Pixiv (tag: "pixiv_bookmarks:").

The PR may be readily used to fetch engagement metrics for the first time, but it may cause problems if called multiple times, as described in issue #1140


psilabs-dev:
I picked the namespaces for some futureproofing but they can be adjusted.

Additionally these are features I'd like to have, but I see metadata plugins often not including these, and there may be a valid reason we don't want to include these options, so if that's the case I'm fine with moving this downstream.

psilabs-dev:
needs some work

Issue #743, opened 2023-01-22
Add Scroll Snap for infinite scroll & single page view


Asinin3:
Rather than have users trying to precisely scroll-down to the bottom of an image to ensure they keep as much of the image on-screen, without the next image showing (infinite scroll) or LRR's navigation/text (single page view). We could avoid that entirely by implementing this: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type

The demo at the bottom of the page showcases how proximity scroll is possible through CSS, and I think it would be a great addition. Ideally set a proximity of around 20% of the top and bottom of an image to avoid people over/undershooting. Would also help with one-handed usage (﻿ ͡° ͜ʖ ͡°)

Difegue:
I briefly tried toying with this in infinite scroll by adding `scroll-snap-type: y proximity` in `body.infinite-scroll #display` but didn't see any changes - I might be missing something? PRs on this welcome.

Asinin3:
@Difegue I can get it to snap with this:

```
#display {
  scroll-snap-type: y mandatory;
  overflow-y: scroll;
  height: 100vh;
}

.reader-image {
  scroll-snap-align: start;
}
```
Its not good enough for a PR, I spent hours trying many, many different approaches and its always the same problem. If I don't set overflow-y, and height to something static, then it doesn't snap to anything. But if I do set it, then it causes the div for the display container to cut off the start of the images, and also gives the image div a scrollbar, so now you have two, one for the images in the div, and one for the pages. Using the same CSS in other classes e.g body doesn't work. Using "body.infinite-scroll #display" works, but i preferred .reader-image so it doesnt snap to the margins.

Tried setting it in other divs, tried many  different approaches with flex, setting the size to the object, etc but not getting anywhere. I'm a total noob when it comes to this and mostly spent hours testing with ChatGPT and got nowhere.

Asinin3:
Closing, i think my PR for snap on spacebar makes more sense, and I spent hour trying to get the snap in CSS to work, and could only get it to snap on one axis of the image, of course its possible to fix, but it required changes to the HTML layout to add more divs, or JS changes. And figured spacebar made more sense anyways.

PR #1208, opened 2025-03-22
Add Scroll snap (Help Needed)

Asinin3:
Spent 7 hours straight trying many different methods to get scroll-snapping to work on both the top, and bottom of each image, and have it not snap to the bottom of the last image. But doing this via CSS was proving near-impossible. There's no way to set two snapping positions on one element, and if you use two separate elements, then well your margins overlap and tha causes issues too.

This needs either HTML or JS changes to function properly. And allow us to set the scroll type somewhere.. other than inside the HTML Element. Currently thats the only place it can be set in CSS, setting it to any other element or class will cause the image div to have another scrollbar, and now you can only scroll images from the middle of the screen and it breaks the interface.

Well, this extremely basic change in CSS does work, but there's no snapping when scrolling up to previous page, or scrolling to top of current.. Its not ideal. I played around with trying different margins and start/end on the elements but it only works with one at a time.

https://github.com/Difegue/LANraragi/issues/743

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -1,5 +1,9 @@
 /* Global LANraragi CSS. Don't mess !*/
 
+html {
+  scroll-snap-type: y mandatory;
+}
+
 body {
   touch-action: manipulation;
 }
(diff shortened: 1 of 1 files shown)


Asinin3:
Just gonna close this, not sure why I opened. Other PR with JS change should be enough and this in its current state.. is not worth keeping. Scroll snap is nice but some people also might not like it making it annoying to scroll down for settings. Whereas space is only used for that.

PR #1211, opened 2025-03-24
 Spacebar snapping logic & next page fix

Asinin3:
The new spacebar logic to go to next page doesn't work on Firefox at all, whereas the previous logic only worked on some pages. It turns out my theory about fractual pixels was correct, if an image had its pixel count rounded up/down which changed the dimensions/aspect ratio, LRR wouldn't know it was at the end of the page.

Fixing that could've been as simple as setting a pixel offset of 3 here:

if (($(window).height() + $(window).scrollTop()) >= (LRR.getDocHeight() - 3)) {
    (Reader.mangaMode) ? Reader.changePage(-1) : Reader.changePage(1);
}

But I decided to make the spacebar logic better, now you can easily spam it to keep going down, snap to bottom of images, or skip to next image if already at the bottom. And the opposite for going up and to previous pages with shift+space.

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -318,7 +318,7 @@ Reader.handleShortcuts = function (e) {
         break;
     case 32: { // spacebar
     //Break if overlay is open, browser detects repeatkey but not held, or webtoon gallery and in infinite scroll
-    if ($(".page-overlay").is(":visible") || e.repeat || (Reader.infiniteScroll && Reader.tags?.includes("other:webtoon"))) break;
+    if ($(".page-overlay").is(":visible") || e.repeat || (Reader.infiniteScroll && Reader.tags?.includes("webtoon"))) break;
     e.preventDefault();
 
     const scrollDown = !e.shiftKey;
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -16,6 +16,7 @@ Reader.initializeAll = function () {
     Reader.initializeSettings();
     Reader.applyContainerWidth();
     Reader.registerPreload();
+    document.documentElement.style.scrollBehavior = 'smooth';
 
     // Bind events to DOM
     $(document).on("keyup", Reader.handleShortcuts);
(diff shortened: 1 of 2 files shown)


Asinin3:
Sorry for 2nd PR, turns out that when you rename a branch it deletes old branch without any warnings, and also closes PR. I restored it, but before re-opening the PR I deleted the commits that I needed to remove, which then caused me to be blocked from re-opening the PR due to the force push not happening while PR was open..

Asinin3:
New commit adds more things. 

- It will skip all the snap logic if the gallery has the webtoon tag, and is in infinite scroll mode (Since pages are continuous snapping would just be annoying)
- On testing, I noticed that the JavaScript scroll functions would ignore the browser keyrepeat detections, so pushing spacebar for 0.3 seconds instead of 0.2seconds would result in two inputs, it was hard to make it always only do one on keypress, that was fixed
- Better detection of current position and prevention of going past current page when snapping, seems to work well now even if in infinite scroll and theres one image that is landscape i.e double page spread, and nowhere near as high as the other images
- Added logic to the changePage function as I noticed it was inconsistent in infinite scroll, if you scrolled down a bunch sometimes it wouldn't know what page you were on and going next could for example randomly move you back 40 pages.

I tested it all a bunch and it seems good.

public/js/reader.js
```diff
@@ -315,16 +316,44 @@ Reader.handleShortcuts = function (e) {
...
-            (Reader.mangaMode) ? Reader.changePage(1) : Reader.changePage(-1);
-        } else if (($(window).height() + $(window).scrollTop()) >= LRR.getDocHeight()) {
-            (Reader.mangaMode) ? Reader.changePage(-1) : Reader.changePage(1);
-        }
-        // spacebar is always forward regardless of reading direction, so it needs to be flipped
-        // to always result in a positive offset when it reaches the changePage() logic
+    case 32: { // spacebar
+    //Break if overlay is open, browser detects repeatkey but not held, or webtoon gallery and in infinite scroll
+    if ($(".page-overlay").is(":visible") || e.repeat || (Reader.infiniteScroll && Reader.tags?.includes("other:webtoon"))) break;
+    e.preventDefault();
```
Difegue, line 322: Since this function is only fired on `keyup`, I still get the browser default behavior when I press space.  

This means that on default settings:  
- When holding down space, the reader scrolls down to the bottom of the current page
- When releasing the key, this function triggers and it goes to the next page

I don't actually hate this -- It's one less keystroke to fully read a page and switch to the next!  
I've found it to be inconsistent when quickly pressing and releasing space however, it sometimes does both steps and scrolls + goes to next page, or only scrolls down.  

Should we just add another `preventDefault` on `keydown` for spacebar to kill the default browser behavior?

Difegue (changes requested):
A couple of questions, but nice work overall, this is an old one 😤

public/js/reader.js
```diff
@@ -315,16 +316,44 @@ Reader.handleShortcuts = function (e) {
...
-        if (e.originalEvent.getModifierState("Shift") && (window.scrollY) === 0) {
-            (Reader.mangaMode) ? Reader.changePage(1) : Reader.changePage(-1);
-        } else if (($(window).height() + $(window).scrollTop()) >= LRR.getDocHeight()) {
-            (Reader.mangaMode) ? Reader.changePage(-1) : Reader.changePage(1);
-        }
-        // spacebar is always forward regardless of reading direction, so it needs to be flipped
-        // to always result in a positive offset when it reaches the changePage() logic
+    case 32: { // spacebar
+    //Break if overlay is open, browser detects repeatkey but not held, or webtoon gallery and in infinite scroll
+    if ($(".page-overlay").is(":visible") || e.repeat || (Reader.infiniteScroll && Reader.tags?.includes("other:webtoon"))) break;
```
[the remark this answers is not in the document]
Asinin3: Some pages in webtoon galleries aren't very tall because they were cut. For example you might have three images that are all part of one strip but the image in-between 2nd image isnt tall. I can for sure change it to not check the namespace though. I think checking image dimensions might not work all that well unless it has a bunch of checks to see if the average height on the majority of the pages is tall, but wouldnt that just be slow to run?

public/js/reader.js
Asinin3: I tried preventDefault first, but it caused the function to just not work at all. If I don't use either function then pushing the spacebar for 0.2 seconds could result in it activating twice, it was so sensitive that it became difficult to push and let go of space fast enough for it to only register once.

I'll give it another shot and see if I can get it to work on keydown instead.

public/js/reader.js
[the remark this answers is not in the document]
Difegue: hadn't thought of cut pages -- I guess it's ok to just check on the unnamespaced tag then.

public/js/reader.js
Asinin3: Tried your suggestion to disable keydown but after spending some hours testing.. I don't like it not being able to scroll when held with normal browser function. But allowing browser scroll on key.repeat then blocking on keydown doesnt work because it will skip the repeat check in first instance due to not being a repeat on first run, then it just breaks and will never hit the key.repeat as its already been pushed and hit the break.

Not sure how I can handle this

Difegue: hmm, well I don't think I have any better ideas.  
As it stands this is already better behavior and the inconsistency really only happens if you're used to the old way it works... We'll see how it goes with users, imo fixing the snap issue is a bigger win.

Difegue (approved):

psilabs-dev:
https://github.com/user-attachments/assets/68b8412f-2f84-4091-a88b-b997b02b1bf1

I'm using safari on macos. Not sure what happened, but the space key feels way too sensitive now compared to before...

Asinin3:
@psilabs-dev Yea, thats what I was working on trying to fix, probably should've stated it wasnt ready for merge. Its ok if you push space really fast but its hard to really fix it properly. I can make it so spacebar prevents default behaviour on non-webtoon galleries but it will mean that holding space wont let you use it to quickly scroll down in infinite reader through a bunch of pages which I wanted to keep.

I'll try a bit more and if I cant get it to work I'll just resort to disabling default behaviour on keydown

psilabs-dev:
@Asinin3 oh that's not a problem, it's a dev branch after all. (Though I felt like I was being insanely fast with the spacebar presses... 😅)

I'm not really concerned about this bug for now, as long as it is a known issue.

Issue #979, opened 2024-06-04
Option to disable storing pages in temp folder


lambdareader:
I couldn't find any information in the docs to disable it.
Setting temp folder size to 0 doesn't work. 
I don't have it on SSD so it just creates a copy on my hard drive for no reason at all.

Most manga/comic readers (komga, kavita, yacreader) just read/unzip single pages of a cbz in memory through random access. It is much faster than unpacking it, writing it and then reading.

Could this be implemented please? It just creates a lot of extra hardware strain and the gains are negligible for a single user. Even with SSD it wouldn't be worth it. Hard drives take a fraction of a second to read a few MB for a page if you have them on a NAS spinning the entire time.  
tbh I'm confused why caching is standard. It would only make sense if there are multiple users or you are on desktop and the drive platter are spinning down when not used. But aren't most using this on a NAS?
Sure it would use a bit of memory to unzip pages in memory but isn't this a case where it should be used?

Difegue:
👋 The full extract/caching was designed to avoid performing the unarchiving job which (used) to take a bunch of time.. This is a pretty old part of the design, the server used to rely on `unar` instead of libarchive and that couldn't do random access.  

The server does support random access when you hit individual page endpoints, but it's true that it always copies to the filesystem first. I'm not sure how much work it'd take to make it all happen in-memory.  
As a workaround, I think you could just point the temp folder to a RAM disk?

lambdareader:
Thanks for the fast reply. 
Wow, stupid of me to not think of simply using a ramdisk. I'll be doing that, though always having to load all pages at once is a bit wasteful. 
It would be pretty neat to eventually rewrite it.

As another thought. I'm running lanraragi on a very low powered device. Is there a option to just generate page thumbnails if i open the archive overview and not when i open the archive (or disable them entirely)? It causes the server to lag because all cores are at 100% if there are triple digit pages. I don't really use the overview very much, so every time i want to read a big archive it takes 30 seconds before i can turn to the next page.

Difegue:
No specific option, but I believe the changes from #885 will help as they restrain page thumbnail jobs to run sequentially on one worker only.

Asinin3:
Yeah, this would be great if we can just disable the cache saving to disk. A lot of people use LRR locally, and I also have all of my archives uncompressed, using the store level of compression. Which means they should be fast to open, and the caching of galleries is fairly redundant.

siliconfeces:
[...]

Issue #1205, opened 2025-03-22
Automatic Cache Cleanup issues (corrupted images & endlessly loop)


Asinin3:
LRR's Automatic Cache Cleanup has two issues which can both be fixed by the same solution.

1. If your maximum cache size is lower than the gallery filesize you opened, LRR will be unable to display any images, and endlessly loop performing cache cleanups over and over again.
2. If LRR performs a cache cleanup while opening a gallery, images can become corrupted, as they were included in the cache cleanup

Both issues can be fixed by adding some logic to the automatic cache cleanup, only removing old cache that isnt from the gallery that is currently open would solve both problems. I also think the cache size shouldn't be able to be set lower than 500MB or 1GB.

To reproduce 1st issue set temp cache to 10MB, and open a gallery.
To reproduce 2nd issue set temp cache to 500MB, and open a gallery with LRR log open, keep pushing R to open a random gallery and once you see the cache cleanup occur go through the images and you should find a corrupted image.

Difegue:
Fixed by the upcoming CHI implementation.

PR #1212, opened 2025-03-25
A new cache system, or how I learned to stop writing to disk and love redis^H^H^H^H^H CHI

siliconfeces:
This is something resembling a proof of concept for not writing stuff quite as much to disk. It instead only uses redis for caching in a pretty half-assed way by reusing an existing database and relying on maxmemory-policy for evictions.

This might never get merge-worthy since I've partially taken a sledgehammer-approach to it, but may at least be useful for testing (#979) and such? Stuff like PDF still hits disk, and ensuring that even works anymore haven't really been a priority.

It doesn't immediately crash and burn, but expect stuff to be broken!


siliconfeces:
Okay so it doesn't even use Redis anymore, there were a nifty module called CHI that was a great fit.

It seems that this has solved issues where images could randomly become cut off earlier (such as #1205), I've been running this for a little while and things seem to be really solid now.

Unfortunately, you get a lot less insight into the cache now. You don't get to know how large the cache is anymore, for instance. This will probably annoy some people.

psilabs-dev:
this is interesting, I'd like to test it out

The corrupted image issue is also something I want to look for fixes for. Thiugh not a big issue it feels like a thousand mosquito bites over time, it's pretty annoying when it happens

siliconfeces:
I flipped the PR to non-draft, as there's no _extremely obvious_ bugs right now. It's probably a good idea to have a positive amount of testing, despite the net LOC being slightly negative.

Difegue:
Thanks, I'll try to give this a spin relatively soon.

lib/LANraragi/Model/Reader.pm
```diff
@@ -113,9 +113,6 @@ sub build_reader_JSON ( $self, $id, $force ) {
 
     return {
         pages => \@images_browser,
-
-        # filesizes => \@sizes,
-        job => $jobid
```
Guerra24, line 118: Ichaival (when force refreshing) and LRReader (always) use this to properly wait for extraction, Ichaival silently fails and LRReader shows an error message (as it is expected to be present).

lib/LANraragi/Model/Archive.pm
```diff
@@ -72,7 +73,7 @@ sub update_thumbnail {
     # Get the required thumbnail we want to make the main one
     no warnings 'experimental::try';
     try {
-        $newthumb = extract_thumbnail( $thumbdir, $id, $page, 1, 1 )
+        $newthumb = extract_thumbnail( $id, $page, 1, 1 )
```
Guerra24, line 76: Updating thumbnails from the api appears to not be working at all.

Guerra24 (commented):
Also, I did a quick benchmark by loading all thumbnails and:

Old file cache: 30s
New cache: 2m2s

The new cache was fully warmed up and then restarted so that no in-memory stuff was used, OS caches were also dropped before both tests.

Fully cold cache performance is awful, completely unusable.

siliconfeces:
> Also, I did a quick benchmark by loading all thumbnails and:
> 
> Old file cache: 30s New cache: 2m2s
> 
> The new cache was fully warmed up and then restarted so that no in-memory stuff was used, OS caches were also dropped before both tests.
> 
> Fully cold cache performance is awful, completely unusable.

Huh weird, what are you running on?

Guerra24:
> Huh weird, what are you running on?

Host: Debian 12 + Docker. Root on a SSD y content on ZFS /w HDD.

The test simply loads all cover thumbnails of 4063 archives.

I think thumbnails should be keep persistent, I don't see a reason why they should be handled by the page cache. If the container is updated or they are evicted they need to be generated again.

The concurrency lock system used by Cache::FastMmap appears to be the main performance issue, it locks the entire file instead of per entry. This is not an issue on a browser since they limit connections, but on other clients they do not limit them which leads to requests getting blocked or dropped due to the workers fighting for the lock on the file.

siliconfeces:
>The test simply loads all cover thumbnails of 4063 archives.
Sequentially or in parallel? If parallel, roughly how many requests simultaneously? I'd like to make a repeatable k6 test for that.

I see no problem letting thumbnails stay on disk in a non-evicting place (heck, CHI can help with that), but I'd like everything else to be full sanic as well.

Guerra24:
> Sequentially or in parallel? If parallel, roughly how many requests simultaneously? I'd like to make a repeatable k6 test for that.

Parallel with 6 requests.

siliconfeces:
Okay, so I've ran some tests, done some benchmarks, conducted some science, etc. This branch ends up faster(!) on my dev VM if thumbnails are created with one or very few parallel requests, dev is faster otherwise.

Hopefully resolvable, could be that HDD access times go BRRR when attempting to create multiple thumbnails in parallel, as dev limits that to something sane.

I'll close this in the meantime until it scales better.

PR #1214, opened 2025-03-29
Implement archive navigation and read previous/next feature

### LINKED ISSUE #299: [Suggestion] Add a next/previous gallery button

### LINKED ISSUE #1187: [UI Feature Request] Read Next/Read Previous Archive
When in reader mode, add the option to read the next archive or the previous archive, if available, IF the user entered reader mode via a search as derived from the index page. Present reasonable access methods for this functionality when applicable, eg key shortcuts, buttons, and/or in-screen click. This function may be hidden if the user directly went into reader mode, eg from a blank tab.

This is inspired by Pixiv's next/previous artwork functionality when viewing any artwork, though this functionality exists in almost all photo gallery applications. However, the proposed functionality is much more flexible than simply viewing in chronological order; next and previous are derived by user choice of search and sort and landing arcid. (Perhaps even more powerful: dynamically retreat to the correct page containing the last archive the user has read...) Related: #299

Useful for reading a series of archives connected together via metadata, or for scrolling through artworks without retreating back to the index page. Say I do a search and get results 1-100. I can start by reading archive 5, then jump to 6 directly in reader mode, then 7, and so on, without ever going back once to the home page.

Basic idea: given a search filter, sort namespace and the current arcid, it is possible to find the next and previous arcids in the list, and this is deterministic in an unchanging contents directory. Search results are equivalent to producing a doubly-linked list. Obviously, this is only a theoretical implementation and should not be done in practice; a time/memory optimized implementation is required.

When jumping to the next archive or previous archive, a question is "which page to go to"; we have two good choices: jump to the first page, or jump to the page in-progress. Perhaps an option in the reader settings would suffice.

I'd be open to designing/developing this if it's wanted. But this is not a high priority feature, as I need to work on the rest of my backlog

psilabs-dev:
Issues:

- #1187 
- https://github.com/Difegue/LANraragi/issues/299

This is a preview (docker image [link](https://github.com/psilabs-dev/LANraragi/pkgs/container/lanraragi/384448640?tag=dev-0.9.40-1180-2fe84858cb501dac1be1f89d61f11bae9560ac87)) of an implementation for reading previous/next archives without going back to the index page. Feel free to try it out and give feedback!

To aid in the browsing experience, I've added two keyboard shortcuts:
- shift+L/R: go to first page/last page
- square bracket: read previous/next archive

Additionally, any action that results in going back from page 1 or going forward from the last page will automatically redirect to the previous/next archive.

Currently there are some limitations which I've decided to impose to make this happen:
1. navigation support is limited to archives from the search results/datatables page; i.e., this is not supported when reading archives in the carousel or when pasting an archive link and visiting it directly (no referer). Mainly the reason I want to limit it to search results is because I think that's where most users would be using it. However, it is possible to extend this feature to support carousel navigation, that would be a second-phase feature.
2. archive-jumping functionality is supported only when the user is not in full-screen viewing mode. I've tried changing archives while in full-screen, but it's just really ugly with the browser kicking you out of full-screen mode when changing URLs. It's probably possible to change archives while maintaining full-screen, but that might be a second-phase feature too. (though I would *very* much like to implement this)

There are still some concrete things to add to the current PR. There's the actual icon to click to read previous/next (currently the only way to directly jump is with squre brackets, I actually forgot about this because I'm used to keyboard navigation lol), I'm thinking of using [forward-step](https://fontawesome.com/icons/forward-step?f=classic&s=solid). The other thing to do is exception handling in the case of missing archives or general changes in datatable results. Then clean up on some logic/remove debt.

We might also want to discuss how to integrate tankoubons in this PR too.

---

### Design

There are two components to this PR: archive navigation and reading previous/next functionality.

Archive navigation refers to the functionality of keeping datatable (DT) state when moving between archives. When reading archives, we need to know what archives are ahead of us/before us, at all times, while navigation is supported, and this information must be persisted through a reading session.

Further, this foresight must extend beyond the current DT page; if a user reaches the last archive of page 1, they should be able to move to the first archive of page 2, of whatever DT search result was yielded from the search filter applied at the point of entering reader mode from the search page.

This foresight logic is implemented at the first/last archives of the DT list, and a fetch for the upcoming DT list will be performed if not already cached, keeping the number of archives IDs stored at around 200-300 (or the search limit) at any given time.

Read previous/next functionality is based on the implementation of navigation, but its logic is straightforward.

A consequence of implementing read previous/next is the problem of returning-to-index. Without the DT query/filter, we would always return to page 1 and lose our search filter. By maintaining DT query state throughout the reading session, the navigation support will also allow us to go back to the correct DT index page.

This implementation persists these (new) variables to local/session storage:

- `navigationState` (session): to enforce when navigation should be supported
- `datatablesPageSize`: part of maintaining DT query state
- `currentSearch`: same reason as page size
- `selectedCategory`: same reason as page size
- `currDatatablesPage`: same reason as page size
- `currArchiveIds`: for reading previous/next
- `previousArchiveIds`: pre-emptively loaded when user approaches end of search results
- `nextArchiveIds`: same reason as previous archive IDs

Also the commit 04e6e01deee5ca13fd4fda731a524466b74d4b11 fixes some rare instance where you get "Fetch API cannot load http://localhost:3000/api/minion/2301 due to access control checks" errors. I considered putting this as a separate PR, but I don't think anyone would encounter this bug unless you're rapidly swapping between archives, which is enabled by this feature


public/js/reader.js
```diff
@@ -783,7 +881,13 @@ Reader.changePage = function (targetPage) {
         }
         destination = Reader.currentPage + (Reader.mangaMode ? -offset : offset);
     }
-    Reader.goToPage(destination);
+    if (destination < 0) {
+        return Reader.readPreviousArchive();
```
psilabs-dev, line 885: tbh not too sure about being able to jump around so easily. I was considering doing a double-tap for archive hopping when using arrow keys, so I might also experiment with the accessibility here

psilabs-dev:
> Does the Random gallery function work with it (R Keybind)? 

This feature doesn't support the R-key unfortunately, because I'm not sure how these functionalities are compatible with each other. Currently the feature is limited only to galleries from search query results, so it would be disabled if you press the r-key, use direct navigation, or enter from carousel.

> adding shift+A/D to go to previous/next gallery

A/D should work with this. The read previous/next gallery is wired onto the previous/next page functionality, so anytime you go out of bounds of the current gallery using existing page switching methods, you'll end up in the neighboring one.

psilabs-dev:
fffff Should be stable 🤞

Fixed the last of the bugs (that are caused by me: If you encounter carousel and stat loading type errors while going back to index, that is not a navigation issue that is an existing safari issue!). The main big one was the slow loading of index page on browser back, because changing to another archive erased some data (see https://github.com/psilabs-dev/LANraragi/pull/39). Now browser back should always return to the index page, and it "should" use bfcache and not take any longer than before PR.

Below is the flowchart for the expected behavior of read next archive and reader initialization if it helps at all (AN stands for archive navigation):
lrr-reader-next-page drawio

A CSS change was made at 968d3725eafe095d98934ccc35f268e0137e49cc to remove the undesirable behavior for small (mobile) clients:
Screenshot 2025-05-07 at 23 18 24

I think this is all I *reasonably* do for this feature for now. After wrestling with navigation + full screen for a while, I think (whether it be true or out oif ignorance) it would be better if we turned the reader page to be an SPA, within which we swap between archives more seamlessly. Any way that removes the need to reload the HTML that kicks the user out of full screen would help.

public/js/reader.js
```diff
@@ -1016,7 +1128,191 @@ Reader.handlePaginator = function () {
...
+}
+
+/**
+ * Return to the index page, preserving the search filter from when 
+ * the user went from index to reader.
+ * 
+ * If for some reason history is unavailable, fall back with search
+ * parameters.
+ */
+Reader.returnToIndex = function () {
```
psilabs-dev, line 1291: There is a situation where returnToIndex removes the search filter in the search box (although the search results are correct). Kind of an annoying problem.. One fix would be to do a page reload (?), but that would also re-trigger the search API, which is even more expensive and annoying and the reason the implementation uses history.back to begin with. 

I'm thinking there should be a way to recover the search filter somehow... when going back? but currently the initialization methods aren't called

psilabs-dev: Not a regression from this pr, but an existing bug with safari

psilabs-dev:
undrafting this as no changes are planned at this moment

alethiophile:
I had planned to implement similar functionality as this for tankoubons, as part of my tankoubon UI project. It seems like I should probably just adapt this instead?

Difegue:
> I had planned to implement similar functionality as this for tankoubons, as part of my tankoubon UI project. It seems like I should probably just adapt this instead?

Not quite I think; I haven't given the tank UI PR a deep look yet but ideally the aggregation of archives within the reader should use the interfaces that were added as part of the ToC implementation.

For switching from one tank to the next though, I think that's where this PR would come in.

psilabs-dev:
Ok, seems there is a bug caused by the slideshow functionality. Will have to look into this.

psilabs-dev:
Previously when slideshow was enabled, it would stop at the first page of the next archive. I consider this buggy behavior, so now slideshow will persist across all archives, like a multi-archive slideshow.

psilabs-dev:
After thinking about it a bit, I've decided to remove the popstate logic, so that archive navigation pushes archives into history, and browser-back will pop state/go to the previously read archive (instead of going straight back to index). This is in contrast to previous logic, where users may navigate between 20 archives, but if they hit BB they instantly go back to index (though it might not have worked 100% of the time).

This means that now the only way to go back to index page (immediately) would be to click on the "return to index" icon. If we want to have a keyboard shortcut, I would prefer using an alternative to browser-back, and keep archives in browser history.

@Difegue currently navigation uses keypress "["/"]", but do you otherwise have a preference for French keyboards/compatibility? We can also do shift+a/shift+d.

locales/template/en.po
```diff
@@ -1073,6 +1073,15 @@ msgstr "Your keyboard arrows (and the spacebar)"
 msgid "Touching the left/right side of the image."
 msgstr "Touching the left/right side of the image."
 
+msgid "When reading an archive from search results, you can also navigate between archives using:"
+msgstr "When reading an archive from search results, you can also navigate between archives using:"
+
+msgid "The square bracket keys"
```
Difegue: Brackets is def a bit jank for azerty keyboards but I don't necessarily mind it given most people will be on qwerty... I don't know if there's existing SOTA for this in other reader apps we could pilfer, might be worth taking a look

public/js/mod/index_datatables.js
```diff
@@ -34,6 +34,13 @@ export function initializeAll() {
         doSearch();
     });
 
+    // Mark datatables-originated reader links so Reader can decide whether to enable cross-archive navigation.
+    // Excludes anything inside the carousel; that's tagged separately in Index.initializeAll.
```
Difegue:
```suggestion
    // Excludes anything inside the carousel; that's tagged separately in ContextMenu.initializeAll.
```

public/js/mod/index_datatables.js
```diff
@@ -67,6 +74,9 @@ export function initializeAll() {
         data: "tags", className: "tags itd", name: "tags", orderable: false, render: renderTags,
     });
 
+    // Store the page size in localStorage for use in the reader
+    localStorage.setItem("datatablesPageSize", Index.pageSize.toString());
```
Difegue, line 78: This could technically also be leveraged by an /api/info call in the Reader, but it seems a bit overkill to do so. (And you'd potentially have desyncs if someone is boneheaded enough to change the page pref then open Reader)

public/js/mod/server.js
```diff
@@ -95,7 +95,11 @@ export function callAPIBody(endpoint, method, body, successMessage, errorMessage
  */
 export function checkJobStatus(jobId, useDetail, callback, failureCallback, progressCallback = null) {
     let endpoint = new LRR.ApiURL(useDetail ? `/api/minion/${jobId}/detail` : `/api/minion/${jobId}`);
-    fetch(endpoint, { method: "GET" })
+    fetch(endpoint, {
```
Difegue: 🤔 ? Not sure why this is needed in this PR, might be a leftover

public/js/reader.js
```diff
@@ -268,6 +274,15 @@ export function initializeAll(trackProgressLocally, authenticateProgress) {
     force = params.get("force_reload") !== null;
     currentPage = (+params.get("p") || 1) - 1;
 
+    // Set up archive navigation state from the entry source (datatables vs carousel vs direct nav)
+    await setupArchiveNavigation();
+
+    // Intercept return-to-index so we can re-apply the search/page state the user came from
+    $(document).on("click.return-to-index", "#return-to-index", (e) => {
+        e.preventDefault();
```
Difegue: Could probably just remove the href in the template for the button instead of bothering to block it here.  

It breaks the functionality if there's no JS but Reader just doesn't work without JS anyway so

public/js/reader.js
```diff
@@ -718,7 +739,8 @@ function handleShortcuts(e) {
     }
     switch (e.which) {
         case 8: // backspace
-            document.location.href = $("#return-to-index").attr("href");
+            e.preventDefault();
```
Difegue: Not sure why you need a preventDefault here

public/js/reader.js
```diff
@@ -1818,7 +1872,11 @@ function generateThumbnails() {
     };
 
     const fetchThumbsForArc = function(arc) {
-        fetch(new LRR.ApiURL(`/api/archives/${arc.id}/files/thumbnails`), { method: "POST" })
+        fetch(new LRR.ApiURL(`/api/archives/${arc.id}/files/thumbnails`), {
+            method: "POST",
+            mode: "same-origin",
```
Difegue: Same question here as for the other same-origin I guess

public/js/reader.js
```diff
@@ -1879,21 +1937,37 @@ function changePage(targetPage, resetAuto = false) {
     }
     let destination;
     if (targetPage === "first") {
-        destination = mangaMode ? maxPage : 0;
+        const firstPage = mangaMode ? maxPage : 0;
+        if (currentPage === firstPage) {
```
Difegue: Is this behavior we actually want? It feels off for the "first page" shortcut to warp you to the previous archive if you hit it twice.

public/js/reader.js
```diff
@@ -1915,6 +1992,217 @@ function getFilename(index) {
...
+            if (previousArchiveIdsJson && currArchiveIdsJson) {
+                const previousArchiveIds = JSON.parse(previousArchiveIdsJson);
+                localStorage.removeItem("previousArchiveIds");
+                localStorage.setItem("currArchiveIds", previousArchiveIdsJson);
+                localStorage.setItem("nextArchiveIds", currArchiveIdsJson);
+                previousArchiveId = previousArchiveIds[previousArchiveIds.length - 1];
+                const currentDTPage = parseInt(localStorage.getItem("currDatatablesPage") || "1", 10);
+                localStorage.setItem("currDatatablesPage", currentDTPage - 1);
+            } else {
+                LRR.toast({ text: "This is the first archive" });
```
Difegue: Needs localization

public/js/reader.js
```diff
@@ -1915,6 +1992,217 @@ function getFilename(index) {
...
+async function loadNextDatatablesArchives() {
+    if (localStorage.getItem("nextArchiveIds")) {
+        return JSON.parse(localStorage.getItem("nextArchiveIds"));
+    }
+    const currentDTPage = parseInt(localStorage.getItem("currDatatablesPage") || "1", 10);
+    return loadDatatablesArchives(currentDTPage + 1);
+}
+
+function readPreviousArchive() {
+    const isIphone = /iPhone/.test(navigator.userAgent);
```
Difegue: Safari-specific issue I suppose? Might need a comment so we don't forget the context

public/js/reader.js
```diff
@@ -1915,6 +1992,217 @@ function getFilename(index) {
...
+                localStorage.setItem("currDatatablesPage", currentDTPage - 1);
+            } else {
+                LRR.toast({ text: "This is the first archive" });
+                return;
+            }
+        } else {
+            previousArchiveId = archiveIds[archiveIndex - 1];
+        }
+        if (autoNextPage) {
+            sessionStorage.setItem("autoNextPage", "true");
```
Difegue, line 2080: Feels like you could just throw this in the URL as a query param instead of adding one more key to browser storage (even if it is the session one). 

Might be an overoptimization tho

public/js/reader.js
```diff
@@ -1915,6 +1992,217 @@ function getFilename(index) {
...
+            if (nextArchiveIdsJson && currArchiveIdsJson) {
+                const nextArchiveIds = JSON.parse(nextArchiveIdsJson);
+                localStorage.removeItem("nextArchiveIds");
+                localStorage.setItem("currArchiveIds", nextArchiveIdsJson);
+                localStorage.setItem("previousArchiveIds", currArchiveIdsJson);
+                nextArchiveId = nextArchiveIds[0];
+                const currentDTPage = parseInt(localStorage.getItem("currDatatablesPage") || "1", 10);
+                localStorage.setItem("currDatatablesPage", currentDTPage + 1);
+            } else {
+                LRR.toast({ text: "This is the last archive" });
```
Difegue: Ditto re: loc

templates/reader.html.tt2
```diff
@@ -284,6 +290,7 @@
 <!-- -->
 [% BLOCK arrows %]
 <div class="sn paginator">
+    <a class="fa fa-backward-step page-link" style="font-size: 1.5em;" value="outermost-left"></a>
```
Difegue: Those should probably get hidden if we're not in archive navigation mode

public/js/reader.js
```diff
@@ -1915,6 +1992,217 @@ function getFilename(index) {
...
+        window.location.replace(newUrl);
+    } else {
+        LRR.toast({ text: "This is the last archive" });
+    }
+}
+
+/**
+ * Loads the archives for the given datatables page so the Reader can navigate
+ * between archives across DT page boundaries without re-rendering the index.
+ * TODO: given this can drift from how index builds DT search requests we might
```
Difegue, line 2127: Yeah this all feels a bit wasteful since at its core, all you need here is a list of archive IDs and we're stuck running full search API calls...  

The server has the list of IDs for a search ready to go at any time via LRR_SEARCHCACHE, so it should ideally cost nearly nothing.  

I'm wondering if this feature shouldn't keep track of the search cache key instead?  
`$category_id-$filter-$sortkey-$sortorder_inv-$newonly-$untaggedonly-$grouptanks-$hidecompleted`
You can derive the Index URL relatively easily from that for returns too. (minus page we'd still need to keep track of). 

If we need a new small API endpoint to deliver a list of IDs from a cache key, I'm fine with that.

Difegue (changes requested):
The overall logic is sound, although I have some concerns with the complexity this needs atm to redo searches in the background.. Feels like we could make this much leaner. 

I would also consider spinning off the archive navigation into its own file/module but maybe that's a job for the dethiccening force.

public/js/reader.js
psilabs-dev: yeah an api would help here, at the time (last year) I wanted to keep this feature frontend only so it's not too entangled

public/js/reader.js
psilabs-dev: Given that `fscreen.inFullscreen = () => !!fscreen.fullscreenElement;` is cast to false on iphone, i'd say this line can probably be removed.

public/js/mod/server.js
psilabs-dev: I recall this is safari-related, but I don't remember what so removing it.
If it comes back, i'll see what to do

public/js/reader.js
psilabs-dev: I guess the question is why someone would hit first page twice?

psilabs-dev: eh, might as well just remove this and keep first/last pages idempotent

public/js/reader.js
psilabs-dev: this was from an old implementation where the PR did a bunch of history altering stuff, should be removed

templates/reader.html.tt2
psilabs-dev: done

public/js/reader.js
psilabs-dev: agree removed

public/js/reader.js
psilabs-dev: Not sure how that would play out, because that makes the URL dependent on the user toggling slideshow button?

locales/template/en.po
psilabs-dev: afaik from a google search, komga doesn't use keys for navigation, mangadex does (","/".", tho see https://forums.mangadex.org/threads/reader-alternative-default-forward-backward-key.2105682/) and supports keybind customizations, suwayomi ctrl+n only for next archive.

So we could go ","/"." tho personally I wouldn't touch "ctrl+" keybinds as a mac user

psilabs-dev: Updated to mangadex way. Used e.key here bc "." keycode is not consistent across keyboards (and also https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which states the use of which is deprecated and shouldn't be used anyways)

psilabs-dev:
Did some last-minute testing with mangadex navigation, don't think there's anything else to change here, keeping behavior of the api threads same

groutoutlook:
...do you mind rebase to the latest commit.. would like to test this PR

psilabs-dev:
> ...do you mind rebase to the latest commit.. would like to test this PR

done

psilabs-dev:
Unified the sort id to the namespace and added sort to the URL query params.

Ftr, this may break some bookmarks on the browser which store the numeric sort id, but idk how much demand there is for back compat here tbh

public/js/reader.js
```diff
@@ -269,6 +275,14 @@ export function initializeAll(trackProgressLocally, authenticateProgress) {
     force = params.get("force_reload") !== null;
     currentPage = (+params.get("p") || 1) - 1;
 
+    // Set up archive navigation state from the entry source (datatables vs carousel vs direct nav)
+    await setupArchiveNavigation();
+
+    // Intercept return-to-index so we can re-apply the search/page state the user came from
+    $(document).on("click.return-to-index", "#return-to-index", () => {
```
Difegue: I think this can be moved with the other document click hooks now.  
(Comment is also out of date, this doesn't intercept anymore, it is the only binding for the button)

public/js/reader.js
```diff
@@ -1920,6 +1988,219 @@ function getFilename(index) {
...
+        window.location.replace(newUrl);
+    } else {
+        LRR.toast({ text: I18N.ReaderLastArchive });
+    }
+}
+
+/**
+ * Loads the archives for the given datatables page so the Reader can navigate
+ * between archives across DT page boundaries without re-rendering the index.
+ * TODO: given this can drift from how index builds DT search requests we might
```
Difegue, line 2127: TODO is outdated now i guess?

public/js/reader.js
Difegue: I was thinking you'd just add `&autoNextPage=true` in `newUrl` and parse that on Reader reload.  

(also once again, not necessary imo)

Difegue (approved):
I think we could simplify the logic a bit further now that we have the API - It'd be possible to just requery that more often instead of having the client keep track of previous/current/next IDs in three separate localStorage objects.  

Then again, this has been open for a while and we're due a Reader revamp in the next release so I'm fine merging that as-is.  

The sort changes are nice, it does make more sense than just trying to keep track of column indices...

public/js/reader.js
psilabs-dev: Reason i kept is bc the client still has two ways of building search, once here and once in the index page

psilabs-dev: 2b38ec24acb8951f45a8cef1b98689d615bc024c probably most demonstrates the current drift risk

public/js/reader.js
psilabs-dev: updated this

public/js/reader.js
Difegue: Right yeah I see it, not much to be done unless we unify search URL building at some point

PR #1217, opened 2025-03-30
Handle Github API rate limits more clearly

psilabs-dev:
Sometimes when visiting LRR I get "Error getting changelog for new version", and this happens to be due to github API rate limits. Just added some logs to make it obvious that it's a github rate limit problem (not a server problem).


public/js/index.js
```diff
@@ -437,7 +437,17 @@ Index.checkVersion = function () {
     const githubAPI = "https://api.github.com/repos/difegue/lanraragi/releases/latest";
 
     fetch(githubAPI)
-        .then((response) => response.json())
+        .then((response) => {
+            if (response.ok) {
+                return response.json();
+            }
+            if (response.status === 403) {
+                console.error("API rate limit exceeded: ", response);
```
Difegue, line 445: Probably worth mentioning Github in those messages too for endusers.

public/js/index.js
```diff
@@ -483,7 +493,17 @@ Index.fetchChangelog = function () {
         localStorage.lrrVersion = Index.serverVersion;
 
         fetch("https://api.github.com/repos/difegue/lanraragi/releases/latest", { method: "GET" })
-            .then((response) => (response.ok ? response.json() : { error: "Response was not OK" }))
+            .then((response) => {
+                if (response.ok) {
+                    return response.json();
+                }
+                if (response.status === 403) {
+                    console.error("API rate limit exceeded: ", response);
```
Difegue, line 501: ditto

Difegue (changes requested):

public/js/index.js
Difegue: And since I merged JS translations I guess you get to make those localizable now!

public/js/index.js
```diff
@@ -437,7 +437,17 @@ Index.checkVersion = function () {
...
+        .then((response) => {
+            if (response.ok) {
+                return response.json();
+            }
+            if (response.status === 403) {
+                console.error("Github API rate limit exceeded: ", response);
+                throw new Error("Github API rate limit exceeded");
+            }
+            console.error("GitHub API returned: ", response);
+            throw new Error(`GitHub API returned ${response.status}`);
```
Difegue, line 449: Those error messages will show up in the toast in case of an error -- Since the toast is localized (`I18N.IndexUpdateError`), I believe those messages shouldn't be hardcoded english.  

The console.log statements can remain as they are.

Difegue (changes requested):

public/js/index.js
Difegue: lmk if you need assistance in setting up the localization - It requires editing multiple files and this one has a placeholder which makes it extra annoying, but it's not a huge ordeal either.

psilabs-dev: Oh yeah this will be the next commit I make, just needed the week to be over so I can do a review.

psilabs-dev:
ok, *looks* correct now and the I18n messages work from console. Obviously can't really test this without triggering the rate limit again, so please confirm

Difegue (approved):
Looks perfect, thank you!

PR #1219, opened 2025-04-04
Add a k6 test, hopefully inspiring more k6 tests

siliconfeces:
Adds a single [k6](https://k6.io/) test, which is useful for checking performance in a fairly repeatable way.

Run a test using `k6 run tools/k6/page_fetching.js`.

### DIFF
--- a/tools/k6/covers_cold.js
+++ b/tools/k6/covers_cold.js
@@ -1,0 +1,57 @@
+import http from "k6/http";
+import { sleep } from "k6";
+
+/**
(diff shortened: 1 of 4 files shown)


siliconfeces:
I'll make a few more before the PR!

siliconfeces:
Added cold and warm archive page fetching.

Difegue:
:+1: Seems to work well. I'm not a big grafanahead but the tool can be used standalone and seems to do the job.  

I do wonder if those couldn't be integrated in CI with a dataset somehow to get a baseline for performance at any time.

PR #1220, opened 2025-04-05
Fix the maketext errors (and add some CN translations)

psilabs-dev:
From the i18n change, a series of maketext error logs begin to appear in the LRR logs; these error logs however do not contain errors. The first priority was to obtain these errors. According to claude, error messages are not showing because they may be getting cleared. I'm inclined to believe it, and so wrote an error capturer which confirmed its suspicion.

Capturing the log messages, we get

screenshot

which basically boil down to en.po having typos. Fixing these typos gets rid of the error logs. However, it doesn't remove error logs from a cn browser. To do that we need to match translations with en 1-1.

I took liberty to add a handful of translations. But there are too many, and this is just a bug fix so I'll stop here :p

### DIFF
--- a/locales/template/zh.po
+++ b/locales/template/zh.po
@@ -1112,3 +1112,434 @@ msgid "Return to Library"
 msgstr "返回库"
 
 # ------End of Upload.html.tt2------
+
+msgid "Rescan Archive Directory"
+msgstr "重新扫描指定文件夹"
+
+msgid "Click this button to trigger a rescan of the Archive Directory in case you're missing files, or some data such as total page counts."
+msgstr "Click this button to trigger a rescan of the Archive Directory in case you're missing files, or some data such as total page counts."
+
+# ------Stary of i18n.html.tt2------
+
+msgid "Backup restored!"
+msgstr "还原备份成功！"
+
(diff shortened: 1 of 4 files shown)


locales/template/zh.po
```diff
@@ -1112,3 +1112,434 @@ msgid "Return to Library"
...
+
+msgid "Click this button to trigger a rescan of the Archive Directory in case you're missing files, or some data such as total page counts."
+msgstr "Click this button to trigger a rescan of the Archive Directory in case you're missing files, or some data such as total page counts."
+
+# ------Stary of i18n.html.tt2------
+
+msgid "Backup restored!"
+msgstr "还原备份成功！"
+
+msgid "An error occured while restoring the backup.<br/> Please check the server logs and that your JSON is correctly formatted."
```
psilabs-dev, line 1127: these are considered not translated, but even if there is no translation we would need the key here.

in the future, when adding messages to en.po, we also need to (at least) copy them to cn.po as well, then leave translations for contribution.

Difegue: Agreed, I think we can just solve this on the workflow side for now by making sure every new string is copied to all xlation files. Thanks for looking into it!

PR #1231, opened 2025-04-13
Use CHI for caching page data

siliconfeces:
This PR replaces the previous system for caching pages with CHI. This fixes some wonky behavior where pages could end up cropped.

This is basically #1212, but reduced in scope. Thumbnail handling is essentially untouched now, except the minimum required for the page data change. (I got #1212 working, but there's not much reason to include the other stuff yet).

Testing with `tools/k6/covers_cold.js` shows no difference in performance compared to dev.

### DIFF
--- a/lib/LANraragi/Controller/Api/Other.pm
+++ b/lib/LANraragi/Controller/Api/Other.pm
@@ -6,7 +6,6 @@ use Redis;
 
 use LANraragi::Model::Stats;
 use LANraragi::Model::Opds;
-use LANraragi::Utils::TempFolder qw(get_tempsize clean_temp_full);
 use LANraragi::Utils::Generic    qw(render_api_response);
 use LANraragi::Utils::Plugins    qw(get_plugin get_plugins use_plugin);
 
@@ -69,14 +68,14 @@ sub clean_tempfolder {
     my $self = shift;
 
     #Run a full clean, errors are dumped into $@ if they occur
-    eval { clean_temp_full() };
+    eval { LANraragi::Utils::PageCache::clear(); };
 
     $self->render(
         json => {
             operation => "cleantemp",
             success   => $@ eq "",
             error     => $@,
-            newsize   => get_tempsize()
+            newsize   => 0,
         }
     );
 }
--- a/lib/LANraragi/Controller/Config.pm
+++ b/lib/LANraragi/Controller/Config.pm
@@ -46,7 +46,6 @@ sub index {
         hqthumbpages    => $self->LRR_CONF->get_hqthumbpages,
         jxlthumbpages   => $self->LRR_CONF->get_jxlthumbpages,
         csshead         => generate_themes_header($self),
-        tempsize        => 0,
         replacedupe     => $self->LRR_CONF->get_replacedupe
     );
 }
--- a/lib/LANraragi/Controller/Config.pm
+++ b/lib/LANraragi/Controller/Config.pm
@@ -4,7 +4,6 @@ use Mojo::Base 'Mojolicious::Controller';
 use LANraragi::Utils::Generic    qw(generate_themes_header);
 use LANraragi::Utils::String     qw(trim trim_CRLF);
 use LANraragi::Utils::Database   qw(redis_encode save_computed_tagrules);
-use LANraragi::Utils::TempFolder qw(get_tempsize);
 use LANraragi::Utils::Tags       qw(tags_rules_to_array replace_CRLF restore_CRLF);
 use Mojo::JSON                   qw(encode_json);
 
@@ -47,7 +46,7 @@ sub index {
         hqthumbpages    => $self->LRR_CONF->get_hqthumbpages,
         jxlthumbpages   => $self->LRR_CONF->get_jxlthumbpages,
         csshead         => generate_themes_header($self),
-        tempsize        => get_tempsize,
+        tempsize        => 0,
         replacedupe     => $self->LRR_CONF->get_replacedupe
     );
 }
--- a/lib/LANraragi/Model/Archive.pm
+++ b/lib/LANraragi/Model/Archive.pm
@@ -19,9 +19,10 @@ use LANraragi::Utils::Generic    qw(render_api_response);
 use LANraragi::Utils::String     qw(trim trim_CRLF);
 use LANraragi::Utils::TempFolder qw(get_temp);
 use LANraragi::Utils::Logging    qw(get_logger);
-use LANraragi::Utils::Archive    qw(extract_single_file extract_thumbnail);
+use LANraragi::Utils::Archive    qw(extract_single_file extract_single_file_to_ram extract_thumbnail);
 use LANraragi::Utils::Database
   qw(redis_encode redis_decode invalidate_cache set_title set_tags set_summary get_archive_json get_archive_json_multi);
+use LANraragi::Utils::PageCache;
 
 # get_title(id)
 #   Returns the title for the archive matching the given id.
@@ -230,6 +231,19 @@ sub serve_thumbnail {
         $self->render_file( filepath => $thumbname );
     }
 }
+sub get_page_data ($id, $path) {
+    my $cachekey     = "page/$id/$path";
+    my $content = LANraragi::Utils::PageCache::fetch($cachekey);
+    if ( !defined($content) ) {
+        # Extract the file from the parent archive if it doesn't exist
+        my $redis = LANraragi::Model::Config->get_redis;
+        my $archive = $redis->hget($id, "file");
+        $redis->quit();
+        $content = extract_single_file_to_ram($archive, $path);
+        LANraragi::Utils::PageCache::put($cachekey, $content);
+    }
+    return $content;
+}
 
 sub serve_page {
     my ( $self, $id, $path ) = @_;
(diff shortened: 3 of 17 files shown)


siliconfeces:
There's some commit spam, so squash is probably a good idea.

lib/LANraragi/Controller/Config.pm
```diff
@@ -47,7 +46,7 @@ sub index {
         hqthumbpages    => $self->LRR_CONF->get_hqthumbpages,
         jxlthumbpages   => $self->LRR_CONF->get_jxlthumbpages,
         csshead         => generate_themes_header($self),
-        tempsize        => get_tempsize,
+        tempsize        => 0,
```
Difegue, line 49: you can probably discard this value altogether now if it's not used in the templates.

lib/LANraragi/Controller/Api/Other.pm
```diff
@@ -69,14 +68,14 @@ sub clean_tempfolder {
...
-    eval { clean_temp_full() };
+    eval { LANraragi::Utils::PageCache::clear(); };
 
     $self->render(
         json => {
             operation => "cleantemp",
             success   => $@ eq "",
             error     => $@,
-            newsize   => get_tempsize()
+            newsize   => 0,
```
Difegue, line 78: This is public API so I guess we need to leave that one in for now - Might need to give it a deprecation notice or something in the future.

public/js/server.js
```diff
@@ -190,11 +190,7 @@ Server.triggerScript = function (namespace) {
 };
 
 Server.cleanTemporaryFolder = function () {
-    Server.callAPI("/api/tempfolder", "DELETE", I18N.CleanedTempFolder, I18N.CleanedTempFolderError,
-        (data) => {
-            $("#tempsize").html(data.newsize);
-        },
-    );
+    Server.callAPI("/api/tempfolder", "DELETE", I18N.CleanedTempFolder, I18N.CleanedTempFolderError, null);
```
Difegue, line 193: It's probably worth changing the localization here and in the template, right?  
This doesn't technically clean the temp folder anymore, it just drops the mmap cache.  

There's no real way to clean /temp anymore with those changes, but then again it's not used for much with pages taken out of it -- Mostly PIDs and inprogress thumbnails.

lib/Shinobu.pm
```diff
@@ -94,11 +91,6 @@ sub initialize_from_new_process {
             $inotifysub->($event);
         }
 
-        # Check the current temp folder size and clean it if necessary
```
Difegue, line 97: Probs worth removing the comment at the top of the file that says it does temp folder watch duty now that it doesn't happen anymore.

lib/LANraragi/Utils/PageCache.pm
```diff
@@ -0,0 +1,70 @@
...
+our @EXPORT_OK = qw(fetch put);
+
+my $cache = undef;
+
+sub initialize() {
+    my $logger = get_logger( "PageCache", "lanraragi" );
+    my $disk_size = LANraragi::Model::Config->get_tempmaxsize."m";
+    $logger->debug("Initializing cache, disk size: ".$disk_size);
+
+    # TODO: Make memory cache size configurable
```
Difegue, line 24: Isn't it technically already configurable with get_tempmaxsize? I guess there's no way to change it without rebooting the server.

lib/LANraragi/Model/Archive.pm
```diff
@@ -19,9 +19,10 @@ use LANraragi::Utils::Generic    qw(render_api_response);
 use LANraragi::Utils::String     qw(trim trim_CRLF);
 use LANraragi::Utils::TempFolder qw(get_temp);
 use LANraragi::Utils::Logging    qw(get_logger);
-use LANraragi::Utils::Archive    qw(extract_single_file extract_thumbnail);
+use LANraragi::Utils::Archive    qw(extract_single_file extract_single_file extract_thumbnail);
 use LANraragi::Utils::Database
   qw(redis_encode redis_decode invalidate_cache set_title set_tags set_summary get_archive_json get_archive_json_multi);
+use LANraragi::Utils::PageCache;
```
Difegue, line 25:
```suggestion
use LANraragi::Utils::PageCache qw(fetch put);
```  

So you don't need to call the full namespace afterwards.

lib/LANraragi/Model/Reader.pm
```diff
@@ -18,59 +18,59 @@ use URI::Escape;
...
 
 }
 
 # build_reader_JSON(mojo, id, forceReload)
 # Opens the archive specified by its ID, and returns a json containing the page names.
 sub build_reader_JSON ( $self, $id, $force ) {
-
-    # Queue a full extract job into Minion.
-    # This'll fill in the missing pages (or regen everything if force = 1)
-    my $jobid = $self->minion->enqueue(
```
Difegue, line 69: This PR switches everything to on-demand page jobs only -- Is there really no need for the previous optimization where it'd queue a full extract in parallel when someone would start reading a book?  

I'm all for it since it makes things simpler and it was probably the root cause of the half-unloaded pages to begin with, but I wonder if it doesn't make performance worse for things like the archive overview in the reader.

lib/LANraragi/Utils/Generic.pm
```diff
@@ -168,6 +168,26 @@ sub shasum {
     return $digest;
 }
 
+sub shasum_str {
```
Difegue, line 171: I think base `shasum` isn't used anywhere else with the switch to string reading for thumbnail hashes, so it could probably be removed.

Difegue (changes requested):
A few nits and questions.  

I'd be ok with putting this change in as it is, but it's probably worth waiting a bit for impressions from the API clients beforehand.

Guerra24 (approved):
Aside from the two comments and the job id being removed... I don't see any issues with it and performance is pretty much the same.

LGTM

lib/LANraragi/Utils/PageCache.pm
siliconfeces: That TODO is outdated, used to have an additional memory caching layer but removed it due to multiprocess killing it.

public/js/server.js
siliconfeces: Is there any "workflow" for handling translation changes?

lib/LANraragi/Controller/Api/Other.pm
siliconfeces: Yeah, probably no rush to remove since a single line to ensure backward compatibility is really cheap.

lib/Shinobu.pm
siliconfeces: Done, could remove a use statement as well.

lib/LANraragi/Model/Reader.pm
siliconfeces: There are some k6 tests to do a basic bench for a single archive, although they might not be 100% representative of reality. Didn't notice anything previously, but haven't run them on the latest code.

Arguments could be made both ways, extracting everything in the background may make things faster in certain cases, but IO & cache will get wasted if the user doesn't view the entire book.

lib/LANraragi/Utils/PageCache.pm
```diff
@@ -0,0 +1,70 @@
...
+        driver     => 'FastMmap',
+        cache_size => $disk_size,
+        root_dir => get_temp,
+    );
+}
+
+# Fetches data from cache if available. Returns undef if nothing is there
+sub fetch( $key ) {
+    if (!defined($cache)) {
+        initialize;
```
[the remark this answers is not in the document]
siliconfeces: Agreed, added an explicit initialization call during startup. Left the checks in PageCache as I'd like to have some sort of check that $cache isn't undef (to avoid ugly error messages).

lib/LANraragi/Model/Reader.pm
siliconfeces: The removed jobId is a potential BC break. Can always add a job that does nothing only to get a new, unique jobid if necessary.

Difegue: I want to say we could just add a bogus jobid (-1) on the api response; clients could try feeding it to the minion check job API, but that'd just return an error they hopefully thought of handling already.

lib/LANraragi/Utils/PageCache.pm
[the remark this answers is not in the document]
Difegue: yeah, I was thinking of that too; it'd have worked ok-ish with lazy init like that but might as well throw it in the server boot sequence.

public/js/server.js
Difegue: Not really; Since here we're going to change the meaning entirely it would probably be better to roll out new loc keys instead of modifying the old ones.

Especially since well, the base keys are just the english text so it'd look fucked having a key with an entirely different value.

You need to modify the en.po file first to add the new text, then copy those new keys verbatim to zh.po. (hopefully chinese TLs will step in to actually translate at some point)

Then in templates you can use them as-is; For frontend there's an extra step where you need to update the variables/methods in the i18n.js template. (it's a tt2 that renders to a javascript file.. kinda cursed but it works)

siliconfeces: Tossed some new strings in there. Left them fairly generic in case things other that pages end up being cached (not really thinking about thumbnails, promise!)

Difegue:
Missing the strings in the .po files themselves but I have some loc work pending so I'll do that myself. Looks good otherwise!

PR #1237, opened 2025-04-21
Image Snapping & Navigation improvements

Asinin3:
I have debug overlays active so you can test it for yourself and give feedback **before merge**

Adds more functionality:
- Fixes the sensitivity issue where you had to tap space stupidly fast or it would trigger twice.
-  Continuous** scroll if spacebar is held for > 250 ms  (same as browsers normally do except no initial scroll on keydown then repeating)
- Overshoot Snapping when exiting continuous scroll this snaps you back to the image in the center. It has a rather large 40% snapping radius but this felt good to me when I was testing
- The goto page function has been modified so if you are infinitereader it goes to the bottom of the previous page if going back a page
- Undershoot Snapping this snaps to the edge of the current image if you almost make it to the edge (15%)

Old functions like preventing you from scrolling past the current image on a single press or not doing anything for webtoon strip galleries are still there

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -351,7 +351,9 @@ Reader.initInfiniteScrollView = function () {
         loaded += 1;
         if (loaded === images.length) {
             allImagesLoaded = true;
-            Reader.goToPage(Reader.currentPage);
+            if (window.scrollY === 0) {
+                Reader.goToPage(Reader.currentPage);
+            }
         }
     });
 };
@@ -672,7 +674,7 @@ Reader.goToPage = function (page) {
     Reader.showingSinglePage = false;
 
     if (Reader.infiniteScroll) {
-        $("#display img").get(Reader.currentPage).scrollIntoView();
+        $("#display img").get(Reader.currentPage).scrollIntoView({ block: 'nearest' });
     } else {
         $("#img_doublepage").attr("src", "");
         $("#display").removeClass("double-mode");
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -11,6 +11,15 @@ Reader.currentPage = -1;
 Reader.showingSinglePage = true;
 Reader.preloadedImg = {};
 Reader.preloadedSizes = {};
+Reader.spaceScroll = { timeout: null, animationId: null };
+//Spacebar Scroll Config
+Reader.scrollConfig = {   
+    scrollDist: 75,      // Viewport % distance to scroll
+    underSnap: 13,       // Distance % for snapping to edge of current image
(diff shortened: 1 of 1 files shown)


Asinin3:
Try changing the holdDelay. You can increase it and it wont be as sensitive to the continuous scroll. I thought 250ms felt good but maybe it should be more.

Asinin3:
I can't reproduce your 2nd or 3rd issue on the stable build by the way. I think maybe those issues could be some other issue on nightly? 

I've also just added code to remove image margins on webtoon galleries when in infinite view so you get clean strips with no gaps.

psilabs-dev:
Thanks for the update, I tried it and it definitely is in the right direction 👌

And you're right about the bug, I can reproduce it in both difegue/lanraragi:dev (https://imgur.com/a/lrr-ui-test-FQQMpnl) and on the PR branch (https://imgur.com/a/lrr-ui-test-b-N8ur4Cp). Both instances have images being cut off. However, this is not found at least since commit 1cfd949c45f00221126a2996df5dabf8ed9964fa, so it's definitely a not-too-good regression. I'm surprised this isn't reproducible..

psilabs-dev:
> I'm surprised this isn't reproducible..

After testing this out on macos chrome, I suppose an explanation for why this isn't reproducible is that I can only observe this bug in Safari. Additionally, the scrolling behavior I'm seeing in chrome feels completely different from safari behavior. in that it's way slower in chrome, and when I hold space it slows scrolling down to a crawl.

Difegue:
I'm happy with the changes after having given this a spin in FF and Arc -- Holding down space scrolls down as much as you want if you're in infinite scrolling and the snapping feels good.  

Mashing space correctly scrolls down and then goes to the next page -- It sometimes takes more than 2 presses if you really spam it since it takes time for the initial scrolldown animation to play, but I'm OK with that. (better to undershoot and make sure we scrolled to bottom than to skip ahead too quickly)  

The debug overlays might be useful to keep as console.log information?

Asinin3:
Debug info is now in the console log only.
Just fixing page load/refreshes scrolling to the bottom of first image instead of top then should be good for merge

Asinin3:
Should be good now.

Function for changing pages now just goes to nearest edge on infinite view (so bottom of previous page, top of next page) and the initial load scroll/GoTopage only runs unless if at the very top of a page so you can refresh without is scrolling back to the top. It seemed like it was just there to scroll down so you didn't see the top margin of first image but it didn't play nice in infinite view

Difegue:
Yup, looking good on a retest. Thanks!

PR #1240, opened 2025-04-30
Implement Pixiv download plugin

psilabs-dev:
Port of PixivUtil2 image/manga downloading logic appropriated for LRR


Difegue:
lmao i don't know how to feel about this  

it's massively hacky but at the same time it perfectly fits within the current API so ????

psilabs-dev:
ok lmao I kind of get it now

The download plugin framework fundamentally does not support downloading from sites like Pixiv, it has to take a url that points directly to a downloadable archive.

There is no way to do this without a hack or extending LRR to support filepath downloads like in https://github.com/psilabs-dev/LANraragi/tree/feature/pixivdl-b

this is hilarious

Difegue:
Ah, I thought you'd understood the way the hack works but yeah, that's basically what it does -- The server just asks itself to give it the zip, and processes that through the regular DL pipeline.

I suppose we should just enable filepath support in DL plugins, since asking plugin devs to setup this hack for every plugin is pretty bad.  
The change in that feature branch looks good for that, although there's some questions about cleanup post-DL with that.  
Maybe the server could just give plugins a premade temp directory that we ensure gets cleaned up after a download is complete?

psilabs-dev:
Yeah I was mostly hands off with this implementation, only planning to work backwards when I have time. I just thought it was hilarious how I said I would figure it out when it was clearly impossible

Supporting filepath would be great for the DL plugin. Like you said, we can create a temp directory resource that dl plugins either all have access to/or are provided on request. Then clean up tempdir after plugin is completed. I'll take a closer look at branch b

lib/LANraragi/Plugin/Metadata/Pixiv.pm
```diff
@@ -31,7 +31,7 @@ sub plugin_info {
         author      => "psilabs-dev",
         version     => "0.4",
         description => "Retrieve metadata of a Pixiv artwork by its artwork ID.
-            <br>Supports ID extraction from these file formats: \"{Id} Title\" or \"pixiv_{Id} Title\".
+            <br>Supports ID extraction from these file formats: \"{\$pixiv_id} Title\", \"\$pixiv_id\" or \"\$pixiv_id Title\".
```
psilabs-dev, line 34: I'm tending towards using "pixiv_1234.zip" for pixivdl file downloads, just to keep filenames alphanumeric+underscore and avoid special characters or brackets that could possibly break something/exceed some limit somewhere (and Pixiv titles sure have who-knows-what in them).

Also bc titles are not a constant in Pixiv; a naming strategy involving only pixiv ID guarantees our gallery to have unique artworks (without worrying about changing titles and creating duplicates).

With pixiv metadata set to auto, calling the download plugin will automatically add a title (alongside all other metadata) to it, because the metadata plugin would support whatever format the download plugin is in. I think adding titles to filenames is a recipe for trouble, particularly when the filenames are managed by a service rather than a manual curating process

lib/LANraragi/Utils/Minion.pm
```diff
@@ -343,13 +344,14 @@ sub add_tasks {
             if ($downloader) {
 
                 $logger->info( "Found downloader " . $downloader->{namespace} );
+                my $tempdir = tempdir(CLEANUP => 1);
```
psilabs-dev, line 347: I'm think it's probably easier to just give every downloader plugin a temp directory to work with, instead of adding a plugin parameter/attribute saying whether they want a directory, since it cleans itself up after leaving scope

lib/LANraragi/Plugin/Download/Pixiv.pm
```diff
@@ -0,0 +1,203 @@
...
+
+        # Downloader-specific metadata
+        url_regex => "https?:\/\/(?:www\.)?pixiv\.net\/(?:en\/)?artworks\/([0-9]+)"
+    );
+
+}
+
+# Mandatory function to be implemented by your downloader
+# TODO: maybe rename this to "provide_resource" instead?
+sub provide_url {
```
psilabs-dev, line 36: The issue with renaming this to something else is that it would break 3rd-party plugins. Another compatible approach is to add a `provide_file` alongside `provide_url` and let LRR discriminate between URL-based and filepath-based downloads, but that is definitely a bit more involved

lib/LANraragi/Plugin/Download/Pixiv.pm
```diff
@@ -0,0 +1,202 @@
...
+    return (
+        file_path => $zip_path
+    );
+}
+
+######
+## Pixiv Specific Methods
+######
+
+# sub sanitize_title {
```
Difegue, line 95: commented out code

Difegue (approved):
tbqh I don't review plugin code as deeply since it's not core, but the added filepath download functionality is sweet and simple.  

I'll merge this after 0.9.41 ships just so that release is stable.

lib/LANraragi/Plugin/Download/Pixiv.pm
psilabs-dev: Sure any name works, I'll leave the naming and deprecation updates up to you then.

(Some nhentai/twitter-X DL would be pretty cool)

PR #1286, opened 2025-06-18
Add Nginx documentation for large file downloads

psilabs-dev:
When downloading a file from LRR behind an nginx reverse proxy exceeding 1gb, it will fail if [proxy_max_temp_file_size](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_max_temp_file_size) is not configured (which by default is 1024m), and this happens in both the API and UI. 

The simplest fix is to configure proxy_max_temp_file_size; see https://github.com/NginxProxyManager/nginx-proxy-manager/issues/643. There are 2 choices, 1) disable `proxy_max_temp_file_size`, or 2) set it to a higher value than the file downloaded (e.g. 4gb). For documentation I added the disable, but maybe setting a higher value is more appropriate.

### DIFF
--- a/tools/Documentation/advanced-usage/proxy-setup.md
+++ b/tools/Documentation/advanced-usage/proxy-setup.md
@@ -23,7 +23,8 @@ server {
     index index.php index.html index.htm;
     server_name lanraragi.[REDACTED].net;
 
-    client_max_body_size 0;   <----------------------- And this line here
+    client_max_body_size 0;           <---------- And this line here to disable upload limit
+    proxy_max_temp_file_size 0;       <---------- And this line here to support large downloads
 
     # Cert Stuff Omitted
 


psilabs-dev:
ok did an experiment, indeed it is more complicated than I stated. Modifying proxy_max_temp_file_size to different values does affect the outcome, but it isn't exactly a linear relation:
```
filesize: 1411776kb

1536m -> success
1408m -> success
1280m -> fail
1152m -> fail
1024m -> fail
896m -> fail
768m -> fail
640m -> success
512m -> success
0 -> success
```

Without going too deep into the nginx configuration rabbit hole, I think a recommendation of disabling `proxy_max_temp_file_size` stands for most users.

More links:
- https://serverfault.com/questions/820597/nginx-does-not-serve-large-files
- https://www.ruby-forum.com/t/nginx-proxy-truncates-large-file-downloads-to-1-1g/200134/5 (especially the last comment)

Difegue:
Sounds about right yeah, I thought for sure this was already part of the docs but I guess not!

PR #1288, opened 2025-06-23
Fix thumbnail cropping position

MoveToEx:
When showing thumbnails for wide images, the current style will crop the image to show the top center part of it. This behavior affects user experience with images that are made up by combining multiple images horizontally, and is inconsistent with the description of the crop option (_If enabled, thumbnails that don't fit a regular A4 page will be cropped to only show the **left** side_).  

This PR modifies lrr.css to show the top left part of the image.


Difegue:
Yep, that's better indeed - Thanks!

PR #1289, opened 2025-06-25
"/" to search and other keyboard shortcut features

psilabs-dev:
This PR implements 3 keyboard shortcut features I've wanted to have for some time. I feel these are natural enough behaviors to realize.

in the index page:
- "esc" to close release overlay
- "/" to focus to search bar (precedent: github and youtube)

in the metadata edit page:
- "enter" when in the plugin args textbox to automatically invoke plugin

Additionally, if release overlay is open, "/" will also automatically close the overlay before focusing to search.

Since there's no demand for this (afaik), I was considering writing an issue for this as a feature request then linking the branch, but a PR is just more direct, but let me know on thoughts, suggestions!

### DIFF
--- a/public/js/edit.js
+++ b/public/js/edit.js
@@ -17,6 +17,7 @@ Edit.initializeAll = function () {
     $(document).on("click.tagger", ".tagger", Edit.focusTagInput);
     $(document).on("click.goback", "#goback", () => { window.location.href = new LRR.apiURL("/"); });
     $(document).on("paste.tagger", ".tagger-new", Edit.handlePaste);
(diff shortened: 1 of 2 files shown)


IceBreeze:
> Ah, just realized that in firefox "/" corresponds to the "quick find" functionality.

I don't think this should be a problem, the built in functionality is available if you press `SHIFT+/`

Difegue (approved):
As an azerty keyboard user I can't use / without shift to begin with (unless I caps lock lmao) so I don't think I'll really get to use this.. But I see the value regardless.

PR #1290, opened 2025-06-25
Fix the Ehentai plugin search error issue

Putarku:
Compatibility adjustments have been implemented to handle cases where no search results are found.

`[error] Can't call method "parent" on an undefined value at /home/koyomi/lanraragi/script/../lib/LANraragi/Plugin/Metadata/EHentai.pm line 198`


Difegue (approved):
Nice.

PR #1293, opened 2025-07-03
Updated FAKKU login Plugin to include Useragent field

Nixis198:
Either FAKKU or Cloudflare did some updating and broke the FAKKU metadata plugin.
When pulling the page data, the plugin would return the Cloudflare blocked page. After some testing, giving it a proper useragent fixes the issue.
Updated the login plugin to include a spot to add one.
Let's see how long this lasts until something breaks again.

### DIFF
--- a/lib/LANraragi/Plugin/Login/Fakku.pm
+++ b/lib/LANraragi/Plugin/Login/Fakku.pm
@@ -10,16 +10,13 @@ use LANraragi::Utils::Logging qw(get_logger);
 sub plugin_info {
 
     return (
-        name      => "Fakku",
-        type      => "login",
-        namespace => "fakkulogin",
-        author    => "Nodja",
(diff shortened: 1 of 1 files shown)


Difegue:
I take it the sid cookie needs to match the UA string now? Otherwise I don't see why we couldn't just pre-fill a common useragent here.

Nixis198:
Honestly I don't know. I just threw in my own UA and didn't think about using a common one. I can do some checking and see if a common one will work. Might not do it today though, should be done by tomorrow.

Nixis198:
Locked in and did it faster than I thought.
Kept the spot to add a UA, but if one is not given a default one will be used.
With my testing everything seems to work with the default UA, and it doesn't seem that the sid cookie is matched to the UA.
But if it is for some reason, the user can add their own.

PR #1301, opened 2025-08-02
Metrics Exporter

psilabs-dev:
Issue: https://github.com/Difegue/LANraragi/issues/1080

An implementation of the LANraragi [metrics exporter](https://prometheus.io/docs/instrumenting/writing_exporters/), where library, API and process level metrics are conditionally collected and served through the "/api/metrics" endpoint in the Prometheus exposition format, and with Redis as the shared metrics state. Metrics data is stored in Redis at db4, but we can change to an existing db if we want.

Most of the code written by AI, and then reviewed/rewritten by me. Architectural decisions made by me.

Already tested this on personal prod environments for about a week, will continue doing so 👌

## Demo screenshots and pretty pictures
Things you can do in Prometheus/Grafana, with the metrics provided

Things you can do with the metrics provided

Configuring the metrics exporter settings

Configuring the metrics exporter settings

## 3rd-party implementations and shared state
There are 2 perl implementations (probably more) of metrics exporter, [Net::Prometheus](https://metacpan.org/dist/Net-Prometheus) and [mojolicious-plugin-prometheus](https://github.com/tyldum/mojolicious-plugin-prometheus). The main issue with just using them is that we need a shared state. Net::Prometheus is lower-level but no shared state, while mojolicious uses shared state but is higher level with IPC.

On the other hand with LRR, I want to collect *all* the metrics, and might as well use Redis too since that's what it's good for anyways so I just decided to rawdog it

## Opt-in
Metrics collection and endpoint exposure is optional and opt-in (via the `enablemetrics` setting flag from config), and instructions to enable metrics has been documented.

## OS dependent
Each OS (e.g. macos, windows, linux*) needs an implementation of the process-level metrics collection. Currently only Linux process-level metrics are supported, but people are welcome to contribute for other OS's :)

## Openmetrics and general spec stuff
This implementation ~is~ *was* written to be compliant to the [OpenMetrics 1.0 specification](https://prometheus.io/docs/specs/om/open_metrics_spec), with minor adjustments to conform to Prometheus server capabilities (turns out that Prometheus didn't *actually* support the OpenMetrics spec, despite the spec being the *first thing* that shows up when you do Prom exporter specification research...!). A couple deviations from the OpenMetrics spec:

- The metrics endpoint is exposed at "/api/metrics" instead of "/metrics"; this is flexible though we can change it back ([overview](https://prometheus.io/docs/specs/om/open_metrics_spec/#overview))
- Metrics endpoint is exposed on the same port (3000) as the end user port: this is because having mojolicious support a new port requires adding routing conditions to ensure that the right port serves the right content, which incurs performance costs ([security](https://prometheus.io/docs/specs/om/open_metrics_spec/#security))
- Return headers are normal text instead of "application/openmetrics-text". ([security](https://prometheus.io/docs/specs/om/open_metrics_spec/#security))
- Server info is type "GAUGE" not "INFO" ([info](https://prometheus.io/docs/specs/om/open_metrics_spec/#info))

Still, OpenMetrics has some good practices, so most of its rules were followed. There's also [OpenTelemetry](https://opentelemetry.io/docs/specs/otel/protocol/) which is another thing, but we're sticking mostly to Prometheus exposition.

## Metric collection types
There are 3 broad categories of metrics being collected: API/http, library, and process.

### API
API/HTTP refers to metrics collected by a mojolicious worker handling a single HTTP request/endpoint. Metrics collected include duration and bytes sent/received.

The natural way to handle this passive collection is via mojo hooks.

Also, API metrics group requests by endpoint type. I.e., instead of the full "/api/archives/123456..." endpoint, we use the `Routing.pm` "/api/archives/:id` endpoint. This is to handle [cardinality explosion](https://last9.io/blog/how-to-manage-high-cardinality-metrics-in-prometheus/).

### Library
Library/stats refers to the stats mentioned in the initial issue: how many archives, how many pages, etc.. These are usually aggregated values by another worker process during a file monitor event, so metrics can get these data for free and we don't need a separate periodic hook. (archive byte size, on the other hand...)

And since prometheus servers periodically scrape metrics from LRR, it doesn't make sense for metric scraping to *also* trigger expensive calls that drag the whole server down, so it's best for the metrics API to be as lean as possible.

### Process (Minion/Shinobu)
These are the CPU/memory/FD/IO metrics that one may find in node exporter. Process metrics collection is a passive "process", so it's done on as a 30s recurring task.

## DB Cleanups
There are 3 ways of doing cleanups for metrics: shutdown cleanup, startup cleanup, and TTL (continuous cleanup).
TTL isn't exaclty a valid approach because it violates OpenMetrics specification that metrics should generally exist for the lifetime of the process (and it causes metrics to disappear). That leaves startup and shutdown, but startup is generally more reliable.


psilabs-dev:
I still need to work on cleaning things up (and marinate on Model/Metrics.pm; I'm still contemplating whether to do MetricsFamily refactoring) and reviewing tests, but I have some things we might discuss at this point:

- Do we want db4 to be a dedicated metrics store, or put metrics data in an existing db? Either is fine by me
- What endpoint do we want to expose this in?
- What other metrics (process/API/library) should we collect?

On minion/shinobu, I only chose some basic process metrics to collect (cpu/io/disk), but if there are other process metrics (and there are *dozens* of them) that people find useful I wouldn't mind adding them. I just don't want to add things for no reason

Also, I've left out the implementation of more specific purpose-based types of metrics (e.g. plugin invokations, archive addition/deletion, archive size calculation, etc.) because these are kind of scattered over the place and it's not obvious which layer the metrics collection should be, if we want them.

tools/Documentation/extending-lanraragi/architecture.md
Difegue, line 275: This is going to need a `Redis Database 4` with matching metric descriptors. (Sorry!)

lib/LANraragi/Model/Metrics.pm
```diff
@@ -0,0 +1,437 @@
...
+    # "info" metadata type is OpenMetrics 1.0 format, currently not supported by Prometheus.
+    # push @output, "# TYPE lanraragi_server_info info";
+    push @output, "# TYPE lanraragi_server_info gauge";
+    push @output, "# HELP lanraragi_server_info Server information with version and configuration details";
+    push @output, "lanraragi_server_info{$server_labels} 1";
+
+    # Configuration metrics
+    push @output, "# TYPE lanraragi_has_password gauge";
+    push @output, "# HELP lanraragi_has_password Whether the server has password protection enabled";
+    push @output, "lanraragi_has_password " . (LANraragi::Model::Config->enable_pass ? 1 : 0);
```
Difegue: You can use `$controller->LRR_CONF` for all the Model::Config invocations under there as well.

lib/LANraragi.pm
```diff
@@ -217,6 +219,41 @@ sub startup {
...
+                my $c = shift;
+                $c->stash( 'metrics.start_time' => [ gettimeofday ] );
+            }
+        );
+
+        # Hook to collect API metrics
+        $self->hook(
+            after_dispatch => sub {
+                my $c = shift;
+                LANraragi::Model::Metrics::collect_api_metrics($c);
```
Difegue: This technically captures every Mojo HTTP request, not just API calls. Maybe the naming needs to be adjusted a bit?

Difegue (changes requested):
wow i love devops and dashboards and metrics  

This is more straightforward than it seemed at first glance and it's basically a capture layer over Mojo with some /proc and stats sprinkled over, so I'm fine with this.  

A few notes tho (mostly as replies to the questions above):  

- I'd rather have this be in a new DB, if we absolutely had to smoosh it into an existing one I'd have suggested the minion storage, but this is different/isolated enough.   
- this is basically a supercharged version of `/api/info`, so the endpoint could probably go there (`/api/info/metrics` or similar)
- Is mojo_debug / `[SERVER_URL]/debug` still useful with this stuff implemented? I guess it still has some value for people who don't setup grafana in their homelab (and it keeps full API call cardinality unlike this)
- You're not actually capturing Minion metrics with this yet, right? I'm not sure how easy it is to hook unfortunately...  
- I'd be keen to avoid adding a ton of extra metrics everywhere too, but it seems straightforward to add extra `Model::Metrics` later down the line that would increment counters in the Metrics DB. As it is I wouldn't add any extra ones, let's see how this is used first.

lib/LANraragi.pm
psilabs-dev: Rename to "request"?

psilabs-dev:
yeah I didn't want to complicate this, wanted to keep metrics simple, useful, composable, and not explode (and watch the grafana line go up)

- A new DB (and putting it in DB4) is the simplest approach atm, just to keep metrics state across all the workers and LRR and shinobu. I've played around with in-memory redis alternatives, but that felt like overkill (well, it mostly just didn't work). The only issue with flushing is that tests are slightly longer to run.
- I personally haven't used mojo debug/minion that much, though I'm aware it exists and have opened it before. Minion metrics are not added, though we could wire it up to redis like the rest (and keys for minion are cleaned up in this pre-emptive commit)

Right now the only issue (?) I'm noticing is that the memory metrics are not exactly accurate. It's not obvious, but after 1-6 months I'd gaze at the dashboard and go "huh, how is LRR using more memory than my machine has physically" so something somewhere the cleanup isn't cleanupping (maybe I'll find it tomorrow).

We could probably remove system metrics if we want to offload it to cadvisor or something else. The most value I'd say comes from LRR-specific stats, like database (total archives, pages) and server (request latency).

well it's still pretty even when it's wrong lol

tools/Documentation/extending-lanraragi/architecture.md
psilabs-dev: added some

Difegue (approved):
> well it's still pretty even when it's wrong lol  

spoken like a true prometheus user  

lgtm :shipit:

PR #1305, opened 2025-08-04
Split Windows CI into discrete steps

Guerra24:
*gh actions pain*

Hope this works. If not I will test stuff in a private repo.


Difegue:
Well, we can at least try.

PR #1307, opened 2025-08-04
Disable testing as it times out on ARMv6/v7

Guerra24:


Difegue:
Right, we skip tests on the cpanfile stuff so makes sense to do the same here.  
I think we just never realized Net::DNS::Native was timing out since it was installed in the cpanfile beforehand..

Issue #1215, opened 2025-03-29
[QoL] Add context menu item for "copy URL" and "copy archive ID"


psilabs-dev:
The "copy link" option is generally available for most linkable elements through context menus; it would be a nice qol improvement to include it in the custom context menu as well.

Additionally, we might also want to add the "copy archive ID" option alongside it. Perhaps made available via a "shift" key, overriding the "copy link" option. E.g. "copy link" -> right click, "copy archive ID" -> shift+right click

![Image](https://github.com/user-attachments/assets/a848dc28-2603-4527-8763-87fb64f025d1)

PR #1316, opened 2025-08-12
Add context menu item for "copy URL"

gustaavv:
I have implemented this QoL issue: #1215 .

image

I think it is enough to just implement the "copy link" button, because (A) the context menu should resemble that of the browser (take a look at the following image) for better UX; (B) the link already contains the archive ID; (C) listening to a key event and changing the content on the page is quite difficult. I have to break many things in the template file to achieve this, which is something I want to avoid doing.

image

I used clipboard.js to implement the functionality of writing text to clipboard. I did not use the `navigator.clipboard.writeText()` API because [Chrome requires a secure context](https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined), and many lrr users are self hosting docker containers with HTTP endpoints in their LAN. Of course, I would love to implement the feature without installing a 3rd-party library, but clipboard.js just makes life easier.

There's one thing I am not sure, which is i18n. I only updated `locales/template/en.po`. Is there an easier way to set these three new fields for all languages? Thanks.

### DIFF
--- a/locales/template/en.po
+++ b/locales/template/en.po
@@ -712,6 +712,9 @@ msgstr "Read"
 msgid "Download"
 msgstr "Download"
 
+msgid "Copy link"
+msgstr "Copy link"
+
 msgid "Edit Metadata"
 msgstr "Edit Metadata"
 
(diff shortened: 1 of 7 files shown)


Difegue (approved):
It is unfortunate that we need another JS library just for clipboard access, but i guess that can't be helped since google hates us  

Everything here looks good, including i18n -- There's nothing else for you to do, the strings will end up in weblate and translators will get to them when they have time.

Issue #768, opened 2023-03-01
Auto scroll 


xiazeyu:
Have an idea about set a time to allow auto scroll or auto next page.

Difegue:
With the recent reader JS fixes, I believe this could be as simple as just calling `Reader.changePage(1);` on a timer.  

I don't need this feature, but this should be very easy to implement for someone wanting to get a PR in. 😉

PR #1318, opened 2025-08-13
Add "auto next page" feature

gustaavv:
I have implemented this feature request: #768 

First, there is a new reader option:

image

After configuring it, users can click the "auto next page" button to toggle this functionality:

- off state:
  image
- on state:
  image
  The number is an indicator of a countdown timer, which will decrement every second. When it reaches 0, the reader will automatically turn to the next page.

In addition, I added a keyboard shortcut `n` to toggle this functionality.

The biggest barrier for me when implementing this feature is a UI design decision: find a place to hold the countdown timer. I think we'd better place it near the "auto next page" button (to give a counter example, an absolute-positioning block floating above the manga image will make a bad reading experience). Since the text styles offered by the `fa fa-2x` classes are acceptable,  and we need to make a coherent look and feel of the buttons area, I think it is OK to just replace the button with the countdown number when the functionality is on.

### DIFF
--- a/templates/reader.html.tt2
+++ b/templates/reader.html.tt2
@@ -181,6 +181,7 @@
                 <li>[% c.lh("R: open a random archive.") %]</li>
                 <li>[% c.lh("F: toggle fullscreen mode") %]</li>
                 <li>[% c.lh("B: toggle bookmark") %]</li>
+                <li>[% c.lh("N: toggle auto next page") %]</li>
             </ul>
             <br>[% c.lh("To return to the archive index, touch the arrow pointing down or use Backspace.") %]
         </div>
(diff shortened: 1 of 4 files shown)


templates/reader.html.tt2
```diff
@@ -291,6 +298,7 @@
 
 <div class="absolute-options absolute-right">
     <a class="fa fa-arrow-right fa-2x reading-direction" href="#" title="[% c.lh('Reading Direction') %]"></a>
+    <a class="fa fa-arrows-rotate fa-2x toggle-auto-next-page" href="#" title="[% c.lh('Auto Next Page') %]"></a>
```
Difegue, line 301: Not a fan of arrows-rotate for the feature since it feels more like a refresh button...  

Have you considered `fa-stopwatch`?

Difegue (approved):

PR #1319, opened 2025-08-13
Avoid dependency cycles

siliconfeces:
This PR cleans up some imports with the main focus to avoid a dependency cycle between Config and Logger. It also moves some stuff away from Database to vastly reduce the risk of new cycles.

As a treat, there's also some more explicit imports!

### DIFF
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -22,6 +22,7 @@ use LANraragi::Utils::I18NInitializer;
 
 use LANraragi::Model::Search;
 use LANraragi::Model::Config;
+use LANraragi::Model::Setup qw(first_install_actions);
 
 use constant IS_UNIX => ( $Config{osname} ne 'MSWin32' );
 
@@ -194,7 +195,7 @@ sub startup {
     start_shinobu($self);
 
     # Check if this is a first-time installation.
-    LANraragi::Model::Config::first_install_actions();
+    LANraragi::Model::Setup::first_install_actions();
 
     # Hook to SIGTERM to cleanly kill minion+shinobu on server shutdown
     # As this is executed during before_dispatch, this code won't work if you SIGTERM without loading a single page!
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -195,7 +195,7 @@ sub startup {
     start_shinobu($self);
 
     # Check if this is a first-time installation.
-    LANraragi::Model::Setup::first_install_actions();
+    first_install_actions();
 
     # Hook to SIGTERM to cleanly kill minion+shinobu on server shutdown
     # As this is executed during before_dispatch, this code won't work if you SIGTERM without loading a single page!
--- a/lib/LANraragi/Controller/Api/Tankoubon.pm
+++ b/lib/LANraragi/Controller/Api/Tankoubon.pm
@@ -66,7 +66,7 @@ sub delete_tankoubon {
     my $tankid = $self->stash('id');
 
     my $redis = LANraragi::Model::Config->get_redis;
-    
+
     return unless exec_with_lock( $self, $redis, "tankoubon-write:$tankid", "delete_tankoubon", $tankid, sub {
         my $result = LANraragi::Model::Tankoubon::delete_tankoubon($tankid);
 
@@ -85,7 +85,7 @@ sub update_tankoubon {
     my $data   = $self->req->json;
 
     my $redis = LANraragi::Model::Config->get_redis;
-    
+
     return unless exec_with_lock( $self, $redis, "tankoubon-write:$tankid", "update_tankoubon", $tankid, sub {
         my ( $result, $err ) = LANraragi::Model::Tankoubon::update_tankoubon( $tankid, $data );
 
@@ -107,7 +107,7 @@ sub add_to_tankoubon {
     my $arcid  = $self->stash('archive');
 
     my $redis = LANraragi::Model::Config->get_redis;
-    
+
     return unless exec_with_lock( $self, $redis, "tankoubon-write:$tankid", "add_to_tankoubon", $tankid, sub {
         my ( $result, $err ) = LANraragi::Model::Tankoubon::add_to_tankoubon( $tankid, $arcid );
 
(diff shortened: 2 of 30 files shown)


lib/LANraragi.pm
```diff
@@ -194,7 +195,7 @@ sub startup {
     start_shinobu($self);
 
     # Check if this is a first-time installation.
-    LANraragi::Model::Config::first_install_actions();
+    LANraragi::Model::Setup::first_install_actions();
```
Difegue, line 198: If you're doing a use qw, you shouldn't need to specify the namespace when calling `first_install_actions`.

lib/LANraragi/Controller/Api/Tankoubon.pm
```diff
@@ -107,14 +108,14 @@ sub add_to_tankoubon {
...
 
     my $redis = LANraragi::Model::Config->get_redis;
-    
+
     return unless exec_with_lock( $self, $redis, "tankoubon-write:$tankid", "add_to_tankoubon", $tankid, sub {
         my ( $result, $err ) = LANraragi::Model::Tankoubon::add_to_tankoubon( $tankid, $arcid );
 
         if ($result) {
             my $successMessage = "Added $arcid to tankoubon $tankid!";
             my %tankoubon      = LANraragi::Model::Tankoubon::get_tankoubon($tankid);
```
Difegue, line 117: Might as well import get_tankoubon and remove the namespace here too

lib/LANraragi/Controller/Api/Tankoubon.pm
```diff
@@ -134,14 +135,14 @@ sub remove_from_tankoubon {
...
 
     my $redis = LANraragi::Model::Config->get_redis;
-    
+
     return unless exec_with_lock( $self, $redis, "tankoubon-write:$tankid", "remove_from_tankoubon", $tankid, sub {
         my ( $result, $err ) = LANraragi::Model::Tankoubon::remove_from_tankoubon( $tankid, $arcid );
 
         if ($result) {
             my $successMessage = "Removed $arcid from tankoubon $tankid!";
             my %tankoubon      = LANraragi::Model::Tankoubon::get_tankoubon($tankid);
```
Difegue, line 144: ditto, alongside `remove_from_tankoubon` above as well

lib/LANraragi/Utils/Logging.pm
```diff
@@ -11,6 +11,8 @@ use FindBin;
 use Encode;
 use File::ReadBackwards;
 use Compress::Zlib;
+use LANraragi::Model::Config;
```
Difegue, line 14: Do you actually need the Model::Config use here?

Difegue (changes requested):
Looks alright overall.  
Looking at this I now remember the initial idea was to do qw imports for the Utils:: stuff only, which is why nothing in Model:: had it... But it does look cleaner like this I think!

lib/LANraragi/Model/Backup.pm
```diff
@@ -7,12 +7,16 @@ use utf8;
...
-use LANraragi::Model::Category;
-use LANraragi::Model::Tankoubon;
-use LANraragi::Utils::Database;
+use LANraragi::Model::Category qw(create_category add_to_category);
+use LANraragi::Model::Tankoubon qw(get_tankoubon_list create_tankoubon update_archive_list);
 use LANraragi::Utils::String   qw(trim_CRLF);
-use LANraragi::Utils::Database qw(redis_encode redis_decode invalidate_cache set_title set_tags set_summary);
+use LANraragi::Utils::Database qw(invalidate_cache set_title set_tags set_summary);
 use LANraragi::Utils::Logging  qw(get_logger);
+use LANraragi::Utils::Redis    qw(redis_decode);
```
psilabs-dev, line 15: redis_encode is used in Backup.pm but it's not imported here, causing integration test failures on my end.

lib/LANraragi/Model/Category.pm
```diff
@@ -7,8 +7,12 @@ use utf8;
 use Redis;
 use Mojo::JSON qw(decode_json encode_json);
 
-use LANraragi::Utils::Database qw(redis_encode redis_decode invalidate_cache);
-use LANraragi::Utils::Logging qw(get_logger);
+use LANraragi::Utils::Database qw(invalidate_cache);
+use LANraragi::Utils::Logging  qw(get_logger);
+use LANraragi::Utils::Redis    qw(redis_decode);
```
psilabs-dev, line 12: Same here, redis_encode needs an import

lib/LANraragi/Utils/Database.pm
```diff
@@ -22,12 +22,16 @@ use LANraragi::Utils::String  qw(trim trim_CRLF trim_url);
 use LANraragi::Utils::Tags    qw(unflat_tagrules tags_rules_to_array restore_CRLF join_tags_to_string split_tags_to_array );
 use LANraragi::Utils::Archive qw(get_filelist);
 use LANraragi::Utils::Logging qw(get_logger);
+use LANraragi::Utils::Redis   qw(redis_decode redis_encode);
+use LANraragi::Model::Config;
+use LANraragi::Model::Tankoubon qw(get_tankoubons_containing_archive add_to_tankoubon remove_from_tankoubon get_tankoubon);
+use LANraragi::Model::Category qw(remove_from_category add_to_category get_categories_containing_archive);
```
psilabs-dev, line 28: Model::Tankoubon/Category uses Utils::Database, but Utils::Database also uses Model::Tankoubon/Category?
```
E               It's likely there's a bug with the server at this point.<br />
E               <p>
E                       Undefined subroutine &amp;LANraragi::Model::Tankoubon::invalidate_cache called at /home/koyomi/lanraragi/script/../lib/LANraragi/Model/Tankoubon.pm line 109.
E         
E               </p>
```

siliconfeces: I stopped trying to be fancy with LANraragi::Model stuff, so shouldn't be any more of an issue than on dev currently.

lib/LANraragi/Controller/Api/Tankoubon.pm
siliconfeces: Yeah, I've officially given up on fiddling with models now

Issue #1328, opened 2025-08-16
Extracting browsing page images from PDF files, unable to properly recognize usernames in the Chinese language


horxd33:
LRR Version and OS
LANraragi Version 0.9.5 (Windows)
OS Version Windows 10 22H2
OS Build 19045.6093

LANraragi Windows Settings
LANraragi
D:\LANraragi
Content folder location
D:\LANraragi Data\01
Thumbnail folder location
D:\LANraragi Data\00

Bug Details
Extracting browsing page images from PDF files, unable to properly recognize usernames in the Chinese language

test1
Extract thumbnail page images from PDF files, can normally recognize Chinese language user names, and thumbnail pages can display images normally
```
[Minion] [debug] Generating thumbnail for page 217... (D:/LANraragi Data/00/dd/dde92334c1d7d2badfd40a81934f09fd3ae8071f/217.jpg)
[Archive] [debug] Extracting thumbnail for dde92334c1d7d2badfd40a81934f09fd3ae8071f page 217 from 217.jpg
[Archive] [debug] Extracting page 217.jpg from PDF D:/LANraragi Data/01/VOL.001.pdf
[Archive] [debug] gs -dNOPAUSE -dFirstPage=217 -dLastPage=217 -sDEVICE=jpeg -r200 -o "C:\Users\測試用\AppData\Local\Temp\pkQNd8tPcP" "D:/LANraragi Data/01/VOL.001.pdf"
```

When extracting browsing page images from PDF files, the Chinese user name cannot be recognized normally (the user name will become garbled characters), causing the browsing page image to not display properly
```
[File Serving] [debug] Page /dde92334c1d7d2badfd40a81934f09fd3ae8071f/217.jpg was requested
[PageCache] [debug] Fetch page/dde92334c1d7d2badfd40a81934f09fd3ae8071f/217.jpg
[PageCache] [debug] Cache MISS for page/dde92334c1d7d2badfd40a81934f09fd3ae8071f/217.jpg
[Archive] [debug] Extracting page 217.jpg from PDF D:/LANraragi Data/01/VOL.001.pdf
[Archive] [debug] gs -dNOPAUSE -dFirstPage=217 -dLastPage=217 -sDEVICE=jpeg -r200 -o "C:\Users\å»æç¾\AppData\Local\Temp\0fYyIG2zXQ" "D:/LANraragi Data/01/VOL.001.pdf"
[...]

PR #1330, opened 2025-08-17
Fix/workaround excessively long archive names

Guerra24:
Allows LRR to handle super extra long paths. This is more of a workaround using 8.3 filenames rather than a proper fix but said fix would involve a more invasive change as perl is hardcoded to paths of 260 characters max.

This PR adds 2 dependencies Windows:
- [Guerra24/Win32-FileSystemHelper](https://github.com/Guerra24/Win32-FileSystemHelper)
  C# lib and FFI layer.
- [Guerra24/File-ChangeNotify-Watcher-Win32](https://github.com/Guerra24/File-ChangeNotify-Watcher-Win32)
  File watcher implementation.

The helper has some limitations (single directory only and [hardcoded list](https://github.com/Guerra24/Win32-FileSystemHelper/blob/master/csharp/NativeExports.cs#L32) of file types) as I don't expect other projects to use it.

Both dependencies can be installed using cpanm as long as a dotnet sdk is installed in the default location.

### DIFF
--- a/lib/LANraragi/Model/Upload.pm
+++ b/lib/LANraragi/Model/Upload.pm
@@ -84,7 +84,7 @@ sub handle_incoming_file ( $tempfile, $catid, $tags, $title, $summary ) {
 
     # Add the file to the database ourselves so Shinobu doesn't do it
     # This allows autoplugin to be ran ASAP.
-    my $name = LANraragi::Utils::Database::add_archive_to_redis( $id, $output_file, $redis, $redis_search );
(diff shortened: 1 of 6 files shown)


Guerra24:
Added a fix for #1332 and maybe #1328.

Difegue (approved):
Looks great.  
It'd be nice to eventually get those packages onto CPAN so we don't depend on github for them, but not a big deal.

PR #1331, opened 2025-08-17
Fix reader.js regex pattern

psilabs-dev:
When artist name is the last tag, the UI will not display "{title} by {artist}" because the previous regex requires the artist match to be followed by a ",".

Regex changes:
- enforce artist name to be nonempty
- artist name can end in comma or EOS

The leading `.*` is also removed, it makes no difference in practice.


Difegue (approved):
nice catch, thanks!

Issue #1332, opened 2025-08-17
PDF file names, issues with non-English and numeric names


horxd33:
LRR Version and OS
LANraragi Version 0.9.5 (Windows)
OS Version Windows 10 22H2
OS Build 19045.6093

LANraragi Windows Settings
LANraragi
D:\LANraragi
Content folder location
D:\LANraragi Data\01
Thumbnail folder location
D:\LANraragi Data\00

Bug Details
PDF file names, issues with non-English and numeric names

Tested with an English user name
Username: Administrator
C:\Users\Administrator

PDF file names, issues with non-English and numeric names
The PDF file name contains Traditional Chinese characters. Thumbnails can be extracted, but browse images cannot be extracted.
The PDF file name contains simplified Chinese characters, and the thumbnails and browse images cannot be extracted.

On the preset gallery page, the following error is displayed in the LANraragi Log Console
The test file name is "导读手册.pdf"
```
GPL Ghostscript 10.05.1: Unrecoverable error, exit code 1
Argument "Error: /ioerror in --file--\nOperand stack:\n   (D:/LANr..." isn't numeric in foreach loop entry at D:/LANraragi/lanraragi/lib/../lib/LANraragi/Utils/Archive.pm line 189, <__ANONIO__> line 10.
[Minion] [error] Error building thumbnail: Requested image not found: 8f0308da4d9715ec448757e588c9fd514f6fd442 page 0 at D:/LANraragi/lanraragi/lib/../lib/LANraragi/Utils/Archive.pm line 139, <__ANONIO__> line 10.

GPL Ghostscript 10.05.1: Unrecoverable error, exit code 1
Argument "Error: /ioerror in --file--\nOperand stack:\n   (D:/LANr..." isn't numeric in foreach loop entry at D:/LANraragi/lanraragi/lib/../lib/LANraragi/Utils/Archive.pm line 189, <__ANONIO__> line 15.
[Minion] [error] Error building thumbnail: Requested image not found: 8f0308da4d9715ec448757e588c9fd514f6fd442 page 0 at D:/LANraragi/lanraragi/lib/../lib/LANraragi/Utils/Archive.pm line 139, <__ANONIO__> line 15.
```

When you click a thumbnail with a PDF file name containing Traditional Chinese characters, the PDF file name will be displayed as garbled characters in the LANraragi Log Console.
The test file name is "仙劍奇俠傳_遊戲說明書.pdf"
```
[...]

Issue #1326, opened 2025-08-15
[Windows] Cannot show content after updating to 0.9.50


cheng236:
**LRR Version and OS**
Windows 11 Home 24H2 26100.4946
LRR 0.9.50

**Bug Details**
It should have scanned my archives correctly but no content is shown.
But I can use the same database/folder in my dual boot Ubuntu without issue.
I tried:
-reinstall LRR 0.9.50
-opening in private tab
-clearing all browser data
-using different browsers(Chrome, Brave, Edge, Firefox)
-using mobile device
-using different browser languages(English-US, Traditional Chinese)
All failed.

**Matching Logs**
```ｷﾀ━━━━━━(ﾟ∀ﾟ)━━━━━━!!!!!
Redis error encountered: [KEYS] LOADING Redis is loading the dataset in memory,  at C:/Users/.../AppData/Roaming/LANraragi/lanraragi/runtime/lib/perl5/site_perl/5.38.4/Redis.pm line 321.

Trying again in 2 seconds...
Migrating old settings to new format...
Migrating LRR_CONFIG to database 2
[LANraragi] [info] LANraragi 0.9.50 started. (Production Mode)
[LANraragi] [info] Plugin Detected: Chaika.moe
[LANraragi] [info] Plugin Detected: Chaika.moe api.json
[LANraragi] [info] Plugin Detected: ComicInfo
[LANraragi] [info] Plugin Detected: Copy Archive Tags
[LANraragi] [info] Plugin Detected: Tag Copier
[LANraragi] [info] Plugin Detected: Add Timestamp Tag
[LANraragi] [info] Plugin Detected: EHDL info.txt
[LANraragi] [info] Plugin Detected: E-Hentai
[LANraragi] [info] Plugin Detected: eze
[LANraragi] [info] Plugin Detected: FAKKU
[LANraragi] [info] Plugin Detected: GalleryDL
[LANraragi] [info] Plugin Detected: HDoujin
[LANraragi] [info] Plugin Detected: HentaiAtHome plugin
[LANraragi] [info] Plugin Detected: Hentag
[LANraragi] [info] Plugin Detected: Hentag Online Lookups
[LANraragi] [info] Plugin Detected: Hitomi
[LANraragi] [info] Plugin Detected: koromo
[LANraragi] [info] Plugin Detected: Koushoku/Koharu.yaml
[LANraragi] [info] Plugin Detected: Mayriad's EH Master Script
[...]

PR #1334, opened 2025-08-18
Adjust long -> short path check and other minor tweaks

Guerra24:
~~This should also (hopefully) fix the remaining encoding issues.~~

This PR fixes:
- #1325 
- #1328 
- #1332 
- #1337 
- #1338 

Theorically should fix:
- #1326


Guerra24:
Should be good, I think

PR #1336, opened 2025-08-19
fix: a url not using `LRR.apiURL` wrapper

gustaavv:
It was my mistake when implementing #1316 .


Difegue (approved):
Ah, my bad for not having noticed that either.. This URL stuff is way too easy to forget about, wish I could put it in the linter rules somehow.  

Thanks!

PR #1340, opened 2025-08-21
Bail if cpanm fails in install_package

TomaSajt:
This way, if you don't have internet access while running this script, you'll get an error if you don't already have the package installed.

Previously, even if the cpanm call failed, the script would continue running.

Note: The other `cpanm` call in the script already has similar bail logic.


TomaSajt:
Looks like this broke the Windows build!
https://github.com/Difegue/LANraragi/actions/runs/17139922339/job/48624932266

PR #1343, opened 2025-08-22
Improve BC with non-bundled plugins

siliconfeces:
One day a stable plugin API will be made. Today is not that day.

### DIFF
--- a/lib/LANraragi/Utils/Database.pm
+++ b/lib/LANraragi/Utils/Database.pm
@@ -616,15 +616,13 @@ sub get_arcsize ( $redis, $id ) {
     return $redis->hget( $id, "arcsize" );
 }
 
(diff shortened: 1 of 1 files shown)


lib/LANraragi/Utils/Database.pm
```diff
@@ -615,4 +616,16 @@ sub get_arcsize ( $redis, $id ) {
     return $redis->hget( $id, "arcsize" );
 }
 
+sub redis_encode ($data) {
+    my $logger = get_logger("Database", "lanraragi");
+    $logger->info("DEPRECATED - Please use LANraragi::Utils::Redis::redis_encode instead, this function will be removed at some point");
```
Guerra24, line 621: I don't think this is a good idea, there are plugins that call those in a loop so you would en up spamming the logs. I think properly documenting this in a release, with a migration timeframe, is way better.

PR #1347, opened 2025-08-29
Embed search link to artist name for archives

psilabs-dev:
> "Oh, this is from *this* artist, would be nice if I had a quick way to get all their works..."

Turns the artist name (if exists for an archive) to a link in reader mode.

image

When clicked, opens search by the exact artist name.

image

Seems to still work with special characters as well:

image

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -144,7 +144,6 @@ Reader.initializeAll = function () {
                 const artistSearchUrl = `/?sort=0&q=artist%3A${encodeURIComponent(artistName)}%24&`;
                 const link = $('<a></a>')
                     .attr('href', artistSearchUrl)
-                    .attr('style', 'color: inherit; text-decoration: underline;')
                     .text(artistName);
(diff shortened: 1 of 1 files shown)


public/js/reader.js
```diff
@@ -140,11 +140,21 @@ Reader.initializeAll = function () {
             // Regex look in tags for artist
             const artist = data.tags.match(/artist:([^,]+)(?:,|$)/i);
             if (artist) {
-                title = `${title} by ${artist[1]}`;
+                const artistName = artist[1];
+                const artistSearchUrl = `/?sort=0&q=artist%3A${encodeURIComponent(artistName)}%24&`;
+                const link = $('<a></a>')
+                    .attr('href', artistSearchUrl)
+                    .attr('style', 'color: inherit; text-decoration: underline;')
```
Difegue, line 147: I'm not sure this warrants a style override compared to the normal look of links within the app.

Difegue (changes requested):

PR #1361, opened 2025-09-17
Integration test updates

psilabs-dev:
Edited 9/27

compare: https://github.com/psilabs-dev/aio-lanraragi/compare/bff0142d79128312950b5ec2fb5be28e321857f2...a6b6780fe89698a133c0f5561f60a55ab0f75dcb

- extends integration test cases to windist deployments
- adds a test case for concurrent LRR deployments
- adds a matrix of auth-related tests (nofun, api key)
- adds discard search cache test
- fix failed archive upload integration tests if server is busy
- fix shinobu integration tests (stop shinobu may not immediately stop shinobu)
- cache docker images between test cases (tho windows tests will take longer anyways)

The auth tests will guard against my current openapi implementation, which for some third-party-library reason overrides existing auth settings.


psilabs-dev:
There's a rare test [failure](https://github.com/psilabs-dev/aio-lanraragi/actions/runs/17849007539/job/50753899530) in windows where the "remote computer" refuses connection. This should theoretically have nothing to do with name resolution, but changing client-side server address from "localhost" to "127.0.0.1" and running the test 10 times didn't reproduce the issue...

psilabs-dev:
Nope, test failures still happening on windows. I'll just remove the windows tests until it is stable, for now we can still run it locally with pytest.

psilabs-dev:
I see, it seems like this is indeed a limitation on windows. In that case, the reported fail should be handled by tests as expected behavior.

psilabs-dev:
> Limiting to around 20 pooled connections

I already limit request concurrency to 8 with a semaphore, but guess I need to make it explicit

psilabs-dev:
I reinstated the windows integration tests, adding a couple tricks to get better stability (besides basic improvements).

Added a cache priming test that runs at start of test_simple to catch network flakes. It’s expected to tank incoming fails to ensure the other cases pass.

See https://github.com/psilabs-dev/aio-lanraragi/actions/runs/18055849601 for a list of ongoing test runs. If you see a passing test with xfail, that's the flake test doing its job.

In case the a case still fails, I added a test case retry in windows runners covering network related errors.

Separately I disabled the checksum requirement for uploads on most tests, this should reduce server memory and cpu requirements.

See attempt 30 for both xfail and rerun in action.

With these changes (and existing semaphore) I don’t think it’s necessary to set a connection limit, or weaken concurrent testing pressure.

There are ~8 post intensive tests with archive upload (excluding flake tank). Looking at 25 recent tests, we have 1 xfail and 1 rerun, a roughly <1% test case failure rate.

There is still the chance that the windows VM itself fails (resulting in github runner losing connection). In this case there’s nothing we can really do.

PR #1369, opened 2025-10-04
remove api key requirement in plugin docs

psilabs-dev:
The "api key requirement" probably shouldn't be specified in the query parameter, and it is not consistent with the other modules (e.g. shinobu, database), I'm suspecting this is just outdated documentation.


psilabs-dev:
Right, I was imprecise. If it is specified, it's not mandatory, and either the other docs should include this as well, or we remove it from the docs, but currently it's not consistent

Guerra24:
Went thru the changelogs because I couldn't remember why.

TL;DR: The API got reworked around 0.6.9~0.7.3. Key param got removed. Got restored in 0.8.8 for OPDS *"only"*.

Should be fine to remove.

PR #1379, opened 2025-10-26
Enable lazy loading for the quick thumbnail list

siliconfeces:
This PR enables lazy-loading for the thumbnails in the archive overview. This can cut down the request spam a metric fuck-ton for large archives, which hopefully makes things a bit less painful.

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -286,6 +286,11 @@ p#nb {
   color: lightskyblue;
 }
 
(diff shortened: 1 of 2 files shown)


Difegue:
I'd tried this back then but didn't realize it needed a css minwidth to actually work!  
I think this will help, a bunch of users don't interact with the overview at all and just use the reader, so those won't get thumbnails genned at all.

PR #1382, opened 2025-10-28
Use fetchpriority to make browsers fetch current image before thumbnails

siliconfeces:
Another minor PR that can improve perceived performance a touch. By using fetchpriority on the active image, the browser SHOULD prioritize that/those over other images (such as thumbnails).

(The effect is partially nullified due to the other PR that prevents thumbnails from being loaded willy-nilly, but eh it's something)

### DIFF
--- a/templates/reader.html.tt2
+++ b/templates/reader.html.tt2
@@ -58,7 +58,7 @@
             </div>
 
             <a id="display">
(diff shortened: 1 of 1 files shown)


Difegue:
Oh that's some new browser tech -- Not sure it's that useful considering lazyloading yeah but let's take it regardless.

PR #1388, opened 2025-10-30
Demo: customize integration test origin with workflow dispatch

psilabs-dev:
- Docs: https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_dispatch
- Tutorial: https://www.youtube.com/watch?v=Sb_zLeHEVqQ

Adds an option of triggering a manual workload for users to be able to run integration tests from their own forks and refs if desired.

In order for workflow dispatch/manual workload triggering to occur, the branch containing "workflow_dispatch" configuration must be default.

Only users with write permissions to the repository will be able to trigger these workflows (see [docs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)). This means an external PR authors of "difegue/LANraragi" cannot trigger workflows against a custom integration test origin, so nothing changes for this project from a PR author's view.

But repo/fork *owners* will have the ability to change where integration tests are run from. For example in my fork, I'll be able to change my integration test ref to "feature-test-branch" to run workflows against some LRR feature, allowing me to run nightly cross-platform integration tests without git committing a ref change in push-continuous-integration.yml.

## Screenshot
Screenshot 2025-10-29 at 11 52 49 PM

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -1,4 +1,17 @@
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      integration_test_repository:
+        description: 'Integration testing repository path'
+        required: true
+        default: 'psilabs-dev/aio-lanraragi'
+      integration_test_ref:
+        description: 'Integration testing repository git commit hash or branch'
+        required: true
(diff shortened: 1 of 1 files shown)


Difegue:
I knew of workflow_dispatch but didn't realize you could actually have UI for parameters when starting those from github... I think the capability is pretty useful and non-intrusive like that, so it looks OK to me.

PR #1389, opened 2025-10-31
Log more on archive read failure

psilabs-dev:
If a file fails to open (e.g. during thumbnail extraction), the current logs don't say much. Just another attempt to make problems more transparent (and log all possible info based what libarchive provides).

Previous:
```
[2025-10-31 07:01:12] [Archive] [error] Couldn't open archive, libarchive says:Failed to open '/home/koyomi/lanraragi/content/69829671/pixiv_102851995.zip'
```

After:
```
[2025-10-31 07:36:36] [Archive] [error] Couldn't open archive '/home/koyomi/lanraragi/content/69829671/pixiv_102851995.zip' (exists:no readable:no size:NA) libarchive: Failed to open '/home/koyomi/lanraragi/content/69829671/pixiv_102851995.zip' (errno 2: No such file or directory)
```

### DIFF
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -23,6 +23,7 @@ use Archive::Libarchive qw( ARCHIVE_OK );
 use Archive::Libarchive::Extract;
 use Archive::Libarchive::Peek;
 use File::Temp qw(tempdir);
+use POSIX qw(strerror);
 
 use LANraragi::Utils::TempFolder qw(get_temp);
 use LANraragi::Utils::Logging    qw(get_logger);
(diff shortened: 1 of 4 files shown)


lib/LANraragi/Utils/Archive.pm
```diff
@@ -153,8 +153,8 @@ sub expand {
     return lc($file);
 }
 
-# Returns a list of all the files contained in the given archive.
-sub get_filelist ($archive) {
+# Returns a list of all the files contained in the given archive with corresponding archive ID.
+sub get_filelist ($archive, $arcid) {
```
psilabs-dev, line 157: It might be more helpful imo to also include arcid in the arguments (not many users of get_filelist to refactor anyways), or doing an outer eval instead of touching this sub

lib/LANraragi/Utils/Archive.pm
```diff
@@ -178,8 +179,15 @@ sub get_filelist ($archive) {
```
Guerra24, line 179:
```suggestion

$archive = create_path( $archive );

```

And the create_path from line 205 needs to be removed.

psilabs-dev: done

Difegue (approved):

PR #1398, opened 2025-11-05
Add option to only save server-side progress if authenticated

jacobgkau:
As explained in the interface text, this is useful if you want to sync your progress _and_ want your server to be publicly browsable, but don't want guests or scrapers changing your progress. (This is an additional middle ground between a publicly accessible server and no-fun mode, as a stopgap until actual accounts are added if that ever happens.) The new setting's off by default to retain the current behavior, and is hidden from the config page if local tracking is enabled.

Tested working. Open to changes with the interface text being added (wasn't sure if the second sentence explaining the usefulness was necessary, but many other settings seem to have sentences like that). Also let me know if it needs to be added to other documentation anywhere.


public/js/config.js
```diff
@@ -118,6 +120,11 @@ Config.enable_resize = function () {
     else $(".resizefields").hide();
 };
 
+Config.enable_localprogress = function () {
+    if ($("#localprogress").prop("checked")) $(".serversideprogressfields").hide();
+    else $(".serversideprogressfields").show();
+};
```
Guerra24: Would it be better to have both options work at the same time? If both are enabled use server progress for authenticated requests and local progress for everything else.

Guerra24 (changes requested):
The field needs to be added to the api info endpoint [lib/LANraragi/Controller/Api/Other.pm](https://github.com/Difegue/LANraragi/blob/dev/lib/LANraragi/Controller/Api/Other.pm#L23) and docs [tools/Documentation/api-documentation/miscellaneous-other-api.md](https://github.com/Difegue/LANraragi/blob/dev/tools/Documentation/api-documentation/miscellaneous-other-api.md)

public/js/config.js
jacobgkau: That's a good idea. I'll see if it's not too much more complicated to implement.

jacobgkau: All four combinations of settings seem to be working as expected:

- [X] No Clientside Progress Tracking + No Authenticated Progress Tracking (default)
- [X] Clientside Progress Tracking + No Authenticated Progress Tracking
- [X] No Clientside Progress Tracking + Authenticated Progress Tracking
- [X] Clientside Progress Tracking + Authenticated Progress Tracking

Migration from client-side to server-side (in the ascending direction only) is still working when authenticated progress tracking's disabled. When authenticated progress tracking's enabled, there's no need for migration (since the authenticated user still saves to the server, and guest users shouldn't sync over that).

Guerra24 (approved):
LGTM!

PR #1401, opened 2025-11-06
reverse direction for spacebar hotkey and slideshow when in manga mode

### LINKED ISSUE #1399: RTL reading direction causes spacebar hotkey and slideshow to navigate backwards
**LRR Version and OS**
Please tell which version you're running, as well as your OS and installation method. (Windows, Docker, Native)
v.0.9.60 (Docker)

**Bug Details**
Please detail the bug here.

When the reading direction is changed to right to left, both the spacebar and slideshow will progress the archive backwards. I first notices this sometime in between v.0.9.50 and v.0.9.60 but thought it might be a browser bug so I didn't report it. After upgrading to v.0.9.60 though I experienced this with the new slideshow feature which at first i thought wasn't working since nothing happened after countdown, but when I moved forward some pages then started it, the slideshow started going backwards.

**EDIT**: Apparently this changed much earlier than I remember, [v0.9.31](https://github.com/Difegue/LANraragi/releases/tag/v0.9.31) is the last versioned release in which the spacebar moves to the next page in in RTL mode, at some point between that and [v.0.9.41](https://github.com/Difegue/LANraragi/releases/tag/v.0.9.41) onwards it goes to the previous page.

Expected behavior can be restore by editing [this line](https://github.com/Difegue/LANraragi/blob//v.0.9.60/public/js/reader.js#L393) to be
``` javascript
    let direction = e.shiftKey ? -1 : 1;
    if (Reader.mangaMode) direction *= -1;
```

**Matching Logs**
Post any logs you have encountered when reproducing the bug.  

If possible, switch the app to Debug Mode in order to have more logs.
I don't believe there would be logs for this

**Screenshots**

Same as above, but I can produce a video if that would be desirable

fancydancers:
closes https://github.com/Difegue/LANraragi/issues/1399

When the reading direction is changed to right to left, both the spacebar and slideshow will progress the archive backwards. This is a small patch to reverse (and correct) the direction when `Reader.mangaMode` is `true`.

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -390,7 +390,8 @@ Reader.handleShortcuts = function (e) {
     e.preventDefault();
 
     // Capture direction now so we dont lose it if shift state changes while held
-    const direction = e.shiftKey ? -1 : 1;
+    let direction = e.shiftKey ? -1 : 1;
+    if (Reader.mangaMode) direction *= -1;
     const cfg = Reader.scrollConfig;
 
     if (e.type === "keydown") {


Difegue:
This only fixes the spacebar, but I'll get the slideshow corrected as well.

PR #1402, opened 2025-11-06
Integration test updates

psilabs-dev:
compare: https://github.com/psilabs-dev/aio-lanraragi/compare/328c792313f24d667c73377fd67aadfa7df72d26...41e436b59f605d53307f7e2b420b31fed3d17013

Another update from the integration testing repository with a side of client:

- added 2 integration tests for enabled/disabled CORS preflight
- added test to upload archives to symlinked archive directory
- extended coverage of existing tests to count archives directory files before/after uploads
- enforce no error-severity logs from LRR and Shinobu before end of test cases

Docker-based integration testing has moved from volumes to bind mounts, to simulate a more realistic testing environment. This extends to archives, thumb, redis, and logs. As such, all testing environments require a staging directory on the host.

- The docker workflow now uses buildkit, and logic changes in deployment setup are made to remove intermediate images, volumes, and container artifacts.
- Because bind mount is supported, tests involving 1) user operations on the archive directory (and shinobu archive discovery) and 2) symlinked archives/thumbs/logs are supported. However, archive discovery is too slow (2-3s/archive), so no tests planned for the time being.

When a VM is involved (e.g. macos lima + apple virtualization virtiosf/sshfs), then quick teardown/setup of docker containers + removal/creation of directories create stale FS cache problems which result in LRR seeing ghost files from a previous test run, so an explicit in-VM teardown is called to try to handle this, though success is not guaranteed.

This concern doesn't apply to Github runners, where the whole test environment is in basically the same FS.

Other stuff:
- Logs of internal server errors parse the HTML and return only the error message (no more "\<DOC HTML\>Oops I flubbed..." on workflow logs)
- Handle yet another Windows OS error that comes up during upload
- Display both LRR and Shinobu logs on failure, and display logs for all environments involved on failure

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -10,11 +10,11 @@ on:
       integration_test_ref:
         description: 'Integration testing repository git commit hash or branch'
         required: true
-        default: '328c792313f24d667c73377fd67aadfa7df72d26'
+        default: '41e436b59f605d53307f7e2b420b31fed3d17013'
 
 env:
   INTEGRATION_TEST_REPOSITORY: ${{ github.event.inputs.integration_test_repository || 'psilabs-dev/aio-lanraragi' }}
-  INTEGRATION_TEST_REF: ${{ github.event.inputs.integration_test_ref || '328c792313f24d667c73377fd67aadfa7df72d26' }}
+  INTEGRATION_TEST_REF: ${{ github.event.inputs.integration_test_ref || '41e436b59f605d53307f7e2b420b31fed3d17013' }}
 
 name: "Continuous Integration \U0001F44C\U0001F440"
 jobs:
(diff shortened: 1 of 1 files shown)


psilabs-dev:
OK, guess I found a new species of error

compare: https://github.com/psilabs-dev/aio-lanraragi/compare/41e436b59f605d53307f7e2b420b31fed3d17013...d6db424d1a2fa68ebaea15af915938746f146cad

Updates:
- Reduce the chance of windows sudokuing with concurrency decrease 8->2 and increased workflow timeout window
- Add test case for https://github.com/Difegue/LANraragi/pull/1398
- On failure, fall back to process logs (docker container logs, windows console process logs) when log files aren't available
- When network err during multi-archive upload: skip rest of uploads and retry immediately
- Include host system profile during test session start
- Synthetic tag sampling doesn't use raw poisson now, instead it follows a [zipfian distribution](https://en.wikipedia.org/wiki/Zipf%27s_law) a.k.a. more similar to human language frequency; this applies to many namespaced tags as well like artists (if you end up analyzing your local LRR tag distribution, it will also end up looking zipfian too)

PR #1406, opened 2025-11-11
Add lock to log rotation

### LINKED ISSUE #1405: Concurrent API calls cause log rotation race condition
Uncovered with a benchmark test.

When multiple API calls are called concurrently to LRR, multiple loggers will compete for log rotation, causing the log file to be temporarily deleted and triggering a die:

```
WARNING:aio_lanraragi_tests.helpers:[upload_archive] Encountered server error when uploading archive-03948.zip (message: Internal server error: Can't open file "/home/koyomi/lanraragi/script/../log/lanraragi.log": Permission denied at /usr/local/share/perl5/site_perl/Mojo/Log.pm line 19.). Retrying in 10s (1/4)...
INFO:aio_lanraragi_tests.helpers:[upload_archive] Duplicate upload archive-03948.zip to arcid daed23d187e3a0e07eacfb075ea1730925f74c68; skipping.
WARNING:aio_lanraragi_tests.helpers:[upload_archive] Encountered server error when uploading archive-07528.zip (message: Internal server error: error: could not delete /home/koyomi/lanraragi/script/../log/lanraragi.log: No such file or directory at /home/koyomi/lanraragi/script/../lib/LANraragi/Utils/Logging.pm line 68.). Retrying in 10s (1/4)...
WARNING:aio_lanraragi_tests.helpers:[upload_archive] Encountered server error when uploading archive-07529.zip (message: Internal server error: error: could not delete /home/koyomi/lanraragi/script/../log/lanraragi.log: No such file or directory at /home/koyomi/lanraragi/script/../lib/LANraragi/Utils/Logging.pm line 68.). Retrying in 10s (1/4)...
WARNING:aio_lanraragi_tests.helpers:[upload_archive] Encountered server error when uploading archive-07530.zip (message: Internal server error: error: could not delete /home/koyomi/lanraragi/script/../log/lanraragi.log: No such file or directory at /home/koyomi/lanraragi/script/../lib/LANraragi/Utils/Logging.pm line 68.). Retrying in 10s (1/4)...
```

This should be guarded (maybe with redis), however it should not cause an internal server error.

psilabs-dev:
Issue #1405 . Adds a lock to log rotation, and moves gzip target to prevent new writes from happening.

### DIFF
--- a/lib/LANraragi/Utils/Logging.pm
+++ b/lib/LANraragi/Utils/Logging.pm
@@ -5,18 +5,28 @@ use warnings;
 use utf8;
 
 use feature 'say';
+use feature 'isa';
 use POSIX;
 use FindBin;
+use Time::HiRes;
+use Config;
 
 use Encode;
 use File::ReadBackwards;
 use Compress::Zlib;
+use Mojo::Log;
 use LANraragi::Model::Config;
+use LANraragi::Utils::RotatingLog;
 use LANraragi::Utils::Redis qw(redis_decode);
+use LANraragi::Utils::TempFolder qw(get_temp);
+
+use constant IS_UNIX => ( $Config{osname} ne 'MSWin32' );
 
 # Contains all functions related to logging.
 use Exporter 'import';
 our @EXPORT_OK = qw(get_logger get_plugin_logger get_logdir get_lines_from_file);
+
+our %LOGGER_CACHE;
 
 # Get the Log folder.
 sub get_logdir {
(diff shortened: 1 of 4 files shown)


lib/LANraragi/Utils/Logging.pm
```diff
@@ -43,29 +43,52 @@ sub get_logger {
...
-        my $gz = gzopen( "$logpath.1.gz", "wb" ) or die "error: could not gzopen $logpath: $!";
+        }
 
-        open( my $handle, '<', $logpath ) or die "Couldn't open $logpath :" . $!;
-        my $buffer;
-        $gz->gzwrite($buffer) while read( $handle, $buffer, 4096 ) > 0;
-        $gz->gzclose();
-        close $handle;
+        $redis->quit();
+        die $rotation_error if $rotation_error;
```
psilabs-dev: Part of me thinks logger should never die, given how low-level of a tool this is used in many non-atomic write ops (which leads to corrupt/broken data), and logging should be fail open.

But also don't have a really good alternative, and a die would be much rarer now, so maybe the importance of a die message outweighs the potential data problem? Best would be if the die can be thrown at the end of a transactional operation.

Maybe `say` is better but idk, I doubt I'd ever look for says.

psilabs-dev:
This is done. Went ahead to just implement RotatingLog, among other things. A temporary solution (given it's only been a week), but should fix the current problem of log-baased data integrity issues and a *certain* class of random, logging-induced 500 server errors, while maintaining existing performance.

RotatingLog inherits from Mojo::Log, and does a couple extra things on top:
- handles are created eagerly during init and frequently refreshed
- rotation checks and rotation are done at both init and append time
- rotation is guarded by flock
- supports env configuration of logfiles to retain + max bytes size of logfiles (dev purposes)
- Throughout all my later integration tests (based on python UUID stress testing + log reading), RotatingLog on LRR has never lost log events thanks to flock and rename/unlink. But I'll stay on the safe side and say that up to 0.1% of logs *may* be lost due to rotation.
    - In case you're curious: the current implementation of logger will lose around 0.01-0.05% of logs due to rotation races (provided the logger doesn't die so that you can collect the data)

Additionally, loggers are now cached in-memory per-process, which noticably helped with permission failures (probably due to file races). And if cache fails, we have the following redundancy to escape a die:

- no cache -> create RotatingLog -> retry 3 times -> fallback to Mojo::Log

Current and beta integration tests passed, including windows/docker logging api stress tests on both concurrent logging pressure and append-time log rotation. I won't include them here.

QAs bc decisions are made:
- "Why do we need a class?" to override Mojo::Log `append` and `new`, and get better low-level synergy.
- "Why check only once every 1k times during append?" to reduce syscalls ~~and redis network connections~~, and bc "exactly 1MB" isn't a rotation goal.
- "Why refresh handle?" bc if a file is removed the handle might still point to it and the logs emitted would just disappear
- "Why use win32api?" bc without it, we can only do copy-truncate which is lossy. If we try the same as unix (copy/delete) using only perl file handlers, we'll run into windows perms errors (see below response).
- "Do we really need Win32API::File?" [Probably yes](https://github.com/psilabs-dev/LANraragi/pull/76#issuecomment-3539849064); unless you're familiar with windows API (which I'm not really).
- "Why not use Redis?" bc it's not really a good idea to rely on a separate service (and network) for an internal locking system, despite how easy it would be. But you'll have to consider what happens to rotation if redis goes down.

Some things I should probably note:
- A lock file will be created in the temp directory for purposes of log rotation exclusive lock control. This is going to be used by flock.
- most of my testing is done on API-side python calling + unix-based perl unit testing, not Windows. My observation from gh actions is that even with prefork, mojo only seems to run single-process on windows where a race would never happen.
- On Windows, handles are refreshed on every write. Maybe this is a performance issue maybe not.

Dogfooding this now, lmk if there are any regressions in performance, or Windows-related. I'm curious how the unit tests go, I can only hope they don't fail.

lib/LANraragi/Utils/Logging.pm
```diff
@@ -38,80 +53,80 @@ sub get_logger {
...
-        # Rotate existing logs
-        for ( my $i = 7; $i > 1; $i-- ) {
-            my $j = $i - 1;
-            my $next = "$logpath.$i.gz";
-            my $prev = "$logpath.$j.gz";
-            if ( -r $prev && -f $prev ) {
-                rename( $prev, $next ) or die "error: rename failed: ($prev,$next)";
-            }
+    my $logpath     = get_logdir . "/$logfile.log";
+    my $cache_key   = "$logfile|$pgname";
```
Guerra24: It should be cached per log file. This causes issues with the counter not actually counting correctly as it counts per *"script"* and it breaks rotation on Windows because each logger instance has an open handle to the lock file such that flock exclusive access will never work.

psilabs-dev: Pressure tests pass for me so no complaints here.

Guerra24: With that change it uses the same pgname for everything.
image
image

psilabs-dev: Huh...

psilabs-dev: Switched to mojo log's context feature after a bit of TDD, hopefully it works now (possible refactor notwithstanding).

lib/LANraragi/Utils/Logging.pm
```diff
@@ -31,89 +46,105 @@ sub get_logdir {
...
 
-        # Based on Logfile::Rotate
-        # Rotate existing logs
-        for ( my $i = 7; $i > 1; $i-- ) {
-            my $j = $i - 1;
-            my $next = "$logpath.$i.gz";
-            my $prev = "$logpath.$j.gz";
-            if ( -r $prev && -f $prev ) {
-                rename( $prev, $next ) or die "error: rename failed: ($prev,$next)";
+        if ($ok) {
```
Difegue: IMO this can be flattened a bit, I don't think you really need `$ok` (or the localized error context but it _might_ pollute the later loop so might as well leave it in)  

Using `experimental::try` might also be a valid approach here? 🤔 

```suggestion
        {
            local $@;
            eval {
                    LANraragi::Utils::RotatingLog::refresh_logger_handle($log); 
            };
        
            $cache_refresh_error = $@;
        }
        
        unless ($cache_refresh_error) {
```

lib/LANraragi/Utils/Logging.pm
```diff
@@ -31,89 +46,105 @@ sub get_logdir {
...
+    }
 
-    return $log;
+    my $ctx;
+    if ( eval { $log->isa('LANraragi::Utils::RotatingLog') } ) {
+        $ctx = Mojo::Log->new(parent => $log, context => ["[$pgname]"], level => $log->level);
+    } else {
+        $ctx = $log->context("[$pgname]");
+    }
+    return $ctx;
```
Difegue: This block can probably be extracted into a `sub get_log_context($log)` since it's used twice (and with different indentation styles 👀 )

lib/LANraragi/Utils/Logging.pm
```diff
@@ -31,89 +46,105 @@ sub get_logdir {
...
+                    path    => $logpath,
+                    level   => 'info',
+                    logfile => $logfile,
+                    tempdir => get_temp()
+                );
+                1;
+            };
+            $err = $@ unless $ok;
+        }
+        if ($ok) {
```
Difegue: Similarly here I think you could remove both `$ok` and `1;` from the eval and just do a `unless $err`, but I might be missing a detail

lib/LANraragi/Utils/Logging.pm
```diff
@@ -31,89 +46,105 @@ sub get_logdir {
...
-        }
-    );
+    # Report cache refresh error if exists
+    if ( $cache_refresh_error ) {
+        $log->error($cache_refresh_error);
+    }
 
-    return $log;
+    my $ctx;
+    if ( eval { $log->isa('LANraragi::Utils::RotatingLog') } ) {
```
Difegue: We're on perl 5.36 at minimum so you shouldn't need to eval wrap `isa`, see https://www.effectiveperlprogramming.com/2020/01/use-the-infix-class-instance-operator/  


```suggestion
    if ( $log isa 'LANraragi::Utils::RotatingLog' ) {
```

lib/LANraragi/Utils/Logging.pm
```diff
@@ -7,17 +7,32 @@ use utf8;
...
 
 # Contains all functions related to logging.
 use Exporter 'import';
 our @EXPORT_OK = qw(get_logger get_plugin_logger get_logdir get_lines_from_file);
 
+BEGIN {
+    if ( !IS_UNIX ) {
+        require Win32API::File;
+    }
+}
```
Difegue: This class doesn't do any file work anymore so I don't think it needs the `IS_UNIX` and win32 usings

Difegue (changes requested):
Need to actually give this a spin but I have a few codestyle suggestions to start with.  

I think this is the first time we're actually doing real perl oop with extension classes and all that jazz!

psilabs-dev:
Thanks, yeah the code is kinda messy atm, will work around eventually to get it into a presentable way!

I'm aiming for the unit tests to test against the interface provided by get_logger, so that it's (somewhat) implementation-free. Luckily, logging is a feature for which unit testing works wonders towards development.

There's a lingering issue (not prod-side) of RotatingLog causing my LSP (https://github.com/richterger/Perl-LanguageServer) to act up and create lock/log files in not-so-random places. Not sure what's going on there but once that's out I will undraft

lib/LANraragi/Utils/Logging.pm
psilabs-dev: I think I might need the arrows here

psilabs-dev:
> Not sure what's going on there but once that's out I will undraft

What's going on is I misconfigured the LSP; setting LRR_LOG_DIRECTORY in "perl.env" funneled logfiles to a dedicated logging directory.

Difegue (approved):
Got back to reading through this and all seems good. Let's get it in :shipit:

PR #1408, opened 2025-11-14
add read archive button to edit page

fancydancers:
Adds a read archive button the edit page. Not sure if this is a  wanted feature but I found it quite handy.

read-archive


Difegue:
:+1: lgtm

PR #1409, opened 2025-11-15
Allow right-click context-menu on entire row in compact mode

anon-watcher:
Previously you could only right-click the title of a row to bring up the context-menu. Now can right-click anywhere on a row.


Difegue (approved):
Looks functional, thanks for the improvement!

PR #1416, opened 2025-11-24
Die with more info on open_path

psilabs-dev:
Addresses obfuscated error logs: https://github.com/Difegue/LANraragi/actions/runs/19600552085/job/56131617332#step:12:326

Applicable to other path APIs which require die.

### DIFF
--- a/lib/LANraragi/Utils/Path.pm
+++ b/lib/LANraragi/Utils/Path.pm
@@ -9,11 +9,12 @@ no warnings 'experimental::signatures';
 use Encode;
 use Config;
 use File::Find;
+use POSIX qw(strerror);
 
 use constant IS_UNIX => ( $Config{osname} ne 'MSWin32' );
 
 use Exporter 'import';
-our @EXPORT_OK = qw(create_path open_path date_modified compat_path unlink_path find_path);
+our @EXPORT_OK = qw(create_path create_path_or_die open_path open_path_or_die date_modified compat_path unlink_path find_path);
 
 BEGIN {
     if ( !IS_UNIX ) {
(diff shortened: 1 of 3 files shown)


lib/LANraragi/Utils/Path.pm
```diff
@@ -83,4 +90,33 @@ sub find_path( $wanted, $path ) {
...
+    }
+
+    my $exists          = -e $file ? 'yes' : 'no';
+    my $readable        = -r $file ? 'yes' : 'no';
+    my $size            = -e $file ? (-s _) : 'NA';
+
+    my $err = "(exists:$exists; readable:$readable; size:$size) (errno $errno_num: $errno_str)";
+
+    if ( !IS_UNIX ) {
+        $err .= " (winerr $winerr_num: $winerr_str)";
```
Difegue:
```suggestion
        $err .= " (win32 errno $winerr_num: $winerr_str)";
```

lib/LANraragi/Utils/Path.pm
```diff
@@ -83,4 +90,33 @@ sub find_path( $wanted, $path ) {
...
+sub _file_error_msg {
+    my ( $file ) = @_;
+
+    my $errno_num       = 0 + $!;
+    my $errno_str       = strerror($errno_num);
+    my $winerr_num;
+    my $winerr_str;
+
+    if ( !IS_UNIX ) {
+        $winerr_num     = 0 + $^E;
```
Difegue, line 106:
```suggestion
        # This fetches the value of GetLastError() on Win32 as per perldoc: https://perldoc.perl.org/variables/$%5EE
        $winerr_num     = 0 + $^E;
```

Difegue (changes requested):
Might as well use this in the other two uses of `open_path` in the codebase (shinobu filemap timer and eze files)

psilabs-dev:
> Might as well use this in the other two uses of `open_path` in the codebase (shinobu filemap timer and eze files)

We probably should keep open_path for shinobu, bc we don't die there to start with, the loop is polling until the file is openable

Difegue:
> 
> We probably should keep open_path for shinobu, bc we don't die there to start with, the loop is polling until the file is openable

Hmm my idea was we could tweak the loop to add a try/catch in it but I didn't think it over properly, that'd 100% fuck up the polling 🤔 

looks good as is!

PR #1423, opened 2025-11-29
Fixed broken apostrophe in FAKKU description

Nixis198:
@kazuuuuu on Discord found an issue with the descriptions made from the FAKKU plugin. Before it wouldn't properly decode apostrophes and ampersands, now it wil. Updated the tests to add an apostrophe to check for. Also noticed that if a gallery doesn't have a description on FAKKU, in the HTML it still has one but it is more of an advertisement. Added a check to find that and not return a description if found.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/Fakku.pm
+++ b/lib/LANraragi/Plugin/Metadata/Fakku.pm
@@ -9,6 +9,7 @@ use URI::Escape;
 use Mojo::JSON qw(decode_json);
 use Mojo::UserAgent;
 use Mojo::DOM;
+use Mojo::Util qw(html_unescape);
 
 #You can also use the LRR Internal API when fitting.
 use LANraragi::Model::Plugins;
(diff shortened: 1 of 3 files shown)


Nixis198:
FAKKU is going through major changes with their website. Seems they are battling payment processors and they went scorch Earth on a bunch of tags and got rid of the summaries. Will try to update the plugin when everything settles down.

Azriel-Satan:
> FAKKU is going through major changes with their website. Seems they are battling payment processors and they went scorch Earth on a bunch of tags and got rid of the summaries. Will try to update the plugin when everything settles down.

As it has been a month with the only progress being the deletion of the Fakku forums I am not one to hold out hope and it is getting in the way of downloading.

Could there be an update to:
- Not write description when it is blank or the advertisement message
- An option to disable writing description

These would allow it to operate in the current state without overwriting existing book descriptions and would still be functional if Fakku at some point returns to full functionality.

Nixis198:
The updated plugin in my pull request should still work. If it can't find the description div it returns a blank string. I can add a toggle for it, but it will have to wait until I am free. You will have to manually download the plugin and upload it to your LANraragi instance.

Difegue (approved):
Let's get this in since there haven't been any major changes in a while.

PR #1424, opened 2025-11-29
Add ratings section to reader overlay

fancydancers:
Adds a section to reader overlay to set ratings. Includes buttons to clear and set rating, both will rebuild tags div after setting. 

* add `$` to the end of url returned by LRR.getTagSearchURL to return exact tag. This was particularly noticeable for ratings as `rating:⭐⭐⭐` would also return for 4 and 5 stars but could also result in unwanted results for normal tags.
* move tag object to list functionality found in index.js into function as the same pattern is used for updating reader tags

### DIFF
--- a/templates/reader.html.tt2
+++ b/templates/reader.html.tt2
@@ -141,7 +141,6 @@
                     <a class="fas fa-plus" id="add-category" href="#" title="[% c.lh('Add Archive to Category') %]"></a>
 
                     <h2>[% c.lh("Rating") %]</h2>
-                    <br />
 
                     <select id="rating" class="favtag-btn" style="width:200px" autocomplete="off">
                         <option selected value="">[% c.lh(" -- Select Rating -- ") %]</option>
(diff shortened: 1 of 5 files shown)


templates/reader.html.tt2
```diff
@@ -139,6 +139,23 @@
                         [% END %]
                     </select>
                     <a class="fas fa-plus" id="add-category" href="#" title="[% c.lh('Add Archive to Category') %]"></a>
+
+                    <h2>[% c.lh("Rating") %]</h2>
+                    <br />
```
Difegue:
```suggestion
```

Difegue (approved):
Functional, thanks for this! I think the UI could be improved to be a more canonical ratings control instead of a dropdown, but this is a nice first step.

fancydancers:
Originally I had a more fancy UI with the stars in a single row but it required a fair bit of css and a bit more html so i thought this minimal implementation would be better. I'll see about adding a PR with better UI later on.

PR #1429, opened 2025-12-28
Fix renamed msys2 dependency

Guerra24:


Difegue:
Integration test failures are due to the removed `draw` parameter in search, nbd for now

PR #1434, opened 2026-01-06
Updated RegexParse to allow configuring regex

alethiophile:
An update to the RegexParse addon that allows the user to specify a regex in the UI, rather than needing to edit the source code.

This should be backward-compatible but I haven't tested it. Maybe a good candidate for automated tests.

### DIFF
--- a/lib/LANraragi/Plugin/Metadata/RegexParse.pm
+++ b/lib/LANraragi/Plugin/Metadata/RegexParse.pm
@@ -65,7 +65,7 @@ sub plugin_info {
             {   type => "string",
                 desc =>
                     "Regex to use for parsing",
-                default_value => "(\((?<event>[^([]+)\))?\s*(\[(?<artist>[^]]+)\])?\s*(?<title>[^([]+)\s*(\((?<series>[^([)]+)\))?\s*(\[(?<language>[^]]+)\])?(?<tail>.*)?"
+                default_value => '(\((?<event>[^([]+)\))?\s*(\[(?<artist>[^]]+)\])?\s*(?<title>[^([]+)\s*(\((?<series>[^([)]+)\))?\s*(\[(?<language>[^]]+)\])?(?<tail>.*)?'
             }
         ],
     );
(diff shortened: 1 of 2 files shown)


Difegue:
Yeah, it'd be nice to add a small test for the plugin as part of this, following the examples that already exist for the other plugins.  

The change looks OK otherwise, I'm wondering whether passing the regex as a string doesn't break it in some weird way but that's where tests would be useful.

alethiophile:
It turns out there were some existing tests that were pretty comprehensive already.

I added a few more tests around remaining edge cases, verified the behavior of the old code, then updated tests for the new code. Tests all pass with one exception.

The only difference between the old and the new code per these tests is that a few possible results were not trimmed for whitespace. I've updated the code so that they now are. I think the new behavior is closer to correct, but I could always change it back, if we really wanted total backward compatibility.

lib/LANraragi/Plugin/Metadata/RegexParse.pm
```diff
@@ -181,7 +219,7 @@ sub _classify_item {
 #(\[([^]]+)\])? returns the content of [Language]. Optional.
 #(?<tail>.*)? returns everything that is out of E-Hentai standard for further processing. Optional.
 #\s* indicates zero or more whitespaces.
-my $regex = qr/(\(([^([]+)\))?\s*(\[([^]]+)\])?\s*([^([]+)\s*(\(([^([)]+)\))?\s*(\[([^]]+)\])?(?<tail>.*)?/;
-sub get_regex { return $regex }
+#my $regex = qr/(\((?<artist>[^([]+)\))?\s*(\[([^]]+)\])?\s*([^([]+)\s*(\(([^([)]+)\))?\s*(\[([^]]+)\])?(?<tail>.*)?/;
```
Difegue: This comment block should be updated and moved to the docs now that the regex is user-configurable IMO -- I think this plugin is still pretty hard to customize so most people won't do it, but we should keep the information around.

Difegue (changes requested):
Nice to have the extra tests -- I think the new trimming behavior makes more sense so I don't really mind the compat break.  

Just a few more doc/cleanup requests from me.

Difegue (approved):
:shipit:

PR #1436, opened 2026-01-06
ESLint config migration + package.json dependencies pin

anon-watcher:
The .eslintrc.json format has been deprecated, so I've migrated it to the new "flat" config format. This was done using ESLint's migration process that copies a lot of the original config. I've then further cleaned it up with a few tweaks:
1. Removed the "airbnb" eslint rule-set. It's very old and doesn't reference any new ESLint or ECMAScript features, and is still in the old .json format, making it extra work to integrate.
2. Added jsdoc typing to the config so VS Code intellisense can show rule documentation.
3. Added some missing globals
4. Separated out the previously used rules that are deprecated

Since this required some changes to dependencies in package.json, I also pined all dependency versions. This is good practice, since it prevents random packages from updating when making dependency changes, possibly breaking things, or accidentally downloading a compromised package version. The "marked" package was bumped to 4.0.15 to match the package-lock.json, but everything else already matched.

All the js files have been touched up to  conform to the linting, mainly single-quotes to double-quotes, indentation, and html strings to backticks so double-quotes can be used.
Other changes of note:
- duplicates.js - used localStorage.hasOwnProperty isn't allowed due to ESLint no-prototype-builtins. Swapped with standard localStorage functions
- index_datatables.js - columnCount claimed to not exist but worked in the browser. Turns out it's a feature I've never heard of and shouldn't ever be used, see [Named Element IDs Can Be Referenced as JavaScript Globals](https://css-tricks.com/named-element-ids-can-be-referenced-as-javascript-globals/), so swapped with a standard JQuery selector
- reader.js
  - Typed the handleShortcuts input to match the JQuery event types for keyup and keydown
  - Swapped "keyCode" with "which" as it's (recommended)[https://api.jquery.com/event.which/]
  - Moved all the spacebar event code into a new function Reader.spaceScrollProcessInput as linting recommended not declaring variables within a switch statement, and it seemed weird to be the only input key without a dedicated function

Also updated documentation to replace mention of .eslintrc.json with eslint.config.mjs


public/js/index_datatables.js
```diff
@@ -293,15 +293,14 @@ IndexTable.drawCallback = function () {
...
-        // eslint-disable-next-line eqeqeq
         // get current columns count, except title and tags
         const currentCustomColumnCount = IndexTable.dataTable.columns().count() - 2;
         // check currentSort, if out of range, back to use title
         if (currentSort > currentCustomColumnCount) {
             localStorage.indexSort = 0;
         }
-        if (currentSort >= 1 && currentSort <= columnCount.value) {
-            currentSort = localStorage[`customColumn${currentSort}`] || `Header ${currentSort}`;
+        if (currentSort >= 1 && currentSort <= $("#columnCount").value) {
```
psilabs-dev: I don't think jquery objects have `.value`.

tools/Documentation/extending-lanraragi/architecture.md
```diff
@@ -127,11 +127,11 @@ root/
...
 |  +- lanraragi-systemd.service <- Example SystemD service
 |
-|- lrr.conf      <- Mojolicious configuration file
-|- .perltidy.rc  <- PerlTidy config file to match the coding style
-|- .eslintrc.json   <- ESLint config file to match the coding style
-|- package.json  <- NPM file, contains front-end dependency listing and shortcuts
-+- package-lock.json <- NPM lockfile used by installer/`npm ci` for reproducible builds
+|- lrr.conf           <- Mojolicious configuration file
+|- .perltidy.rc       <- PerlTidy config file to match the coding style
+|- eslint.config.js   <- ESLint config file for linting of JavaScript files
```
psilabs-dev: `eslint.config.js` doesn't exist, it's `eslint.config.mjs` right

eslint.config.mjs
```diff
@@ -0,0 +1,88 @@
...
+            "argsIgnorePattern": "^_",
+            "varsIgnorePattern": "^_",
+        }],
+        "one-var": "off",
+        "prefer-destructuring": ["warn", {
+            object: true,
+            array: false,
+        }],
+        
+        // TODO Deprecated ESLint rules that still work but will be fully removed in ESLint v10
```
Difegue, line 79: Are there going to be replacements for those rules or should we just drop 'em in the future?  

Having to get rid of the indent/quotes ones would kinda suck tbh

public/js/index_datatables.js
Difegue: Yeah, you should use `[0]` to get the underlying value.  

I think the issue in here though is that there was a mistake writing the original code to begin with, this was meant to use the `columnCount` const that's declared in `IndexTable.initializeAll` iirc.  

Since that const is unscoped to this function we can probably just do an `Index.getColumnCount()` call again here... Or maybe just move the initial declaration to an `IndexTable.columnCount` global

public/js/reader.js
```diff
@@ -1046,8 +1054,7 @@ Reader.initializeArchiveOverlay = function () {
 
         // Look at all the numbered keys in notes, aka notes.1, notes.2..
         for (let i = 1; i <= notes.total_pages; i++) {
-
-            if (notes.hasOwnProperty(i) && notes[i] === "processed") {
+            if (Object.prototype.hasOwnProperty.call(notes, i) && notes[i] === "processed") {
```
Difegue: Is this change actually needed? `notes` already being an object I thought you could just call `hasOwnProperty` directly

Difegue (changes requested):

public/js/reader.js
anon-watcher: You're correct that it would work, however one of the recommended eslint rules [no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins) suggests against it due to other patterns that may mess with the prototype space for some objects. Also turns out there's yet another new method of [Object.hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) so I've tested and swapped to that.

eslint.config.mjs
anon-watcher: There won't be replacements within eslint itself, however there are alternatives. There's [eslint-stylistic](https://github.com/eslint-stylistic/eslint-stylistic), which is an eslint plugin that apparently has replacements for all the deprecated rules. There's also prettier, which is rather popular, but would require another separate rule file and another installed extension.

tools/Documentation/extending-lanraragi/architecture.md
anon-watcher: woops, fixed :P

public/js/index_datatables.js
anon-watcher: Swapped to the Index.getColumnCount() method since it's correct and doesn't require setting up another variable

anon-watcher:
Suggestions added, ready for another review

Difegue (approved):
:shipit:

eslint.config.mjs
Difegue: Noted, thanks! I guess we'll add a dependency on stylistic when we move to v10 and those rules are removed.

PR #1437, opened 2026-01-06
Support /mpv/ links for EHentai downloader

fancydancers:
Adds support for /mpv/ links on the EHentai downloader plugin, I found it convenient when using with the Tsukihi extension so I didn't have to go back to the gallery view.

### DIFF
--- a/lib/LANraragi/Plugin/Download/EHentai.pm
+++ b/lib/LANraragi/Plugin/Download/EHentai.pm
@@ -26,7 +26,7 @@ sub plugin_info {
             'forceresampled' => {type => "bool", desc => "Force resampled archive download", default_value => "0"}
         },
         # Downloader-specific metadata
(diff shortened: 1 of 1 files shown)


Difegue:
Looks about right, any reason this is in Draft status? The Win integration tests are being a bit finicky but I don't think this is caused by the regex change here.

fancydancers:
My main purpose for this is for use with Tsukihi but the current behavior of the patch is not ideal. After some testing I saw that when adding urls from mpv page the `source:` url will be the /mpv/ page rather than /g/ which I would consider the canonical source url and also if an archive already with a /g/ source url the urlfinder plugin won't find it from a /mpv/ page so it would prompt another download but then fail (or needlessly replace the duplicate archive depending on lanraragi settings). I'd like to fix those two issues first.

Issue #1226, opened 2025-04-07
[search] multi-category logical AND toggling


psilabs-dev:
> I can't believe I'm writing all this on a feature about being able to click more buttons.

This is an interesting, albeit somewhat involved feature in the backlog. Partly came out of personal needs, but also out of looking at current issues. I would be interested (if no one else is :-)) though not now. Right now category toggles support only 1 category and 2 states (disabled and enabled). I'd like to extend this to multiple categories with 3 states: disabled, AND, and NOT. OR is not included for reasons I'll explain later. This may be broken to two parts, this is because the NOT part is harder as it involves logical OR as well. So we might consider ditching NOT and just implementing AND.

- [ ] support multiple categories with 2-states (disabled, AND). This is easy and probably more practical.
- [ ] support multiple categories with 3-states (disabled, AND, NOT). This would require implementing https://github.com/Difegue/LANraragi/issues/754.

This would help users make more complex compositions, while maintaining UI simplicity offered by the commutativity of the AND/NOT operators (by noting that OR does not commute with AND); however, it does not deny the use of OR within the search API logic; basically, this means we don't need parentheses to make searches when using category toggling.

Some concrete examples. To simplify things we'll use boolean notation with predicates. These examples are all dynamic categories; statics are somewhat easier to work with (fingers crossed).

**Example (AND composition)**: suppose category `A` is dynamic with predicate `series:A`, and category `B` is dynamic with predicate `language:english`. Then `A && B` is an anonymous category with predicate `series:A && language:english`, and `A && (!B)` has predicate `series:A && !(language:english)`. However, the user may not perform an OR search, like this: `A || B`, since a composition like `A || B && C` is ambiguous.

This by itself is pretty powerful. I can filter by an artist, then say: "actually I want to also filter to english only", and instead of writing the query, I happen to have an en filter category which can be applied at any time with a single click.

**Example (absorption)**: suppose category `A` is dynamic with predicate `series:A`, and category `B` is dynamic with predicate `series:A && language:english`. Then `A && B = B`, and `(!A) && B = false` induces the empty category.

**A more complex example (DNFs in the wild)**: just to give an example where AND+NOT induce OR. Suppose categories `A`, `B`, `C` are dynamic where `A` has predicate `a && b && c` and `B` has predicate `b && c && d`. Then
[...]

PR #1438, opened 2026-01-08
Tankoubon search and view

alethiophile:
This is one of three PRs I'll be submitting for the tankoubon UI.

Things I've done in this PR:

1. Added a UI flag to enable the grouptanks parameter on search
2. Fixed result display for tankoubons to give them thumbnails and correct links
3. Added a tankoubon view page that allows viewing and editing a particular tankoubon
4. Fixed various backend bugs in the tankoubon support
5. Updated tests for more comprehensive coverage

Things that remain to be done in this PR:

- Tweak UI on the tankoubon viewer page some more
- Make a "view all tankoubons" page
- Possibly add more editing options on the tankoubon viewer page, e.g. removing archives from a tankoubon or adding new ones

The two future PRs are an update to the search UI to enable bulk operations (which is currently the only way of adding archives to a tankoubon) and updates to the reader to support tankoubons.


alethiophile:
This is the ready-for-review version of the tankoubon search and view changes.

A UI is added to view and edit tankoubons, allowing metadata editing, reordering, and removal of archives. Tankoubons can be added to categories. A flag is added to enable or disable tankoubon grouping in searches; when enabled, searches may have tankoubon results, and these can be manipulated in the usual ways.

To allow viewing all tankoubons, I added a new pseudo-category to go with new and untagged, that just filters for tankoubons only. This works alongside all the other filter options. In order to enable more filtering while viewing tankoubons, I allowed searches to have multiple categories or pseudo-categories, which are all intersected for the final result. I'm less sure of the UI here.

Currently there's no way to create new tankoubons or add archives to a tankoubon. My UI for this has bigger changes to the search page, so will be in a separate PR.

lib/LANraragi/Controller/Api/Tankoubon.pm
```diff
@@ -199,5 +200,31 @@ sub get_tankoubons_file {
     );
 }
 
+# Find which categories this tankoubon is saved in.
```
Guerra24: This does basically the same as the archive endpoint. If we want tanks to work just like archives it would be better to have it work with the existing endpoint.

Guerra24 (changes requested):
Docs need to be updated with all the API and behavior changes.

I will to be testing this PR against the clients.

lib/LANraragi/Controller/Api/Tankoubon.pm
alethiophile: yeah, no need for this

alethiophile:
The OpenAPI documentation has been updated for these API changes. As best I can tell, everything should be backward-compatible (unless maybe if there are clients that will crash on the new thumb_archive field in archive/tankoubon metadata).

As regards the UI documentation, I can write it up, but with only the changes for this PR it'll be incomplete (since the new bulk selection system is the only way to add archives to a tankoubon). I'm now wondering if it's best to fold that PR into this one too.

lib/LANraragi/Controller/Api/Category.pm
```diff
@@ -6,8 +6,20 @@ use Encode;
 
 use LANraragi::Model::Category;
 use LANraragi::Model::Config;
+use LANraragi::Model::Tankoubon;
 use LANraragi::Utils::Generic qw(render_api_response exec_with_lock);
 
+# Helper to get title for archive OR tankoubon
+sub _get_item_title {
```
Difegue, line 13:
```suggestion
sub get_item_title {
```
If the intent is to denote this is a sub that isn't tied to an API endpoint, I think it'd be better to move it to one of the Utils:: subpackages.

lib/LANraragi/Controller/Api/Search.pm
```diff
@@ -30,34 +32,46 @@ sub handle_datatables ($self) {
...
-                $categoryfilter = "";
-            }
-
-            if ( $categoryfilter eq "UNTAGGED_ONLY" ) {
-                $untaggedfilter = 1;
-                $categoryfilter = "";
+            my $raw_filter = $req->param("columns[$i][search][value]") // "";
+
+            # Parse comma-separated category filters
+            # Pseudo-categories (NEW_ONLY, UNTAGGED_ONLY, TANKOUBONS_ONLY) become boolean flags
```
Difegue, line 44: I'm not certain we want to allow for multiple categories at once in a search. (pending rest of review while i read)

lib/LANraragi/Controller/Api/Search.pm
```diff
@@ -67,22 +81,49 @@ sub handle_api ($self) {
...
-    my $newfilter  = $req->param('newonly')       || "false";
-    my $untaggedf  = $req->param('untaggedonly')  || "false";
-    my $grouptanks = $req->param('groupby_tanks') || "false";
+    my $filter    = $req->param('filter');
+    my $category  = $req->param('category') || "";
+    my $start     = $req->param('start')    || 0;
+    my $sortkey   = $req->param('sortby');
+    my $sortorder = $req->param('order');
+
+    my ( $newfilter, $err ) = parse_bool( $req->param('newonly'), 'newonly' );
```
Difegue, line 90: This is pretty verbose, I think we could rework parse_bool to `die` on error and wrap all those param getters into a single try/catch block that does render_api_response on exception.

lib/LANraragi/Controller/Api/Search.pm
```diff
@@ -112,18 +153,45 @@ sub get_random_archives ($self) {
...
     my $filter       = $req->param('filter');
-    my $category     = $req->param('category')      || "";
-    my $newfilter    = $req->param('newonly')       || "false";
-    my $untaggedf    = $req->param('untaggedonly')  || "false";
-    my $grouptanks   = $req->param('groupby_tanks') || "false";
-    my $random_count = $req->param('count')         || 5;
+    my $category     = $req->param('category') || "";
+    my $random_count = $req->param('count')    || 5;
+
+    my ( $newfilter, $err ) = parse_bool( $req->param('newonly'), 'newonly' );
```
Difegue, line 159: ditto

lib/LANraragi/Controller/Api/Tankoubon.pm
```diff
@@ -199,5 +199,7 @@ sub get_tankoubons_file {
     );
 }
 
+
```
Difegue, line 194: leftover whitespace

lib/LANraragi/Model/Category.pm
```diff
@@ -186,9 +186,14 @@ sub delete_category {
 }
 
 # add_to_category(categoryid, arcid)
-#   Adds the given archive ID to the given category.
+#   Adds the given archive ID (or tankoubon ID) to the given category.
 #   Only valid if the category is Static.
 #   Returns 1 on success, 0 on failure alongside an error message.
+#
+#   NOTE: When viewing a category with tank grouping enabled, archives that belong
```
Difegue, line 193: Should this be in the docs? I guess there aren't very detailed tankoubon docs just yet anyway.

Guerra24:
`/api/tankoubons` does not include summary or tags, `/api/archives/{id}/metadata` does include them.

`Argument "lastreadtime" isn't numeric in subroutine entry at /lib/LANraragi/Utils/Database.pm line 290.`

I think `thumb_archive` should be automatically handled server-side through the thumbnails APIs rather than relying on the client.

Another suggestion I have is adding the option to include tanks as part of the archive list(`/api/archives`), maybe a get parameter or something.

lib/LANraragi/Controller/Api/Search.pm
Difegue: Doubly so if https://github.com/Difegue/LANraragi/issues/1226 becomes a thing actually, it'd probably conflict with this.

lib/LANraragi/Model/Search.pm
```diff
@@ -429,82 +462,107 @@ sub sort_results ( $sortkey, $sortorder, @filtered ) {
...
+                } else {
+                    # Convert the results into a hash table
+                    foreach my $item (@$data) {
+                        $tmpfilter{$item->[0]} = $item->[1];
+                    }
                 }
             }
         }
 
+        # Handle tanks separately - compute max lastreadtime from contained archives
```
Difegue, line 493: I don't think this should be handled at the search engine level -- We might as well add `lastreadtime` to the tank object itself and get that updated at the Reader level once that's implemented?  

Deviating from the Lua script will cause performance issues.

lib/LANraragi/Model/Search.pm
```diff
@@ -520,4 +578,15 @@ LUA
     return @sorted;
 }
 
+# Check if a tankoubon has any archive that exists in the given set.
+# Used for filters like "newonly" where we want tanks containing new archives.
+sub tank_has_archive_in_set ( $redis, $tank_id, $set_ref ) {
```
Difegue, line 583: Should this be in Model::Tankoubon instead?

lib/LANraragi/Controller/Api/Search.pm
Difegue: Looking at `parse_bool` again... Now that we merged in openAPI validation I'm not sure we actually want to make bool parsing **more** permissive.  

(especially adding stuff like on/off, I think that'd just encourage bad practice)

lib/LANraragi/Utils/Generic.pm
```diff
@@ -45,6 +45,35 @@ sub is_archive {
     return $_[0] =~ /^.+\.(?:zip|rar|7z|tar|tar\.gz|lzma|xz|cbz|cbr|cb7|cbt|pdf|epub|tar\.zst|zst)$/i;
 }
 
+# Parses a boolean parameter value from an API request.
+# Returns ($value, undef) on success, or (undef, $error_message) on invalid input.
+# Accepts: 1/0, true/false, yes/no, on/off (case-insensitive)
+# Empty or undefined values return the default (0 if not specified).
+sub parse_bool {
```
Difegue, line 52: As mentioned in the comment above, not sure this is a good idea

locales/template/en.po
```diff
@@ -1345,6 +1387,99 @@ msgstr "Writing a Predicate"
 msgid "Predicates follow the same syntax as searches in the Archive Index. Check the Documentation for more information."
 msgstr "Predicates follow the same syntax as searches in the Archive Index. Check the <a href=\"https://sugoi.gitbook.io/lanraragi/basic-operations/searching\">Documentation</a> for more information."
 
+msgid "Tankoubon"
+msgstr "Tankoubon"
+
+msgid "Error loading tankoubon"
```
Difegue, line 1393: Please make sure Tankoubon is capitalized every time -- It's a data model object like Archives and Categories, so I think that helps it make more sense for users.

lib/LANraragi/Controller/Category.pm
```diff
@@ -21,25 +22,37 @@ sub index {
...
-    $redis->quit();
+    # Build tankoubon list
+    my ( $total, $filtered, @tanklist_data ) = LANraragi::Model::Tankoubon::get_tankoubon_list(-1);
+    my $tanklist = "";
+
+    foreach my $tank (@tanklist_data) {
+        my $title = xml_escape( $tank->{name} );
+        my $id    = xml_escape( $tank->{id} );
+
+        $tanklist .=
```
Difegue, line 47: Would be nice to update this (and maybe the arclist equivalent too) to feature a small message if there are no tanks on the server  

Image

templates/index.html.tt2
```diff
@@ -138,10 +138,17 @@
...
 
-					<input id="thumbnail-crop" class="fa table-option" type="checkbox"
-						title="If enabled, thumbnails that don't fit a regular A4 page will be cropped to only show the left side.">
-					<label for="thumbnail-crop" style="vertical-align: middle; padding-top:6px">[% c.lh("Crop thumbnails") %]</label>
-				</div>
+				<input id="thumbnail-crop" class="fa table-option" type="checkbox"
+					title="If enabled, thumbnails that don't fit a regular A4 page will be cropped to only show the left side.">
+				<label for="thumbnail-crop" style="vertical-align: middle; padding-top:6px">[% c.lh("Crop thumbnails") %]</label>
+
+				[%# "Group by Tankoubon" is duplicated in both thumbnail-options and compact-options
```
Difegue, line 145: Not sure how i feel about perlstyle comments in the template files -- This feels self-explanatory enough to me that we could do without tbh

tools/openapi.yaml
```diff
@@ -1931,12 +1928,18 @@ paths:
           required: false
           default: false
           description: >-
-            Enable or disable Tankoubon grouping. Defaults to true.
+            Enable or disable Tankoubon grouping. Defaults to false.
```
Difegue, line 1931: I think the spec was wrong here, the intent is very much to have grouptanks to true by default for searches.

tools/openapi.yaml
```diff
@@ -2033,12 +2036,18 @@ paths:
           in: query
           required: false
           description: >-
-            Enable or disable Tankoubon grouping. Defaults to true.
+            Enable or disable Tankoubon grouping. Defaults to false.
```
Difegue, line 2039: ditto

tools/openapi.yaml
```diff
@@ -2892,26 +2913,26 @@ paths:
...
-      description: Adds the specified Archive ID (see Archive API) to the given Category.
+      summary: 🔑 Add an Archive or Tankoubon to a Category
+      description: Adds the specified Archive or Tankoubon ID to the given Category.
       tags:
         - categories
       parameters:
         - name: id
           in: path
-          description: Archive ID to add.
+          description: Category ID to add to.
```
Difegue, line 2923: woops, good catch

public/css/lrr.css
```diff
@@ -379,12 +379,12 @@ li {
   height: 480px;
   overflow: auto;
   text-align: left;
-  margin-left: -30px;
+  margin-left: 10px;
```
Difegue, line 382: Doesn't this break the batch tagging page? It uses the same class

public/js/common.js
```diff
@@ -297,6 +317,19 @@ LRR.buildBookmarkIconElement = function (id, bookmark_class) {
     return `<i id="${id}" class="${bookmarkClass} ${bookmark_class}${disabledClass}" title="${I18N.ToggleBookmark}" ${style}></i>`;
 };
 
+/**
+ * Get the appropriate view URL for an archive or tankoubon.
+ * Tankoubons link to /tankoubon?id=..., archives link to /reader?id=...
+ * @param {string} id The archive or tankoubon ID
+ * @returns {LRR.apiURL} The appropriate URL
+ */
+LRR.getItemViewURL = function (id) {
```
Difegue, line 326: I think this is unnecessary - clicking a tank should take you directly to the Reader with that Tank's set of Archives loaded.  

(Reader work can and should be in a separate PR for now)

public/js/common.js
```diff
@@ -4,6 +4,9 @@
  */
 const LRR = {};
 
+LRR.isProgressLocal = true;          // Whether to use local (localStorage) progress tracking
```
Difegue, line 7: No issues moving those consts out of Index, but they shouldn't be duplicated.

public/js/index.js
```diff
@@ -38,6 +39,17 @@ Index.initializeAll = function () {
     $(document).on("keydown.quick-search", Index.handleQuickSearch);
     $(document).on("keydown.escape-overlay", Index.handleEscapeKey);
 
+    // Selection bar and tank selector event handlers
+    $(document).on("change.archive-checkbox", ".archive-checkbox", function () {
```
Difegue, line 43: It looks like `archive-checkbox` class inputs aren't actually created anywhere in this PR? Am I missing something?

templates/index.html.tt2
```diff
@@ -218,31 +230,53 @@
...
+					if (!isTankoubon) {
+						items["download"] = { name: "[% c.lh('Download') %]", icon: "fas fa-save" };
+					}
+
+					items["copy link"] = { name: "[% c.lh('Copy link') %]", icon: "fas fa-link" };
+
+					[% IF userlogged %]
+					items["sep1"] = "---------";
+
+					// Edit Metadata only for archives
```
Difegue, line 250: Edit should lead to the tank view imo

public/js/index.js
```diff
@@ -290,6 +302,251 @@ Index.toggleCrop = function () {
     IndexTable.dataTable.draw();
 };
 
+// Handler for .group-tanks checkboxes (duplicated in thumbnail-options and compact-options).
+// Uses `this.checked` to read from whichever checkbox triggered the event.
+// Both checkboxes are synced via localStorage in updateTableControls() after each table draw.
+Index.toggleGroupTanks = function () {
```
Difegue, line 308: I wonder if the selection bits shouldn't be in a separate JS file like index_selection.js just to try and make it easier to work with; With this PR index.js gets to 1000 lines and it's a lot..

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -94,16 +97,18 @@ sub create_tankoubon ( $name, $tank_id ) {
     my $tank_title = redis_encode($name);
 
     # Add the tank name to LRR_TITLES so it shows up in tagless searches when tank grouping is enabled.
-    $redis_search->zadd( "LRR_TITLES", 0, "$tank_title\0$tank_id" );
+    # Title must be lowercased to match how search queries are processed.
+    my $tank_title_lower = lc($name);
```
Difegue, line 101: Shouldn't this be 
```suggestion
    my $tank_title_lower = lc($tank_title);
```
?

lib/LANraragi/Model/Tankoubon.pm
```diff
@@ -9,12 +9,15 @@ use utf8;
 
 use Redis;
 use Mojo::JSON qw(decode_json encode_json);
-use List::Util qw(min);
+use List::Util      qw(min);
+use List::MoreUtils qw(uniq);
 
-use LANraragi::Utils::Database qw(invalidate_cache get_archive_json_multi get_tankoubons_by_file);
+use LANraragi::Utils::Database;
```
Difegue, line 15: Please keep the qws so you don't need to invoke the full package name in the subs every time

Difegue (changes requested):
Sorry for the delay in reviewing this, I blame the fact it is quite chunky 😤    
I think there are a few parts of this that should probably be segmented to separate PRs; I know this is already technically part 1 of 3, but even then there are multiple features mixed in this that all need specific attention... 

Thanks a lot for the extended Tank testing and doc updates, very appreciated!  

I left multiple comments but as a tl;dr:  
- The Tank category and matching tanksonly filter should probably be removed for now to keep this simple and less conflicty; I personally think tanks should be integrated as seamlessly as possible in the main UI, so the grouptanks bool is as far as i'd like to take it for now.
- The multi-category AND implementation should probably be in a separate PR: I'm also aware this might conflict with work from @psilabs-dev for #1226 
- While we need a dedicated view to edit tanks, opening them should directly go to Reader, not an inbetween page. (It's fine for now if Reader doesn't support Tanks)
- Said inbetween page should be used only for Tank edition (order/removal of IDs + additional metadata)
- The Index changes for selection don't seem to work at all for me so I can't really test any of it; I'm not sure if this is working better in PR https://github.com/Difegue/LANraragi/pull/1447 or not? 

There are parts I haven't looked at in detail because I ran out of 🧠, so I'll likely have more comments once this is cleaned up a bit further.

lib/LANraragi/Model/Tankoubon.pm
Difegue: nvm, regular titles are also saved as their decoded variant

Difegue:
I've started to push piecemeal bits of this PR to dev (with appropriate co-authoring), since the API bugfixes and test updates are good to go imo.  

I'll likely use the sortableJS bits for a Tank edit view as well afterwards, so I'll keep the PR open as reference for now.

Difegue:
I used sortableJS for the Tank Edit bits, but for the time being I don't think we need a dedicated Tankoubon view at all as Edit/Reader/Index can handle it all.  

Don't think there's anything left to use from this PR so I'll close it.

PR #1439, opened 2026-01-10
Feat: implment language option

### LINKED ISSUE #1278: Option to set the language.
Hello, I think there should be an option to manually set the language instead of relying on the browser headers, I think using the headers is great for a public instance, but I use mine in "no fun mode" and want my instance to be in English no matter what.

I think the best option for all would be an option in setting with all languages supported and "auto", but for me at least, a docker environment variable would be more than enough, for example "FORCE_LANGUAGE=enEN"

gustaavv:
I implmented this FR issue #1278 .

I added a new item in the config page. To keep the styles of the existing inputs and this new select the same, some styles were updated.

One problem is that I can not judge whether the language names are correct since they are AI-generated, e.g. whether "Français" in French is the meaning of "French" in English. It seems OK when I searched these names on Google.

In addition, I updated related documents.

One future improvements may be using some library to map the language codes to their native language names and those in English. Also, the languages available can be fetched from weblate API so that we do not need to update the code when a new language is available.

### DIFF
--- a/lib/LANraragi/Utils/I18NInitializer.pm
+++ b/lib/LANraragi/Utils/I18NInitializer.pm
@@ -9,14 +9,20 @@ use LANraragi::Utils::I18N;
 sub initialize {
     my ($app) = @_;
 
+    # Load forced language setting at startup to avoid querying Redis on every request
+    my $forced_language = $app->LRR_CONF->get_language;
+    $app->defaults( forced_language => $forced_language );
+    
+    $app->LRR_LOGGER->debug("Forced language setting: $forced_language");
+
     $app->helper(
         lh => sub {
             my ( $c, $key, @args ) = @_;
             
             my @langs = ();
             
-            # Check if there's a forced language setting
-            my $forced_language = $c->LRR_CONF->get_language;
+            # Check if there's a forced language setting (loaded at startup)
+            my $forced_language = $c->stash('forced_language');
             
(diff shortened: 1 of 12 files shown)


lib/LANraragi/Utils/I18NInitializer.pm
```diff
@@ -12,23 +12,34 @@ sub initialize {
...
-                # Split the header by commas to get the languages
-                @langs = split /,/, $accept_language;
-
-                # For each language, remove the quality value (aka ;q=0.9)
-                # Browsers usually order the languages by quality already do we don't really need them
-                foreach my $lang (@langs) {
-                    $lang =~ s/;.*//;    # remove any quality value
+            
+            # Check if there's a forced language setting
+            my $forced_language = $c->LRR_CONF->get_language;
```
psilabs-dev: With this change, every `lh` call will create a redis connection, plus languages don't change frequently, better if we can cache this

Difegue: I'd argue we could just specify that this parameter needs a server restart in the config page and load the forced language option at server startup in `I18NInitializer::initialize`. Less complex than setting up and updating a cache

Difegue (changes requested):
Looks good, but let's avoid a redis hit on every `lh` call by just looking at the language setting on server startup.

Difegue (approved):
:shipit:

PR #1440, opened 2026-01-11
Adjust handlePaste so that it can append the pasted text to an existing pending tag

steve1316:
## Description
- This PR enhances `Edit.handlePaste` so it appends the pasted text to a pending tag before combining them together into a new tag.
- Currently when you paste text into the `Tags` text box with no pending tag in it, it submits it as a new tag. If it does have a pending tag like "artist:" and you paste in text like "Hirohiko Araki", it creates "Hirohiko Araki" as a new tag and the "artist:" is still pending.

https://github.com/user-attachments/assets/e812d2d6-6ce1-4fd1-ae31-a26d60cb1213

https://github.com/user-attachments/assets/ed5062d6-af25-4591-8ae9-d1df59a3e221

### DIFF
--- a/public/js/edit.js
+++ b/public/js/edit.js
@@ -77,11 +77,28 @@ Edit.handlePaste = function (event) {
     event.stopPropagation();
     event.preventDefault();
 
+    // Get the pending text already typed in the input field
+    const inputElement = $(".tagger-new").children()[0];
+    const pendingText = inputElement ? inputElement.textContent : "";
+
     // Get pasted data via clipboard API
     const pastedData = event.originalEvent.clipboardData.getData("Text");
 
     if (pastedData !== "") {
-        pastedData.split(/,\s?/).forEach((tag) => {
+        // Split by comma to handle multiple pasted tags
+        const tags = pastedData.split(/,\s?/);
+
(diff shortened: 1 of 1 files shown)


steve1316:
https://github.com/user-attachments/assets/51c915d7-fc29-4ecc-8529-bb50789d66f3

Multi-tag pasting still works via commas. Although I realize that if there is a pending tag in the text box, then it will append the first element to that pending tag.

In that case, would it make sense to revert back to the previous behavior where the pending tag would be kept as-is if you are pasting multiple tags at once?

Difegue:
I think it makes sense to have the same behavior for single and multi-tag pasting -- Users will expect that if they have something pending in the textbox, pasting will append to it in both cases.  

Thanks for this!

psilabs-dev:
What browser was this tested on? I wasn't able to reproduce in safari.

steve1316:
> What browser was this tested on? I wasn't able to reproduce in safari.

I am on Firefox.

Guerra24:
I can reproduce this issue after this pr got merged, on FF148 and FF140esr on Windows and Linux. #1443 fixes it. I don't think `textContent` really works for `input` elements.

PR #1447, opened 2026-01-16
Selection and batch operations in search

### LINKED ISSUE #826: Add a selection mode/mechanism to the main Index
Current UX for quickly working with a lot of archives is painful in a few ways, so this is a design draft to try and fix that.  
I'd already briefly mentioned it here: https://github.com/Difegue/LANraragi/issues/570#issuecomment-1550023455 , but it's more generic than just batch operations and doesn't require a heavy frontend refactor:  

* Current Index page should have a new toggle to enable "Multi-Select Mode" 
  * (or maybe make it one of the carousel dropdown options? I'm worried about discoverability if doing it that way tho)
* When this toggle is enabled, the topmost carousel is **emptied out**, and checkboxes appear next to each archive. 
* Search works as normal -- Ticking a checkbox **adds it to the carousel** for quick review of your selected archives, no matter what you're searching for.
* Once at least one file is selected, you can choose to run **Batch Operations** on the selection. 
  * (which will probably be a popup with an iframe to the current batch page or something because that's entirely too much complexity to move into the index)  
  * Some simpler options could eventually live here alongside Batch, like quick deletion, creating/adding to a category(or meta-archive when those come!), or IceBreeze's crazy poweruser batch syntax mentioned [here](https://github.com/Difegue/LANraragi/issues/570#issuecomment-1544131195)

alethiophile:
This is my UI for doing batch operations on archives from the search page. Currently, this is only used for adding archives to tankoubons, but I intend to add a few more operations on tags and categories.

Each result in a search has a checkbox. When checkboxes are selected, a bar pops up allowing you to do bulk operations on the selection. (Basic pattern inspired by similar operations in file managers.) The selection is maintained across page navigation. Because this is in the search result view, it gives you an easy way to operate on the results of a search: just run the search and select everything.

It's based against my tankoubon-search-and-view tree, since it's intended to work with that code, but is technically separable.


Difegue:
An equivalent has been implemented for #826 , so I'll close this PR for now.

PR #1448, opened 2026-01-18
Implement request/response validation with OpenAPI

psilabs-dev:
This PR aims to use a new dependency to enforce OpenAPI request/response validation based on a schema.

Currently (for the past couple months) this is in draft because:

- pending a migration from openapi.json to the existing yaml,
- I haven't figured out the auth with conditional reader progress update

I was also planning to open this once I have 100% API coverage, but that is... gonna take longer than I expect or like, and waiting for a feature like this to stabilize might not be realistic considering the incoming updates.

The patterns for implementing APIs are pretty straightforward, thanks to `Mojolicious::Plugin::OpenAPI`. I probably could've opened this PR sooner, but I'll open this now while it's still maintainable as a feature for feedback/visibility, and because there are several other PRs that affect the API.


tools/openapi.yaml
```diff
@@ -3011,33 +3026,9 @@ paths:
             type: string
       requestBody:
         required: true
-        description: Json with 2 optional keys "archives" and "metadata" 
+        description: Json with 2 optional keys "archives" and "metadata"
         content:
-          application/x-www-form-urlencoded:
```
psilabs-dev, line 3157: are we using `x-www-form-encoded` or `multipart/form-data` for anything with tanks?

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -341,9 +343,23 @@ sub update_metadata {
 }
 
 sub update_progress {
-    my $self = shift;
+    my $self = shift->openapi->valid_input or return;
     my $id   = check_id_parameter( $self, "update_progress" ) || return;
 
+    # Enforce authentication if authprogress is enabled
+    if ( LANraragi::Model::Config->enable_authprogress ) {
```
psilabs-dev, line 427: from a maintenance standpoint this isn't ideal, but I guess we're dealing with an unusual auth situation here.

lib/LANraragi/Controller/Api/Shinobu.pm
```diff
@@ -39,19 +39,19 @@ sub shinobu_status {
             Win32::Process::Open($shinobu, $pid, 0);
 
             $self->render(
-                json => {
+                openapi => {
                     operation => "shinobu_status",
                     success   => 1,
-                    is_alive  => $shinobu->GetProcessID() != 0,
-                    pid       => "" . $shinobu->GetProcessID()
```
psilabs-dev, line 46: @Guerra24 not sure if there was a reason why this is returned as a string from the windows PR, but I've updated it to return int instead (as that is the format for unix/openapi).

https://github.com/Difegue/LANraragi/pull/1266/changes#diff-558c92d4305c83a422faa6fcaa9e0450023529956fcbe4a6757260e5d4b11c9eR46

psilabs-dev:
No more changes planned (unless another openapi validation error happens) as the auth problem is now addressed, and auth/CORS tests are passing, though I might give cors/auth another pass just in case.

With this, in the future usage of `render_api_response` would remain same; usage of `render->` would switch from "json" to "openapi" ("text" remains same), and all controllers would use OpenAPI controllers instead, e.g.

```perl
sub serve_archivelist {
    my $self   = shift->openapi->valid_input or return;
    my @idlist = LANraragi::Model::Archive::generate_archive_list;
    $self->render( openapi => \@idlist );
}
```

Openapi will handle all "/api" routes and no more definitions will be placed in `Routing.pm`. From experimentation, the placement of cors/openapi/non-api routes do matter. Testing is done locally on docker/windows

lib/LANraragi/Controller/Api/Shinobu.pm
Guerra24: image

Passing the value as-is caused the int to become a float in the json (for whatever reason) and I didn't have time back then to figure out why. Apparently the fix is to cast the value like this `int($shinobu->GetProcessID())`.

image

If you could add the casting fix it would be amazing. If it is being properly converted back to an int with openapi then that should also work.

psilabs-dev: sure np

psilabs-dev:
Added an option to disable OpenAPI request and response validation, UI/doc updates pending, and I'll start doing some progressive dogfooding.

When validation is disabled, overrides will be performed for OpenAPI request/response validation. I don't plan to too thoroughly test this specific functionality in prod, though this implementation is supported by mutation integration testing.

Currently, this can be configured via two ways: `LRR_DISABLE_OPENAPI=1` environment variable, and via a redis configuration flag "disableopenapi=1". Validation is enabled by default. For now, developers can disable validation via environment variable or direct database edits.

lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -39,31 +41,31 @@ sub check_id_parameter {
...
     $redis->quit;
 
-    $self->render( json => \@untagged );
+    $self->render( openapi => \@untagged );
 }
 
 sub serve_metadata {
-    my $self  = shift;
+    my $self  = shift->openapi->valid_input or return;
     my $id    = check_id_parameter( $self, "metadata" ) || return;
```
Difegue: As a result of the validation `check_id_parameter` is basically unnecessary at this point since the ID will always be in the API path when it gets to this point.  

As a result I'm not sure if all the `No archive ID specified` 400s that are in the spec can even still happen? Feels unlikely to me.

lib/LANraragi/Controller/Api/Archive.pm
Difegue: Yeah this one is a bit jank and can't be described in the spec, I'd say it's OK for a first implementation.  

I guess we'd need two endpoints to do it properly 🤔

lib/LANraragi/Model/Config.pm
```diff
@@ -170,6 +170,15 @@ sub get_tagrules {
     );
 }
 
+sub get_disable_openapi {
+
+    if ( $ENV{LRR_DISABLE_OPENAPI} ) {
```
Difegue, line 183: This should probably be documented.

lib/LANraragi/Model/Config.pm
```diff
@@ -170,6 +170,15 @@ sub get_tagrules {
...
+        return 1;
+    }
+
+    return &get_redis_conf( "disableopenapi", "0" );
```
Difegue, line 187: I'm not _sure_ we want this as an exposed preference? It might be good for existing API clients that the validation would break.   
At the very least if we do expose it, it should have matching UI in the config template.

lib/LANraragi.pm
```diff
@@ -142,6 +142,10 @@ sub startup {
         $self->LRR_LOGGER->info("LANraragi $version started. (Production Mode)");
     }
 
+    # Route Mojolicious/plugin logs (including OpenAPI validation warnings)
+    # through LRR's rotating logger pipeline.
+    $self->log( get_logger( "Mojolicious", "lanraragi" ) );
```
Difegue: I think this should go under the `mojo` moniker so it remains coherent with the current log ordering system.  
```suggestion
    $self->log( get_logger( "Mojolicious", "mojo" ) );
```  

Otherwise currently you'll get the openAPI warnings in the "general" tab:  
Image

Doing this means we can probably remove the `$self->log` bit above too, that was a pre-RotatingLog hack to copy the Mojo logs to a dedicated file in Debug mode and I think with this line we now get that for free.

lib/LANraragi/Utils/Routing.pm
```diff
@@ -9,12 +9,55 @@ use Encode;
...
 sub apply_routes {
     my $self = shift;
 
+    # Initialize Mojolicious::Plugin::OpenAPI
+    # And the single "/search" API endpoint because datatables
+    my $api         = $self->routes;
+    my $search_api  = $self->routes;
+
+    # The API router outputs CORS headers if the user allows it in the settings.
+    if ( $self->LRR_CONF->enable_cors ) {
```
Difegue, line 27: Isn't it possible to do this with https://metacpan.org/pod/Mojolicious::Plugin::OpenAPI::Cors now? 

I guess we'd still need it just for the datatables endpoint regardless(😑) so not necessary

lib/LANraragi/Utils/Routing.pm
```diff
@@ -9,12 +9,55 @@ use Encode;
...
+
+        # Private API requests are non-simple due to the Authorization header, so browsers send a preflight request.
+        # Preflight requests are OPTIONS requests, which we need to support explicitly
+        $api        = $api->under('/')->to('login#setup_cors');
+        $search_api = $search_api->under('/')->to('login#setup_cors');
```
Difegue, line 32: Also isn't this missing the explicit OPTIONS setup that was there previously if we don't go with the plugin?
Might not be necessary anymore idk, although i'm guessing it might interact badly with the plugin hook afterwards  
```
$api->options(
            '/api/*' => sub {
                my $self = shift;
                $self->rendered(200);
            }
        );
```

lib/LANraragi/Utils/Routing.pm
Difegue: (although realistically i don't know if we care about the datatables endpoint not being CORS-friendly, it's not really meant to be used by clients anyway?)

lib/LANraragi/Utils/Routing.pm
```diff
@@ -23,35 +66,20 @@ sub apply_routes {
...
-                $self->rendered(200);
-            }
-        );
-    }
-
     # Routers for routes that require auth
-    my $logged_in     = $public_routes->under('/')->to('login#logged_in');
-    my $logged_in_api = $public_api->under('/')->to('login#logged_in_api');
+    my $logged_in               = $public_routes->under('/')->to('login#logged_in');
+    my $logged_in_search_api    = $search_api->under('/')->to('login#logged_in');
```
Difegue: That variant isn't actually used, right?

tools/openapi.yaml
```diff
@@ -449,6 +449,24 @@ paths:
           required: false
           schema:
             type: string
+      requestBody:
+        required: false
+        content:
+          application/x-www-form-urlencoded:
```
Difegue: What's using those endpoints (plugins/use and plugins/queue) with form-urlencoded? The webclient only uses the query variant iirc  

(i can see one of the commits mention risk but there's no description so 🤷 )

tools/openapi.yaml
Difegue: Well, there's no UI that actively uses the API at the moment so anything goes -- I think this was done to match the structure of the other endpoints.

Difegue (changes requested):
Yeah this is pretty banger, the additional validation/error messages I think will be quite helpful for API devs.  
(The messages end up being in a different structure than the existing error messages but well, you're not really meant to hit the validation errors with normal use so it's fine)  

It also pretty much forces me to keep the spec up to date so I'll take it - A few comments beforehand tho.   
Will probably need to ask around API clients too in order to verify this works fine

lib/LANraragi/Model/Config.pm
psilabs-dev: The UI wasn't added bc it was dev-only, but it's easy enough for me to add one.

I personally wouldn't use it tbh; mainly it's for users who're running into validation errors due to bad server-side data and (hopefully) want to figure out what's going on (and don't know what environment variables are or how to pass them).

lib/LANraragi/Controller/Api/Archive.pm
psilabs-dev: > As a result I'm not sure if all the No archive ID specified 400s that are in the spec can even still happen? Feels unlikely to me.

yeah pretty much nope

lib/LANraragi/Utils/Routing.pm
psilabs-dev: The plugin auto-handles the OPTIONS routes, via SpecRenderer (which registers an options route for every openapi path) and Security (which skips auth for said generated routes), basically doing the "->options" job we're doing here.

lib/LANraragi/Utils/Routing.pm
psilabs-dev: It is needed; it can't be in openapi.yaml because it's "/search" and also uses session cookie auth, which isn't in the validation logic. It's also the only API endpoint which doesn't survive a migration into OpenAPI, hence the singling out.

tools/openapi.yaml
psilabs-dev: Yeah... I think we should remove this.

The idea (at the same time around disabling validation) was that LRR with OpenAPI validation should not introduce behavioral regressions compared to LRR before validation, otherwise users may experience broken systems (due to prior LRR permissiveness).

For example, previously form-urlencoded *was* allowed by LRR (or not forbidden), and clients (like mine before) were allowed to send requests in various unofficial formats and mojo would just handle it under the hood.

But if the goal is to minimize behavior changes and accomodate for previous LRR permissiveness, then we'd have a looot more schema changes than just this to make.

lib/LANraragi/Utils/Routing.pm
psilabs-dev: err we could? Though we just do "*" cors wildcard, not sure what value the plugin brings here

tools/openapi.yaml
psilabs-dev: json it is then 🏃

lib/LANraragi/Model/Config.pm
psilabs-dev: Added in architecture.md

lib/LANraragi/Model/Config.pm
psilabs-dev: Added UI in security but could be put anywhere else

tools/openapi.yaml
```diff
@@ -3585,7 +3585,6 @@ components:
         - pagecount
         - lastreadtime
         - size
-        - toc
```
Difegue, line 3551: Tanks don't have ToC but regular archives will.  
I'd say this means we need two descriptions in the schema, but the search API by design can mix tanks and regular archives, so the objects need to match, even if some values are unused.

lib/LANraragi/Utils/Routing.pm
Difegue: Yeah not really necessary tbh, I guess it'd be two less lines but that'd be about it

tools/openapi.yaml
Difegue: I think if we show an explicit toggle to disable validation in settings, it's acceptable to be less permissive as a result. I'd say yeet it™️

locales/template/en.po
```diff
@@ -455,6 +455,15 @@ msgstr "Have API requests support Cross-Origin Resource Sharing, which allows we
 msgid "Turn this on if you want to access this service through a web-based wrapper (e.g. a userscript) used/hosted on another domain."
 msgstr "Turn this on if you want to access this service through a web-based wrapper (e.g. a userscript) used/hosted on another domain."
 
+msgid "Disable OpenAPI Validation"
+msgstr "Disable OpenAPI Validation"
+
+msgid "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors."
+msgstr "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors."
+
+msgid "Fully effective after restarting LANraragi."
```
Difegue: This string is already in the po file, no need to repeat it

locales/template/en.po
```diff
@@ -455,6 +455,15 @@ msgstr "Have API requests support Cross-Origin Resource Sharing, which allows we
 msgid "Turn this on if you want to access this service through a web-based wrapper (e.g. a userscript) used/hosted on another domain."
 msgstr "Turn this on if you want to access this service through a web-based wrapper (e.g. a userscript) used/hosted on another domain."
 
+msgid "Disable OpenAPI Validation"
+msgstr "Disable OpenAPI Validation"
+
+msgid "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors."
+msgstr "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors."
```
Difegue:
```suggestion
msgid "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors with old API clients."
msgstr "Disables OpenAPI API schema validation. Turn this on if you are running into validation-related errors with old API clients."
```

Difegue (changes requested):
Good stuff, just a few nits left 👍

lib/LANraragi/Utils/Routing.pm
Difegue: Well yes, but this specific object (`logged_in_search_api`) is only used with `enable_nofun`, and you already have this above:  

```
if ( $self->LRR_CONF->enable_nofun ) {
        $api        = $api->under('/')->to('login#logged_in_api');
        $search_api = $search_api->under('/')->to('login#logged_in_api');
    }  
```

The difference is in `login#logged_in_api` vs `login#logged_in`, but `logged_in_api` also accepts the session cookie, so I think this is redundant. Am I missing something? 🤔

tools/openapi.yaml
psilabs-dev: > I'd say this means we need two descriptions in the schema

That's what I'm guessing from the API, so we just make toc not required so it's compatible for arcs and tanks. Or are you considering making it required (and adding an empty toc field in LRR API from controller-side)?

lib/LANraragi/Utils/Routing.pm
psilabs-dev: Hmm, I don't remember the reason, just that before integration tests failed when I made this change. But maybe I am misremembering that part too.

We can just merge these two together for now and do some live testing (it's probably fine anyways)

locales/template/en.po
psilabs-dev: thanks, rm-ed

psilabs-dev:
(Whew, almost let this through)

Because `logged_in_api` is (and takes precedence) before the OpenAPI stuff, we do need to handle OPTIONS case (by copy/pasting the Security.pm bypass)

tools/openapi.yaml
Difegue: Right, I hadn't considered we could just make the field optional and have easy-ish compatibility.  
Might as well keep it like that for now yeah, eventually we could just have empty ToCs in tanks if we want the objects to be 100% identical.

Difegue (approved):
:shipit:

PR #1449, opened 2026-01-18
feat: Add archive & reader source links; support fractional auto page-turn intervals

ll2963336:
1. Reader Experience Upgrade: Increased the auto-page turn timer precision from 1s to 0.1s. It now supports decimal intervals with a real-time, single-decimal countdown display, making the transition smoother and more accurate.

2. Source Tracking: Added support for parsing the source: tag. When an archive contains a source URL, a visible external link icon will now appear on both the Index Thumbnails and the Reader Sidebar, allowing for one-click navigation to the original source.

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -831,6 +831,26 @@ div.fullscreen img {
   opacity: 0.6;
 }
 
+.thumbnail-source-icon {
+  cursor: pointer;
+  position: absolute;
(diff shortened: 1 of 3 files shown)


Difegue:
Sorry, I don't see the UX interest in adding additional source tag links besides the ones that already exist when you click the tag in the Reader overview.

PR #1450, opened 2026-01-21
Update integration tests

psilabs-dev:
- compare: https://github.com/psilabs-dev/aio-lanraragi/compare/e50968c19f50b718044335836c98ddfe99841d3f...4c5ffe9d09042023be65807bc288eda41fd91e59
- PR: https://github.com/psilabs-dev/aio-lanraragi/pull/163

No new tests planned atm, instead I've stabilized the existing test cases and make conditions stricter.

- Redis/docker connections will be explicitly closed on teardown when applicable.

The following updates apply to all Playwright UI tests:

- All tests now explicitly create a browser context, and explicitly close context/browser within the scope of a test case.
- All tests assert no errors when checking browser network inspector. This means that a UI test case will fail if the network inspector finds any LRR-side status code that isn't sub-400 or 401. Non-LRR origin URLs are excluded (e.g. Github rate limit errors will not trigger test case failure). This is a stricter client-side condition than expecting no server-side error logs, as 4xx validation errors and silent, server-side error responses will fail the test case. This will address a subset of upcoming OpenAPI integration testing failures, including the missing "/search" endpoint issue that escaped testing from earlier.

Log health assertions are occasionally capturing Shinobu error logs ("giving up adding it to the filesystem"); this is an existing race condition which can be ignored, or we can suppress it in the future test-side or LRR-side.

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -10,11 +10,11 @@ on:
       integration_test_ref:
         description: 'Integration testing repository git commit hash or branch'
         required: true
-        default: '4c5ffe9d09042023be65807bc288eda41fd91e59'
+        default: '78fe5fff8926b295120e764d0bd335b9bf47c440'
 
 env:
   INTEGRATION_TEST_REPOSITORY: ${{ github.event.inputs.integration_test_repository || 'psilabs-dev/aio-lanraragi' }}
-  INTEGRATION_TEST_REF: ${{ github.event.inputs.integration_test_ref || '4c5ffe9d09042023be65807bc288eda41fd91e59' }}
+  INTEGRATION_TEST_REF: ${{ github.event.inputs.integration_test_ref || '78fe5fff8926b295120e764d0bd335b9bf47c440' }}
 
 name: "Continuous Integration \U0001F44C\U0001F440"
 jobs:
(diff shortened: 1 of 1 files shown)


psilabs-dev:
- Compare: https://github.com/psilabs-dev/aio-lanraragi/compare/4c5ffe9d09042023be65807bc288eda41fd91e59..78fe5fff8926b295120e764d0bd335b9bf47c440
- PRs: https://github.com/psilabs-dev/aio-lanraragi/pull/167, https://github.com/psilabs-dev/aio-lanraragi/pull/168

Added a logic fix: previous docker deployment integration test used essentially vanilla redis.conf configuration; this is now fixed so that we will load the LRR redis config into the redis container instead.
- In the scenario that we are testing a docker image: we will actually create an ephemeral container, pull the conf file out and use it to operate our test-time redis instance.

Added client-side optimizations. Mainly moved everything (including all of my other Python projects) over to uv; it (by it I mean my build times) is generally 20-30% faster. It's been pretty good.

Another client-side optimization is the request-handling, previously validation was unnecessarily done in Python, now Pydantic (with rust core) handles it, which also made request handling faster (though it's not a big improvement considering it's not a bottleneck in testing).

And some windows test-side improvements, such as not spawning console windows, which (heuristically) have alleviated the test pressure problems?

Unrelated: tests can be run with a different dockerfile now, if you ever need that for some fork-side reason.

psilabs-dev:
- Compare: https://github.com/psilabs-dev/aio-lanraragi/compare/78fe5fff8926b295120e764d0bd335b9bf47c440..dfdcd948cbc6d6bbd2db6620d8499bc8698843e3
- PRs: https://github.com/psilabs-dev/aio-lanraragi/pull/171, https://github.com/psilabs-dev/aio-lanraragi/pull/176

Thanks to some clankRCA™, instrumentation-spamming, hypothesis testing and process of elimination, the cause of Windows test-side failures and slowness were (practically) identified; see https://github.com/psilabs-dev/aio-lanraragi/pull/173 and https://github.com/psilabs-dev/aio-lanraragi/pull/172.

But *anyways*, the tests (ignoring Shinobu) should be more stable and faster now.

Also added/extended test coverage:

- `test_category` includes a regression behavior check against OpenAPI URL encoding
- added a double page navigation behavioral test case
- added a reader-mode resource handler management test case
- `test_tankoubon` includes similar regression check as `test_category`
- added two local progress-related test cases, previously uncovered
- added a basic plugin functionality test case
- a plugin not available test case
- an invalid plugin type test case (OpenAPI regression)

Added various other tests which don't run by default (OpenAPI feature + bypass, preloading check, etc.)

PR #1452, opened 2026-01-23
Fix some basic bugs

psilabs-dev:
An upstream/downstream diff-based code review identified some issues. This PR fixes the issues raised by review which felt obvious.


Difegue:
yep all pretty obvious, thanks for that.

PR #1453, opened 2026-01-23
ToC related fixes

psilabs-dev:
Addresses 2 issues raised during code review, which I'll explain below.

Issue 1: `add_toc_entry`

If an exception happens, (supposedly judging from the code) a basic toc is still passed into Redis. However, an empty hash gets passed into the database, instead of `{"page": $title}`. Also, otherwise the previous ops on toc wouldn't make much sense.

There's another small issue outside of review. Suppose redis drops while in the try block, then hset will fail and put the process into the catch block. If redis comes back up during the catch block, then LRR will have erased the toc entry due to the database being unstable.

Issue 2: `build_json`

Since eval code already exists, then we might as well just account for the cases when eval fails.

Eval failure already means that none of the subsequent code should be executed. However, even if decode passes, we should probably check that we have a hash instead of `[]`.

### DIFF
--- a/lib/LANraragi/Controller/Api/Archive.pm
+++ b/lib/LANraragi/Controller/Api/Archive.pm
@@ -379,6 +379,10 @@ sub add_toc {
     my $page  = $self->req->param('page');
     my $title = $self->req->param('title');
 
+    unless ( defined $page && defined $title ) {
+        return render_api_response( $self, "add_toc", "Missing page and/or title." );
+    }
+
     return unless exec_with_lock(
         $self,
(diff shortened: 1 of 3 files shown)


Difegue (approved):
yeah, those fixes make sense. 
I'm noting the integration test failure but i'm assuming it's because a change is needed on those.

psilabs-dev:
yeah there are three issues that need fixes, one is a shinobu lrr-side race condition (covered by redis lock coverage), then there are two test-side infrastructure problems which make windows testing unstable (coming up)

PR #1455, opened 2026-01-23
Extend Redis lock coverage

### LINKED ISSUE #1425: add_to_filemap race condition
Caught by a test.

An archive can exist here: https://github.com/Difegue/LANraragi/blob/d79dd8f75262739951976937fdbc81ba17bcf919/lib/Shinobu.pm#L186

But not here: https://github.com/Difegue/LANraragi/blob/d79dd8f75262739951976937fdbc81ba17bcf919/lib/Shinobu.pm#L196

triggering a die.

More generally, there's no guarantee of file existence made to `compute_id` (cnt loop only sleeps at most 5s, which can lead file to be deleted elsewhere). We should probably ~~demote the error logs to warning here~~, or lock the resource and ensure shinobu can read the file without a delete API interfering.

psilabs-dev:
close https://github.com/Difegue/LANraragi/issues/1425

`add_to_filemap` is triggered on file events, which can frequently race against active file changes resulting in missing files during ID computation. If this happens, we don't really have an "error"-severity log.

This is one of the main invalid integration test case failure reasons, so tightening this condition will get rid of some noise. Or we can keep the logs if a file doesn't exist and demote that case to warning.

### DIFF
--- a/lib/LANraragi/Utils/Generic.pm
+++ b/lib/LANraragi/Utils/Generic.pm
@@ -361,11 +361,11 @@ sub exec_with_lock {
 # Redis connection is optional; if not supplied, a managed connection will be created.
 sub exec_with_lock_pure {
 
-    my $lock_names = shift;
-    my $func       = shift;
-    my $redis      = shift;
-    my $ttl        = shift // 10;
-    my $own_redis = 0;
+    my $lock_names  = shift;
+    my $func        = shift;
+    my $redis       = shift;
+    my $ttl         = shift // 10;
+    my $own_redis   = 0;
 
     die "Lock name list cannot be empty" unless scalar(@$lock_names);
 
@@ -390,7 +390,12 @@ LUA
     my @lock_name_stack = ();
     foreach my $lock_name ( @$lock_names ) {
         my $token = sha256_hex( $lock_name . ":" . $$ . ":" . time() . ":" . rand() );
-        my $lock  = $redis->set( $lock_name, $token, 'NX', 'EX', $ttl );
+        my $lock = eval { $redis->set( $lock_name, $token, 'NX', 'EX', $ttl ) };
+        if ( my $acquire_error = $@ ) {
+            # If a lock acquisition failure happens, then a problem has occurred with Redis.
+            get_logger( "Concurrency", "lanraragi" )->error("Failed to acquire lock $lock_name: $acquire_error");
+            last;
+        }
         last unless $lock;
         push( @lock_name_stack, [ $lock_name, $token ] );
     }
@@ -399,9 +404,16 @@ LUA
     unless ( scalar(@lock_name_stack) == scalar(@$lock_names) ) {
         while ( @lock_name_stack ) {
             my ( $lock_name, $token ) = @{ pop(@lock_name_stack) };
-            $redis->eval( $release_lua, 1, $lock_name, $token );
+
+            # This should be best effort release, but if failure happens, 
+            # then continue releasing remaining locks and let TTL take over.
+            my $release_error = eval { $redis->eval( $release_lua, 1, $lock_name, $token ); 1 } ? undef : $@;
+            get_logger( "Concurrency", "lanraragi" )->error("Failed to release lock ($lock_name): $release_error") if $release_error;
         }
-        $redis->quit() if $own_redis;
+        if ($own_redis) {
+            my $close_error = eval { $redis->quit(); 1 } ? undef : $@;
+            get_logger( "Concurrency", "lanraragi" )->error("Failed to close Redis connection: $close_error") if $close_error;
+        }
         return 0, undef;
     }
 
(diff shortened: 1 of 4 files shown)


psilabs-dev:
Well, this PR kind of went a different direction. Might as well just go all the way (but pinky swear this isn't another RotatingLog v2 lol)

The current `exec_with_lock` mechanism is limited to Mojo API but didn't really need to be, as locking applies to any "resource". So we can take out the "lock" mechanism and refactor it to `exec_with_lock_pure`, which handles all the locking logic instead, while doing a better job at it (e.g. redis connection resilience, lock acquisition rollback).

Multi-lock support is added. This is in anticipation of writes involving multiple resources, e.g. "add archive to category" may be a candidate for this, but that's something for another day. Also, the order should be deterministic.

Lock ownership with tokens is added. This means that even if a lock is acquired by a different worker, they still cannot release the lock as they don't "own" it. A sha strategy was chosen, but tbh we can probably get away with raw `rand`s (though the latency of sha will be drowned by those of redis calls).

Optional TTL is added. This is for Shinobu, where I feel the job *could* take longer than 10s, especially with those category operations..

(`exec_with_lock_pure` isn't a good name, I'd rather have `exec_with_lock` be renamed to `exec_with_lock_render/mojo`, but cosmetic changes aside I'd rather not update more code than I need)

This `exec_with_lock_pure` covers few areas which touch archives: Shinobu at `add_filemap`  , metadata plugin, and batch delete. The way current architecture is, we only need locking coverage of "API"-ish locations: Controller, Shinobu, Minion. However, manual upload isn't covered; there are quite a few differences between manual and API, and I don't think it's worth adding it.

The previous limitation with `exec_with_lock` not knowing what to do after TTL still exists. There's no convenient way to introduce TTL extensions, and the benefits for doing so aren't worth the trouble imo.

These changes should beat down these log assertion failures once and for all! Well of course, they would also make LRR more reliable and predictable to use...

lib/LANraragi/Utils/Generic.pm
```diff
@@ -346,13 +347,95 @@ sub exec_with_lock {
...
+    my $release_lua = <<'LUA';
+    if redis.call('GET', KEYS[1]) == ARGV[1] then
+        return redis.call('DEL', KEYS[1])
+    end
+    return 0
+LUA
+
+    # create managed redis connection if not passed
+    unless ( defined $redis ) {
+        $redis = LANraragi::Model::Config->get_redis;
```
Difegue: I don't think locks should live in the archive DB by default if they're to become generic.  

```suggestion
        $redis = LANraragi::Model::Config->get_redis_config; 
```

lib/Shinobu.pm
```diff
@@ -194,77 +194,95 @@ sub add_to_filemap ( $redis_cfg, $file ) {
...
+            return;
         }
 
-        $logger->debug("Computed ID is $id.");
+        # Acquire exclusive metadata and file write access for archive by ID with 1m timeout
+        my $acquired, $_ = exec_with_lock_pure([ "archive-write:$id" ], sub {
 
-        # If the id already exists on the server, throw a warning about duplicates
-        if ( $redis_cfg->hexists( "LRR_FILEMAP", $file ) ) {
+            $logger->debug("Computed ID is $id.");
```
Difegue: Would be better to move this block of code into a separate sub at this point

lib/LANraragi/Utils/Generic.pm
```diff
@@ -346,13 +347,95 @@ sub exec_with_lock {
...
+
+# exec_with_lock_pure( \@lock_names, $func, $redis (optional), $ttl (optional) )
+# Execute a function under a multi-lock context.
+# Does not implement lock_name sorting,
+# so locks should be passed in a deterministic order to avoid deadlocking.
+# For high-level, fast operations, as the expiry defaults to 10s.
+# Returns a tuple ($acquired, $response), where $response is 
+# function response (if any) if acquired, or undef if not acquired.
+# Redis connection is optional; if not supplied, a managed connection will be created.
+sub exec_with_lock_pure {
```
Difegue, line 364: Yeah I'm not a fan of the name either, would be good to change the API of the mojo variant a bit... That should probably be a TODO comment at least.

lib/LANraragi/Utils/Generic.pm
Difegue: (even through realistically locks are just happening when working with archive IDs/archive DB content at the moment and I don't really see that changing too much, I feel the main DB should really only be user data if possible)

Difegue (changes requested):
tbh this feels slightly overengineered at this point but at the same time the locking logic is well isolated and the outfacing API is OK so why not ig  

I feel the mojo version needs a retool to make it less cumbersome to use at this point (it's going to lock on `archive-write:id` in 95% of the cases so that probably could be refactored in with a signature that just takes an archive ID...), but that can go in a subsequent PR.

Difegue (approved):
:shipit:

PR #1456, opened 2026-01-23
Fix chapter select going to the wrong page and not clearing after removing all entries

Guerra24:


Difegue (approved):
classic off by one error smh

PR #1465, opened 2026-03-05
Add support for Redis connections through unix socket

Wiichele:
I tested this on the latest release using both Unix and TCP connections.

Another approach would be to create a Mojo::URL manually and pass it to Minion but I think this is simpler.


Difegue (approved):
Yup, looks accurate. Thank you!

PR #1468, opened 2026-03-11
Fix some compact mode issues

### LINKED ISSUE #1229: [bug] third custom column sorting is incorrect
OS: docker on macos + safari & chrome

Steps to reproduce:
1. have some tagged data
2. go to compact mode
3. create a third custom column
4. assign column with some tag namespace
5. refresh page (for some reason column is not clickable and results are wrong)
6. toggle ascending/descending

Expected result after toggle: correct sort results.

Actual result:

Image

No idea why

psilabs-dev:
Closes https://github.com/Difegue/LANraragi/issues/1229.

One existing issue and one or two bugs.

First, the issue: we have a TODO in compact mode that disables sort toggle via clicking header. Apparently, this was because before we added customizable column width and that makes any drag event become a click, which triggers a DT draw.

For this issue, I propose that we just distinguish between a "drag" and a "click". If an action is a drag, it's cosmetic/alters DT appearance, and we don't draw. If an action is a click, then it's functional, and we do a draw.

Second is bug 1229. Well, there are two issues.

1. updateTableHeaders calls generateTableHeaders, which calls `headerRow.empty();`... so we destroy the header rows, the things DT only binds to during initialization. This meant that when you edit an extra column the headers become outright ineffective, not just giving wrong sorts.

2. Say I change a column name to "Group" (doesn't have to be new column), then the api sent to perl backend carries "Group" namespace sort key. Perl's backend sort logic is case sensitive, so this request returns nonsense zzzz-sort response. But the frontend `const regex = new RegExp(`${namespace}:([^,]+)`, "gi");` is case *insensitive*, so this nonsense is propagated confusingly to the UI layer.

In reality, "Group" shouldn't correspond to anything because in most backend plugins, it's all lowercase. However, the UI *suggests* that "Group" should be passed instead of "group", because all *user-facing* namespaces have their first letters capitalized.

So this PR enforces the behavior that any input that isn't strictly equal to a namespace in the metadata db is invalid. If the DB has "group" and user passes "Group", invalid. If DB has "Group" and user passes "group", invalid. This makes the problem more obvious if the user did make this type of mistake, as they would see an empty column (instead of a column of unsorted elements that don't react to their endless sort toggling).

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -961,6 +961,7 @@ Index.resizableColumns = function () {
     let currentIndex;
     let startX;
     let startWidth;
+    let didDrag = false;
 
     const headers = document.querySelectorAll("#header-row th");
     headers.forEach(header => {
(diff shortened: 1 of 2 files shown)


Difegue (approved):
Nice job finding a better fix for the click.DT handler, this seems to work fine.  

On the case-sensitiveness I'd tend to think it'd be more helpful to discard the casing when doing the search API request, but that's because I consider namespaces should always be full lowercase in the data layer... There are probably some sickos using casing in their metadata and they might want the extra differences.

PR #1470, opened 2026-03-11
fix: remove hardcoded inline styles from plugin-upload button

### LINKED ISSUE #1469: plugin-upload button has hardcoded inline styles that break custom themes
**LRR Version and OS**
0.9.6

**Bug Details**
In templates/plugins.html.tt2, the element #plugin-upload contains a hardcoded inline style:
style="margin-bottom: -10px"
Additionally, the inner text element uses another inline style:
style="position:absolute; top:5px; left:25%"
Because these styles are defined inline, they override external stylesheets and make it difficult for custom themes to properly control the button layout and alignment through CSS.
It would be better to remove these inline styles and move them into lrr.css, allowing themes to override them more easily.
This change would improve theme compatibility and make the layout more flexible for customization.

```
before:
<span id="plugin-upload" class='stdbtn fileinput-button'
        style="margin-bottom: -10px; font-weight:normal">
        <span style="position:absolute; top:5px; left:25%">[% c.lh("Upload Plugin") %]</span>
        <input type='file' name='file' multiple id='fileupload'>
</span>

after:
<span id="plugin-upload" class='stdbtn fileinput-button'>
        [% c.lh("Upload Plugin") %]
        <input type='file' name='file' multiple id='fileupload'>
</span>
```

nicosuki123:
fix #1469

### DIFF
--- a/templates/plugins.html.tt2
+++ b/templates/plugins.html.tt2
@@ -112,9 +112,8 @@
 					[% c.lh("A script is running...") %]
 				</span>
 				<input id='save' class='stdbtn' type='button' value='[% c.lh("Save Plugin Configuration") %]' />
-				<span id="plugin-upload" class='stdbtn fileinput-button'
-					style="margin-bottom: -10px; font-weight:normal">
-					<span style="position:absolute; top:5px; left:25%">[% c.lh("Upload Plugin") %]</span>
+				<span id="plugin-upload" class='stdbtn fileinput-button'>
+					[% c.lh("Upload Plugin") %]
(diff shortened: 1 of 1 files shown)


Difegue (changes requested):
The removed styles should be put in lrr.css so they still function as expected.  
Custom themes can override lrr.css values with `!important`s so I think that'd fulfill the desired request.

nicosuki123:
That makes sense. Moving those styles to lrr.css should solve the issue.
Custom themes can then override them with !important if needed.
Thanks for the clarification!

PR #1471, opened 2026-03-12
Implement excluded namespace logic

### LINKED ISSUE #1404: Support clientside namespace exclusion of stats results
Many artworks have shared namespace tags, which include create/upload/add date times.

This gets computed everytime in the index table (besides stats table), and pollutes the search suggestions.

Image
Image

I think these namespaces should be blacklisted from `build_tag_stats` by default (see https://github.com/psilabs-dev/LANraragi/issues/71 for more downstream blacklist targets). The quick way to do this is to add blacklist entries in `build_tag_stats` to cover all the fields. Though this might end up being a configuration feature.

psilabs-dev:
Closes https://github.com/Difegue/LANraragi/issues/1404

Adds client-side namespace exclusion logic to search and stats. Updates "/api/info" with an "excluded_namespaces" list[str] field.

## Example

Without "date_added" exclusion, these tags naturally show up in search filter.

Screenshot 2026-03-12 at 12 32 59 AM

random numbers also appear in stats.

Screenshot 2026-03-12 at 12 34 17 AM

Obviously, "date_added" is just an example. Source, any type of ID namespace will also have this problem.

### Configuration

To remove "date_added" field, configure excluded tag namespaces:

Screenshot 2026-03-12 at 12 28 38 AM

No more random numbers in tag cloud:

Screenshot 2026-03-12 at 12 28 22 AM

"date_added" no longer shows up in search filter:

Screenshot 2026-03-12 at 12 29 47 AM

Opt-in: starts with an empty string, users pass namespaces they want to exclude (e.g. "source,date_added").

### DIFF
--- a/lib/LANraragi/Model/Config.pm
+++ b/lib/LANraragi/Model/Config.pm
@@ -210,6 +210,6 @@ sub get_jxlthumbpages    { return &get_redis_conf( "jxlthumbpages",   "0" ) }
 sub get_replacedupe      { return &get_redis_conf( "replacedupe",     "0" ) }
 sub can_replacetitles    { return &get_redis_conf( "replacetitles",   "1" ) }
 sub get_language         { return &get_redis_conf( "language",        "auto" ) }
-sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "" ) }
+sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "source, date_added" ) }
 
 1;
--- a/lib/LANraragi/Model/Config.pm
+++ b/lib/LANraragi/Model/Config.pm
@@ -210,5 +210,6 @@ sub get_jxlthumbpages    { return &get_redis_conf( "jxlthumbpages",   "0" ) }
 sub get_replacedupe      { return &get_redis_conf( "replacedupe",     "0" ) }
 sub can_replacetitles    { return &get_redis_conf( "replacetitles",   "1" ) }
 sub get_language         { return &get_redis_conf( "language",        "auto" ) }
+sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "" ) }
 
 1;
--- a/lib/LANraragi/Controller/Api/Database.pm
+++ b/lib/LANraragi/Controller/Api/Database.pm
@@ -27,8 +27,14 @@ sub drop_database {
 sub serve_tag_stats {
     my $self = shift->openapi->valid_input or return;
     my $minscore = $self->req->param('minweight') || "1";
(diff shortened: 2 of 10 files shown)


Difegue (changes requested):
Shouldn't the namespaces be excluded at the API level? It's kinda jank having to do it on the client when the server already has the information.  

It can be tied to a new API query parameter if we don't want to change the default behavior. (IMO we could, if users exclude namespaces you might as well hide them from stats by default and have a pref to revert that)

Guerra24:
@Difegue Remember last time it broke stuff in the clients. I'm fine with it being handled at the api level but there needs to be a way to query all the namespaces for autocompletion, sorting and filtering purposes.

Difegue:
Yeah I think no matter what this'd need an additional parameter like `minweight`, but since the existing design is to give you the full stats and then apply modifiers it is probably better to have in-API exclusion be opt-in.  

So `/api/database/stats?hide_excluded_namespaces=true` ?

Guerra24:
Sounds fine to me. I think having both the list in /api/info and the param in stats is the way to go.

psilabs-dev:
No objections here, I would prefer compute logic moved to the server if possible.

lib/LANraragi/Model/Config.pm
```diff
@@ -210,5 +210,6 @@ sub get_jxlthumbpages    { return &get_redis_conf( "jxlthumbpages",   "0" ) }
 sub get_replacedupe      { return &get_redis_conf( "replacedupe",     "0" ) }
 sub can_replacetitles    { return &get_redis_conf( "replacetitles",   "1" ) }
 sub get_language         { return &get_redis_conf( "language",        "auto" ) }
+sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "" ) }
```
Difegue:
```suggestion
sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "source, date_added" ) }
```  

I think we should put some defaults in there just to match the current behavior for awesomplete.

Guerra24: FYI the change got lost in the force-push.

Difegue: :|

Difegue:
```suggestion
sub get_excludednamespaces { return &get_redis_conf( "excludednamespaces", "source, date_added" ) }
```

ill fuckin do it again

psilabs-dev: woops, I blame it on the worktree setup

PR #1473, opened 2026-03-12
Update search cache logic to prioritize keyed archives

### LINKED ISSUE #1299: When sorting by custom key put untagged archives at the back
When displaying pages of results as a result of calling the search API with a custom sort key (e.g. "source") in descending order, LRR will present untagged archives first, then followed by tagged archives correctly sorted by "source" in descending order. This is because in `sort_results` the sort values of untagged archives are automatically "zzzz".

I think that whenever sorting is called on a custom key, untagged archives should *always* be put to the back, because when a user is sorting by a key they're already expecting archives to *have* the key. Sorting by "source" in descending order has relevance in the context of many sites whose ID generation is increment-based.

A branch on this is in progress.

psilabs-dev:
Closes https://github.com/Difegue/LANraragi/issues/1299

Terminology:
- An archive is *keyed* with respect to a namespace if it contains a tag prefixed with that namespace; otherwise it's *unkeyed*. For example, untagged archives are always unkeyed, and an archive with tag string "artist:test" is unkeyed WRT namespace "date_added", but keyed WRT namespace "artist".

This means that every search filter and namespace sort creates a partition of keyed and unkeyed archives. 

The existing problem is that when a user decides to sort by a namespace, they are expecting to view *keyed archives*. However, LRR may provide unkeyed archives upfront, forcing the user to navigate several pages before finding the first keyed archive in the sorted sequence.

The goal of this PR is to always place unkeyed (not just untagged) archives at the back, while keeping keyed archives sorted and in the front.

Since this must apply regardless of ascending/descending order, the existing "just reverse the list" logic won't apply. We'd need to sort the keyed part while pushing the unkeyed archives at the back for each cache key and order.

A cache key and order produces a (keyed list, unkeyed list) pair. If order is inverted, take the keyed list only and invert this list, then return (inverted keyed list, unkeyed list) in this order.

### DIFF
--- a/lib/LANraragi/Model/Search.pm
+++ b/lib/LANraragi/Model/Search.pm
@@ -89,7 +89,7 @@ sub check_cache ( $cachekey, $cachekey_inv ) {
 
         my $frozendata = $redis->hget( "LRR_SEARCHCACHE", $cachekey );
         my @cached = @{ thaw $frozendata };
-        shift @cached;
+        shift @cached; # Discard the keyed count, since they're at the bottom of the list naturally
         @filtered = @cached;
 
     } elsif ( $redis->exists("LRR_SEARCHCACHE") && $redis->hexists( "LRR_SEARCHCACHE", $cachekey_inv ) ) {
--- a/lib/LANraragi/Model/Search.pm
+++ b/lib/LANraragi/Model/Search.pm
@@ -53,10 +53,11 @@ sub do_search ( $filter, $category_id, $start, $sortkey, $sortorder, $newonly, $
     # Don't use cache for history searches since setting lastreadtime doesn't (and shouldn't) cachebust
     unless ( $cachehit && $sortkey ne "lastread" ) {
         $logger->debug("No cache available (or history-sorted search), doing a full DB parse.");
-        @filtered = search_uncached( $category_id, $filter, $sortkey, $sortorder, $newonly, $untaggedonly, $grouptanks );
+        my $keyed_count;
+        ( $keyed_count, @filtered ) = search_uncached( $category_id, $filter, $sortkey, $sortorder, $newonly, $untaggedonly, $grouptanks );
 
-        # Cache this query in the search database
-        eval { $redis->hset( "LRR_SEARCHCACHE", $cachekey, nfreeze \@filtered ); };
+        # Cache this query in the search database, prepending the keyed count for partition-aware cache inversion
+        eval { $redis->hset( "LRR_SEARCHCACHE", $cachekey, nfreeze [ $keyed_count, @filtered ] ); };
     }
     $redis->quit();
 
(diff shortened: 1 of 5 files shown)


lib/LANraragi/Model/Search.pm
```diff
@@ -86,17 +87,25 @@ sub check_cache ( $cachekey, $cachekey_inv ) {
         $logger->debug("Using cache for this query.");
         $cachehit = 1;
 
-        # Thaw cache and use that as the filtered list
         my $frozendata = $redis->hget( "LRR_SEARCHCACHE", $cachekey );
-        @filtered = @{ thaw $frozendata };
+        my @cached = @{ thaw $frozendata };
+        shift @cached;
```
Difegue:
```suggestion
        shift @cached; # Discard the keyed count, since they're at the bottom of the list naturally
```

lib/LANraragi/Model/Search.pm
```diff
@@ -300,17 +309,20 @@ sub search_uncached ( $category_id, $filter, $sortkey, $sortorder, $newonly, $un
...
+            return ( $keyed_count, @filtered );
         }
     }
 
     $redis->quit();
     $redis_db->quit();
-    return @filtered;
+
+    # Title sort and unfiltered results: all archives are keyed
+    return ( scalar @filtered, @filtered );
```
Difegue: Just return `-1, @filtered` here to avoid an extra `scalar`.

lib/LANraragi/Model/Search.pm
```diff
@@ -505,19 +517,36 @@ LUA
...
-        # Read comments from the bottom up for a better understanding of this sort algorithm.
-        @sorted = map { $_->[0] }                  # Map back to only having the ID
-          sort { ncmp( $a->[1], $b->[1] ) }        # Sort by the tag
-          map  { [ $_, lc( $tmpfilter{$_} ) ] }    # Map to an array containing the ID and the lowercased tag
-          @filtered;                               # List of IDs
+        # Partition: archives with the sort namespace vs those without it
+        my @keyed_ids   = grep { $tmpfilter{$_} ne "zzzz" } @filtered;
+        my @unkeyed_ids = grep { $tmpfilter{$_} eq "zzzz" } @filtered;
+
+        # Sort only the keyed archives
```
Difegue: Please keep the comments of the sort algorithm in there

Difegue (changes requested):
This probably needs a few additional test cases as well.

psilabs-dev:
> This probably needs a few additional test cases as well.

Integration tests cover e2e sorting behavior, given Redis caching is needed. We can add unit testing with mocks, though it might affect the current test data complexity (as mock data is shared across test cases). I would then recommend creating isolated mock data specifically for this behavior, as future test cases may fail this test case.

lib/LANraragi/Model/Search.pm
psilabs-dev: oops myb

Difegue (approved):

Difegue:
imo a few extra mocks would be nice just to test the basic ordering logic without running the full integration suite, but otherwise i'm happy with this 👍

psilabs-dev:
added

PR #1477, opened 2026-03-14
Make image preloading consistent across all browsers

### LINKED ISSUE #1433: Preloading fails over slow networks (Safari)
In a slow/restricted network where image fetches take up to 6-10 seconds, the image preload functionality fails, resulting in each page change to be accompanied by a spinning wheel.

Checking the network tab shows that despite the subsequent image already fetched by the browser, the UI will *still* fetch the same image on page change, causing the user to see 6 seconds of loading.

Image

The simplified expected behavior of preloading should be that (with preload=3):
- user opens page 1
- UI fetches and serves page 1, fetching/cache-writing 2/3 in the background
- user changes to page 2
- UI serves page 2 immediately, fetching 4 in the background

Instead, what happens is:
- user opens page 1
- UI fetches and serves page 1, fetching 2/3 in the background (I think?)
- user changes to page 2
- UI fetches page 2 (again), tells user to wait, then serves page 2, fetching 4 in the background (I think?)

I'm not sure if currently LRR's preloading/img serve functionality are distinct, or if preloading must be done before any images are served at all.

(To clarify, I don't think preload works regardless of whether the network is slow or not; I think that the problem just becomes apparent in a slow network.)

Guerra24:
Use fetch and friends to load images in a stable way across all browsers so that preload works correctly.

Safari should go die in hell.

Fixes #1433.

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -75,7 +75,8 @@ jobs:
             --staging "$GITHUB_WORKSPACE/staging" \
             --playwright \
             --docker-api \
-            --npseed 42
+            --npseed 42 \
+            -k "not test_double_page_navigation and not test_handler_resource_management"
         env:
           DOCKER_HOST: unix:///var/run/docker.sock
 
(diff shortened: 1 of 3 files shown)


psilabs-dev:
Thanks for taking the time to work on this issue!

Regarding the failures, they are test-side and can be ignored with the `-k` flag. You can add this at the end of each `uv run pytest` block in `.github/workflows/push-continuous-integration.yml`:

```
-k "not test_double_page_navigation and not test_handler_resource_management"
```

e.g.

```yaml
run: |
uv run pytest tests --log-cli-level=INFO \
...
-k "not test_double_page_navigation and not test_handler_resource_management"
```

or wait until the next integration test update (if this is merged, then these two cases will no longer apply moving forward).

PR #1482, opened 2026-03-17
fix a bunch of test failures

### LINKED ISSUE #1479: Add a caching layer for test workflows
It would probably help to add a caching layer for both Docker and Windows integration tests, seeing as a portion of failures are build-time rate limits that prevent further testing.

For docker, we could cache at `RUN sh ./tools/install-everything.sh $INSTALL_PARAMETER`.

For Windows, we should cache somewhere similar at the dependency installation stage, or based on a hash of `tools/build/Windows`.

This would shave off 10 minutes from Windows github runs, and also a bit of time off Docker as well, while also avoiding running into that Alien build failure for the nth time.

psilabs-dev:
closes #1479

1. before we discussed moving redis lock from archive DB to config DB, this process was incomplete. (PR https://github.com/Difegue/LANraragi/pull/1455)
2. also had a nested lock issue (PR https://github.com/Difegue/LANraragi/pull/1455)
3. fix maketext
4. demote rate limit error logs to warning
5. cached test-time builds

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -44,6 +44,19 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: lanraragi
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      - name: Build LRR Docker image
+        uses: docker/build-push-action@v7
+        with:
+          context: lanraragi
+          file: lanraragi/tools/build/docker/Dockerfile
+          tags: lanraragi:ci
+          outputs: type=docker,dest=${{ runner.temp }}/lanraragi-ci.tar
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Load LRR Docker image
+        run: docker load -i "${{ runner.temp }}/lanraragi-ci.tar"
       - name: Checkout aio-lanraragi integration tests
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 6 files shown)


psilabs-dev:
just going to undraft this while tests are running, I noticed a maketext error but that's probably for one of the recent raty changes.

psilabs-dev:
cached Windows tests shave off 10-ish minutes, this should clean things up.

.github/workflows/push-continuous-integration.yml
```diff
@@ -119,16 +149,32 @@ jobs:
...
         run: |
           ./tools/build/windows/install.sh
           ./tools/build/windows/cleanup.sh
           ./tools/build/windows/create-dist.sh
       - name: Enable UTF-8
+        if: steps.cache-windist.outputs.cache-hit != 'true'
         working-directory: LANraragi
         run: |
           ./tools/build/windows/utf8-support.ps1
+      - name: Create windist zip for cache
```
Guerra24, line 164: The only issue I see with this is that the code never gets updated after the first time the cache is created. `create-dist.sh` is what copies the code and that will only run on that first time.

Maybe splitting that script into two, one for the actual runtime creation and another that copies the server code. Split `create-dist.sh` at line 47.

psilabs-dev: Oh, yikes. Thanks for catching this.

PR #1493, opened 2026-03-27
Stamps Backend

### LINKED ISSUE #1355: [Feature Request] O-Counter / Nut-Button
I would really like an "O-Counter" (Maybe call it "Nut-Button") in the sense as the github project "stashapp/stash" already incorporated.

Just a button you can press to keep track of how many times the work was used...
Also maybe a small achievement Systems with Milestones? (One could dream big.) XD

The button should have a squirt icon or some kind of Otaku related design.
Would also be maybe an "easy" plugin?
Is there such a plugin already? I didn't find any...

The button should register the time and date each time the button has been pressed in the database and simply count upwards and display how many times the work was "used". 

Ideally there is also a way to edit or add a time and date or simply delete single events. (Stash exactly implemented it like that.)

EfronC:
Add support for Stamps in the backend.

This is just an initial PR to share the progress with the backend changes, just so I can start the get comments about the implementation. Will be working on some UI aspects on my next commits.

Stamps(AKA stickies/favorites/bookmark) are integrated using Hash tables in Redis. The logic is that the field is composed of a string like "page:creation_timestamp", so that way is easy to identify which page the stamp belongs to, also we have the creation date in case is useful. The value inside the stamp is conformed by a string in the shape of "position|content", where content is the text we want the stamp to have, and position for the moment is a combo field just to show that is possible to add more fields inside as long as we follow this structure of using | as a separator. I'm calling it position in case some client developer wants to use them as positional annotations, then it is possible to save the coordinates in this space, otherwise it does nothing.

There might be errors related to OpenAPI in this version, since I don't really have a lot of time to test the specifications, but I hope this can be solved with the time.

### DIFF
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -32,11 +32,6 @@ sub get_stamps_by_page {
 
     my ( $stamps, $err ) = LANraragi::Model::Stamp::get_stamps_by_page($id, $index);
 
-    unless (@$stamps) {
-        render_api_response($self, "get_stamps_by_page", "The given page does not have stamps.");
-        return;
-    }
-
     $self->render( openapi => { result => $stamps } );
 }
 
@@ -94,8 +89,8 @@ sub update_stamp {
     my $self        = shift->openapi->valid_input or return;
     my $id          = $self->stash('id');
     my $stamp_id    = $self->req->param('stamp_id');
-    my $position    = $self->req->param('position');
-    my $content     = $self->req->param('content') || "";
+    my $position    = $self->req->param('position') || undef;
+    my $content     = $self->req->param('content') || undef;
 
     return unless exec_with_lock(
         $self,
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -1,0 +1,145 @@
+package LANraragi::Controller::Api::Stamp;
+use Mojo::Base 'Mojolicious::Controller';
+
+use Redis;
+use Encode;
+
+use LANraragi::Model::Stamp;
+use LANraragi::Utils::Generic qw(render_api_response exec_with_lock);
+
+
+sub get_stamp {
+
+    my $self    = shift->openapi->valid_input or return;
+    my $id      = $self->stash('id');
+    my $stamp_id    = $self->req->param('stamp_id');
+
+    my ( $stamp, $err ) = LANraragi::Model::Stamp::get_stamp($id, $stamp_id);
+
+    unless (%$stamp) {
+        render_api_response($self, "get_stamp", "The given stamp does not exist.");
+        return;
+    }
+
+    $self->render( openapi => { result => $stamp } );
+}
+
+sub get_stamps_by_page {
+
+    my $self    = shift->openapi->valid_input or return;
+    my $id      = $self->stash('id');
+    my $index    = $self->stash('index');
+
+    my ( $stamps, $err ) = LANraragi::Model::Stamp::get_stamps_by_page($id, $index);
+
+    unless (@$stamps) {
+        render_api_response($self, "get_stamps_by_page", "The given page does not have stamps.");
+        return;
+    }
+
+    $self->render( openapi => { result => $stamps } );
+}
+
+sub get_stamped_pages {
+
+    my $self    = shift->openapi->valid_input or return;
+    my $id      = $self->stash('id');
+
+    my ( @indexes, $err ) = LANraragi::Model::Stamp::get_stamped_pages( $id );
+
+    unless (@indexes) {
+        render_api_response( $self, "get_stamped_pages", "The given archive does not have stamps." );
+        return;
+    }
+
+    $self->render( openapi => { result => \@indexes } );
+}
+
+sub add_stamp {
+
+    my $self    = shift->openapi->valid_input or return;
+    my $id      = $self->stash('id');
+    my $index    = $self->stash('index');
+    my $content = $self->req->param('content') || "";
+    my $position = $self->req->param('position') || "";
+
+    unless ( defined $index ) {
+        return render_api_response( $self, "add_stamp", "Archive page." );
+    }
+
+    my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
+
+    if ($created_id) {
+        $self->render(
+            openapi => {
+                operation    => "add_stamp",
+                stamp_id => $created_id,
+                success      => 1
+            }
+        );
+    } else {
+        $self->render(
+            openapi => {
+                operation    => "add_stamp",
+                stamp_id => $created_id,
+                success      => 0
+            }
+        );
+    }
+
+}
+
+sub update_stamp {
+
+    my $self        = shift->openapi->valid_input or return;
+    my $id          = $self->stash('id');
+    my $stamp_id    = $self->req->param('stamp_id');
+    my $position    = $self->req->param('position');
+    my $content     = $self->req->param('content') || "";
+
+    return unless exec_with_lock(
+        $self,
+        "stamp-write:$stamp_id",
+        "update_stamp",
+        $stamp_id,
+        sub {
+            my ( $result, $err ) = LANraragi::Model::Stamp::update_stamp( $id, $stamp_id, $content, $position );
+
+            if ($result) {
+                my %stamp      = LANraragi::Model::Stamp::get_stamp( $id, $stamp_id );
+                my $successMessage = "Updated stamp \"$stamp{id}\"!";
+
+                render_api_response( $self, "update_stamp", undef, $successMessage );
+            } else {
+                render_api_response( $self, "update_stamp", $err );
+            }
+        }
+    );
+
+}
+
+sub delete_stamp {
+
+    my $self   = shift->openapi->valid_input or return;
+    my $id = $self->stash('id');
+    my $stamp_id = $self->req->param('stamp_id');
+
+    return unless exec_with_lock(
+        $self,
+        "stamp-write:$stamp_id",
+        "delete_stamp",
+        $stamp_id,
+        sub {
+            my ( $result, $err ) = LANraragi::Model::Stamp::remove_stamp($id, $stamp_id);
+
+            if ($result) {
+                render_api_response( $self, "delete_stamp" );
+            } else {
+                render_api_response( $self, "delete_stamp", "The given stamp does not exist." );
+            }
+        }
+    );
+}
+
+1;
+
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -5,7 +5,7 @@ use Redis;
 use Encode;
 
 use LANraragi::Model::Stamp;
-use LANraragi::Utils::Generic qw(render_api_response exec_with_lock);
+use LANraragi::Utils::Generic qw(render_api_response exec_with_lock exec_with_lock_pure);
 
 
 sub get_stamp {
@@ -53,29 +53,37 @@ sub add_stamp {
     my $position = $self->req->param('position') || "";
 
     unless ( defined $index ) {
-        return render_api_response( $self, "add_stamp", "Archive page." );
+        return render_api_response( $self, "add_stamp", "No specified page for the stamp to attach to." );
     }
 
-    my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
+    return unless exec_with_lock(
+        $self,
+        "archive-write:$id",
+        "update_archive",
+        $id,
+        sub {
+            my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
 
-    if ($created_id) {
-        $self->render(
-            openapi => {
-                operation   => "add_stamp",
-                stamp_id    => $created_id,
-                success     => 1
+            if ($created_id) {
+                $self->render(
+                    openapi => {
+                        operation   => "add_stamp",
+                        stamp_id    => $created_id,
+                        success     => 1
+                    }
+                );
+            } else {
+                $self->render(
+                    openapi => {
+                        operation   => "add_stamp",
+                        stamp_id    => $created_id,
+                        success     => 0,
+                        error       => $err
+                    }
+                );
             }
-        );
-    } else {
-        $self->render(
-            openapi => {
-                operation   => "add_stamp",
-                stamp_id    => $created_id,
-                success     => 0,
-                error       => $err
-            }
-        );
-    }
+        }
+    );
 }
 
 sub update_stamp {
@@ -111,21 +119,35 @@ sub delete_stamp {
     my $self        = shift->openapi->valid_input or return;
     my $stamp_id    = $self->stash('id');
 
-    return unless exec_with_lock(
-        $self,
-        "stamp-write:$stamp_id",
-        "delete_stamp",
-        $stamp_id,
-        sub {
-            my ( $result, $err ) = LANraragi::Model::Stamp::remove_stamp($stamp_id);
+    my ( $result, $id ) = LANraragi::Model::Stamp::get_stamp_archive_id($stamp_id);
 
-            if ($result) {
-                render_api_response( $self, "delete_stamp" );
-            } else {
-                render_api_response( $self, "delete_stamp", $err );
+    if ( $result ) {
+        my ( $acquired, $response ) = exec_with_lock_pure(
+            [ "archive-write:$id", "stamp-write:$stamp_id" ],
+            sub { 
+                my ( $result, $err ) = LANraragi::Model::Stamp::remove_stamp($stamp_id);
+
+                if ( $result ) {
+                    render_api_response( $self, "delete_stamp" );
+                } else {
+                    render_api_response( $self, "delete_stamp", $err );
+                }
             }
+        );
+
+        if ( !$acquired ) {
+            $self->render(
+                json => {
+                    operation => "delete_stamp",
+                    success   => 0,
+                    error     => "Locked resource"
+                },
+                status => 423
+            );
         }
-    );
+    } else {
+        render_api_response( $self, "delete_stamp", $id );
+    }
 }
 
 1;
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -16,7 +16,7 @@ sub get_stamp {
 
     my ( $stamp, $err ) = LANraragi::Model::Stamp::get_stamp($id, $stamp_id);
 
-    unless (%$stamp) {
+    unless ($stamp) {
         render_api_response($self, "get_stamp", "The given stamp does not exist.");
         return;
     }
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -57,33 +57,26 @@ sub add_stamp {
         return render_api_response( $self, "add_stamp", "Archive page." );
     }
 
-    return unless exec_with_lock(
-        $self,
-        "stamp-write:$id",
-        "create_stamp",
-        $id,
-        sub {
-            my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
+    my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
 
-            if ($created_id) {
-                $self->render(
-                    openapi => {
-                        operation    => "add_stamp",
-                        stamp_id => $created_id,
-                        success      => 1
-                    }
-                );
-            } else {
-                $self->render(
-                    openapi => {
-                        operation    => "add_stamp",
-                        stamp_id => $created_id,
-                        success      => 0
-                    }
-                );
+    if ($created_id) {
+        $self->render(
+            openapi => {
+                operation   => "add_stamp",
+                stamp_id    => $created_id,
+                success     => 1
             }
-        }
-    );
+        );
+    } else {
+        $self->render(
+            openapi => {
+                operation   => "add_stamp",
+                stamp_id    => $created_id,
+                success     => 0,
+                error       => $err
+            }
+        );
+    }
 }
 
 sub update_stamp {
@@ -132,7 +125,7 @@ sub delete_stamp {
             if ($result) {
                 render_api_response( $self, "delete_stamp" );
             } else {
-                render_api_response( $self, "delete_stamp", "The given stamp does not exist." );
+                render_api_response( $self, "delete_stamp", $err );
             }
         }
     );
--- a/lib/LANraragi/Controller/Api/Stamp.pm
+++ b/lib/LANraragi/Controller/Api/Stamp.pm
@@ -10,11 +10,10 @@ use LANraragi::Utils::Generic qw(render_api_response exec_with_lock);
 
 sub get_stamp {
 
-    my $self    = shift->openapi->valid_input or return;
-    my $id      = $self->stash('id');
-    my $stamp_id    = $self->req->param('stamp_id');
+    my $self        = shift->openapi->valid_input or return;
+    my $stamp_id    = $self->stash('id');
 
-    my ( $stamp, $err ) = LANraragi::Model::Stamp::get_stamp($id, $stamp_id);
+    my ( $stamp, $err ) = LANraragi::Model::Stamp::get_stamp($stamp_id);
 
     unless ($stamp) {
         render_api_response($self, "get_stamp", "The given stamp does not exist.");
@@ -82,8 +81,7 @@ sub add_stamp {
 sub update_stamp {
 
     my $self        = shift->openapi->valid_input or return;
-    my $id          = $self->stash('id');
-    my $stamp_id    = $self->req->param('stamp_id');
+    my $stamp_id    = $self->stash('id');
     my $position    = $self->req->param('position') || undef;
     my $content     = $self->req->param('content') || undef;
 
@@ -93,10 +91,10 @@ sub update_stamp {
         "update_stamp",
         $stamp_id,
         sub {
-            my ( $result, $err ) = LANraragi::Model::Stamp::update_stamp( $id, $stamp_id, $content, $position );
+            my ( $result, $err ) = LANraragi::Model::Stamp::update_stamp( $stamp_id, $content, $position );
 
             if ($result) {
-                my %stamp      = LANraragi::Model::Stamp::get_stamp( $id, $stamp_id );
+                my %stamp      = LANraragi::Model::Stamp::get_stamp( $stamp_id );
                 my $successMessage = "Updated stamp \"$stamp_id\"!";
 
                 render_api_response( $self, "update_stamp", undef, $successMessage );
@@ -110,9 +108,8 @@ sub update_stamp {
 
 sub delete_stamp {
 
-    my $self   = shift->openapi->valid_input or return;
-    my $id = $self->stash('id');
-    my $stamp_id = $self->req->param('stamp_id');
+    my $self        = shift->openapi->valid_input or return;
+    my $stamp_id    = $self->stash('id');
 
     return unless exec_with_lock(
         $self,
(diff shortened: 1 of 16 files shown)


EfronC:
Current issues yet to fix:

- Loading spinner does not disappear when filtering only stamped pages in ArchiveOverlay.
- GET stamped pages returns as an array instead of an object.

Also:

- Check support for Double pages and Infinite scroll(Not sure if stamps can be implemented on those)

EfronC:
Stamps is ready for a PR review. 

Little video of how it works

[Stamps.webm](https://github.com/user-attachments/assets/fabbd31e-8a1f-43d6-b0c5-21b9ecf62ff7)

This PR still has a few issues:

- For the localization, I'm not sure how it works, I just added some lines to the i18n.html.tt2, but I probably need to modify other files too.
- Since this set new events in the page, I'm 99% sure that I probably broke something in the reader, since I cannot check that every single feature still works after this. For starters, when you open the ArchiveOverlay, set a stamp, and it goes back to the ArchiveOverlay, the shaded part behind dissapears(Not sure why).
- The EP to get all Stamps in a page still needs pagination, but I'm not sure if we require pagination for this feature...
- ~~I attempted to implement a drag and drop to also modify the position of the stamp, but didn't work, gave me a lot of problems that couldn't understand, from the page change happening after the keyup event triggered, to the wrong stamp getting modified, that I ended giving up. I left the code commented in case that someone else wants to attempt to make it work, otherwise can be removed and leave it with the current behavior~~.

I'll wait for the comments with the required changes from the review.

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,135 @@
...
+    my $id      = $self->stash('id');
+    my $index    = $self->stash('index');
+    my $content = $self->req->param('content') || "";
+    my $position = $self->req->param('position') || "";
+
+    unless ( defined $index ) {
+        return render_api_response( $self, "add_stamp", "Archive page." );
+    }
+
+    my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
```
psilabs-dev: This is missing a Redis lock

tools/openapi.yaml
```diff
@@ -3336,6 +3338,291 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/OperationResponse'
+  /stamps/{id}:
+    get:
+      security:
+        - api_key: []
```
psilabs-dev: Do we consider stamps private info/requiring authentication?

EfronC: That's an excellent question that I could not answer myself when making the EPs, so that's why I left everything with api key requirements, although probably this should not be necessary at least for the Get, so will remove it from there.

EfronC:
Something I forgot to put in my previous comment

Disclaimer: Many of the UI designs(Like CSS and JS DOM interactions) were made using AI, because I'm not proficient with the UI side. Also in this new commit, the hscan mock was made also by with AI to save me some time.

I feel like is necessary to mention these kind of things.

tools/openapi.yaml
EfronC: Done, removed the requirement from the GETs, still will wait for opinions over if stamps should be user specific or public.

EfronC:
Today I woke up inspired and fixed the draggable stamps. Here an example:

output

tools/openapi.yaml
```diff
@@ -3783,5 +4064,24 @@ components:
...
+      required:
+        - position
+        - content
+      properties:
+        id:
+          type: string
+          description: ID of the stamp
+        position:
+          type: string
+          description: Position of the stamp in the page
```
Difegue:
```suggestion
          description: Position of the stamp in the page in pixel coordinates
```

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+        $position = $c_content[0]
+    }
+
+    if ( defined $content ) {
+        $content = remove_separator($content, "|");
+    } else {
+        $content = $c_content[1]
+    }
+
+    if ( $redis->exists($faves_id) ) {
```
psilabs-dev: should be put before hget or wrapped around to gate against updating ghost stamps

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+use LANraragi::Utils::Redis      qw(redis_encode);
+
+
+# get_stamp(id, stamp_id)
+#   Gets the requested stamp.
+#   Returns the stamp object.
+sub get_stamp {
+    my ( $id, $stamp_id ) = @_;
+
+    my $redis  = LANraragi::Model::Config->get_redis;
```
psilabs-dev, line 25: needs a `$redis->quit`

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+    my $logger = get_logger( "Stamps", "lanraragi" );
+    my $faves_id = "FAVES_" . $id;
+    my $err     = "";
+
+    my $data = get_stamps_data($redis, $faves_id, $index);
+    my @stamps = convert_stamps_to_object(%$data);
+
+    return ( \@stamps, $err );
```
psilabs-dev, line 73: same redis cleanup needed

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+
+    foreach my $field (@$fields) {
+        # Extract the page number
+        my ($index) = split(/:/, $field, 2);
+        $indexes{$index} = 1;
+    }
+
+    my @keys = keys %indexes;
+
+    return ( \@keys, $err );
```
psilabs-dev, line 118: same redis cleanup needed

tools/openapi.yaml
```diff
@@ -3367,6 +3369,285 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: ID of the archive.
+          schema:
+            type: string
+        - name: stamp_id
+          in: query
+          required: True
```
psilabs-dev: minor inconsistency (probably not worth a review) but `True` vs. `true`

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,142 @@
...
+    my $content = $self->req->param('content') || "";
+    my $position = $self->req->param('position') || "";
+
+    unless ( defined $index ) {
+        return render_api_response( $self, "add_stamp", "Archive page." );
+    }
+
+    return unless exec_with_lock(
+        $self,
+        "stamp-write:$id",
```
psilabs-dev: Stamps could probably be archive-scoped?

Difegue: If they're not stored in the archive object in the DB there's technically no need for it, right?

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,142 @@
...
+                        stamp_id => $created_id,
+                        success      => 1
+                    }
+                );
+            } else {
+                $self->render(
+                    openapi => {
+                        operation    => "add_stamp",
+                        stamp_id => $created_id,
+                        success      => 0
```
Difegue: No `err` in the output as message?

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+
+# get_stamp(id, stamp_id)
+#   Gets the requested stamp.
+#   Returns the stamp object.
+sub get_stamp {
+    my ( $id, $stamp_id ) = @_;
+
+    my $redis  = LANraragi::Model::Config->get_redis;
+    my $logger = get_logger( "Stamps", "lanraragi" );
+    my $faves_id = "FAVES_" . $id;
```
Difegue: Imo the internal key should match the feature, so `STAMPS_[id]`.  

(I have potential thoughts about the DB architecture of this too but if we keep it like this I'd at least like that to be changed)

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,142 @@
...
+        "stamp-write:$stamp_id",
+        "delete_stamp",
+        $stamp_id,
+        sub {
+            my ( $result, $err ) = LANraragi::Model::Stamp::remove_stamp($id, $stamp_id);
+
+            if ($result) {
+                render_api_response( $self, "delete_stamp" );
+            } else {
+                render_api_response( $self, "delete_stamp", "The given stamp does not exist." );
```
Difegue: DItto, should probably use `err` here? Since the function has its own "id not present" error message on return already

lib/LANraragi/Model/Stamp.pm
Difegue: An early return if the ID doesn't exist would probably be cleaner tbh

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+
+    $err = "$faves_id doesn't exist in the database!";
+    $logger->warn($err);
+    $redis->quit;
+    return ( 0, $err );
+}
+
+# Replaces | for " " in the given string
+sub remove_separator {
+    my ($string, $char) = @_;
```
Difegue: Should be in Utils::String

tools/openapi.yaml
```diff
@@ -3367,6 +3369,285 @@ paths:
...
+                    enum:
+                      - 0
+                      - 1
+        '400':
+          description: Error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/OperationResponse'
+  /stamps/pages/{id}:
```
Difegue: Should this be `/api/archives/{id}/stamps`? Similarly to what we do for categories and tanks.

tools/openapi.yaml
```diff
@@ -3367,6 +3369,285 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/OperationResponse'
+    put:
+      security:
+        - api_key: []
+      operationId: addStamp
+      x-mojo-to: api-stamp#add_stamp
+      summary: 🔑 Add a stamp annotation
+      description: Add a new annotation to the page as a page sticky.
```
Difegue:
```suggestion
      description: Add a new Stamp to the page at the given coordinates.
```

lib/LANraragi/Controller/Api/Stamp.pm
EfronC: What do you mean by this?

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+    my ($redis, $faves_id, $index) = @_;
+
+    my $cursor = 0;
+    my %result;
+    my $pattern = "$index:*";
+    my $logger = get_logger( "Stamps", "lanraragi" );
+
+    # Use a Do While until the cursor goes back to 0
+    do {
+        my ($next_cursor, $data) = $redis->hscan($faves_id, $cursor, 'MATCH', $pattern);
```
Difegue: I have concerns about going for a "store all the stamps for an ID in a single hash" approach, I think it makes the code pretty hard to work with and extend if we ever want to add more fields.  

I'd also like to potentially add a "view all stamps" page to the main UI (with a matching API endpoint) which would be a bit more annoying to do with this approach.  

I don't see why we couldn't go with a simpler approach by making each stamp a dedicated hash?  

- Archive objects gain a new `stamps` field that's an array of STAMP_[timestamp] IDs (or timestamp+guid if we want to make automation easier)
- A Stamp hash contains the page, the X/Y values, the timestamp, and the text.

Difegue (changes requested):
Thanks for splitting the frontend, that makes review easier on my end.  
The API looks sound to me, but I have questions on the DB structure itself as I think we could do it in a potentially simpler way.  

Also misc points:  
- This needs documentation of the DB structure in Architecture
- Stamps should probably be added to backup.json exports like categories.

lib/LANraragi/Model/Stamp.pm
EfronC: Damnit! I was hoping to be able to infiltrate this easter egg hahaha

lib/LANraragi/Model/Stamp.pm
EfronC: This is also something that I thought about, and TBH, maybe is because I've never used Redis before as a storage DB(I've only used it for Queues), but one of the things I considered is that, because there is no limit in how Stamps can be used, then I'm not sure how using it as a distributed set of tables could impact the platform, for example:

Let's say that you have a user with l archives, each with m pages each, you can put any amount of stamps on every page, hence lets say we have someone that for some reason starts stamping every single detail on every page, to the point that puts n stamps on every page. Following this idea you will end up with `(l archives)*(m pages)*(n stamps per page) `number of new Hash tables, which means that stamps can grow faster than the number of files. I'm not sure if this might end up impacting the performance of the DB, or the RAM consumption, at some point, that's why I choose to have them collected on one table per archive.

Also, nitpick, but this model allows us to use hscan to filter only the stamps for one page, otherwise, on this comment version, you would need to go through every stamp in the stamps field, and check the page to verify if it belongs to the requested page. Not something hard to do, but kind of feels more correct this way. Still this is not a deal breaker.

lib/LANraragi/Controller/Api/Stamp.pm
psilabs-dev: Yes, on second look it should be stamp-scoped locks, but I'm getting confused here. `my $id` is an archive ID, not a stamp ID (not created yet). But both are used for "stamp-write:" lock prefixes?

psilabs-dev: Either way, probably would bring the ID generation up before `exec_with_lock` happens, and make ID used for resource lock acquisition consistent

EfronC: The answer to all your questions is: I didn't understand what the lock message usage was, so I just copy-pasted it from Tanks, and just put the file id. If you think that there is a better field to use for this, I'm open to opinions. :+1:

psilabs-dev: Then for simplicity, I think we should remove locking logic from `add_stamp`, making the logic consistent with adding category/tank predecessors. `exec_with_lock` is not necessarily required here for this endpoint (we can address this issue later).

Now to explain the lock: the redis lock is used to protect concurrent write operations and prevent two processes from updating stamp metadata at the same time. The way this is handled is by having one process acquire a lock with ID "stamp-write:stamp-id", which means "I will be writing to stamp-id, so no one else may write to it".

Stamp ID corresponds to the stamp resource, so processes may update a stamp if and only if they acquired the lock corresponding to the stamp ID. The purpose is defeated if this lock ID is instead something completely different (such as the arcid used in this case).

EfronC: Makes sense, although, Could we use something like `"stamp-write:<arcid>_<stamp-id>"`, because in its current state, stamp-id is not unique, if you add a stamp at page 0 of 2 different archives at the exact same time, the 2 would have the same stamp-id, hence the lock would block writtings on unrelated stamps.

psilabs-dev: Yeah, tho that's the issue somewhat present in all existing tank/categories too (it's going to happen all the time during tests) ... There are two approaches: 

1. add the retry loop that tank/categories do (uniqueness check with increment). This is more of a patch, we'll still run into conflicts but they are "handled".
2. uuid, which sidesteps this problem altogether

Either way is fine by me tbh, bc this wouldn't be considered a "surprise" if the feature is merged

EfronC: Will wait for Difegue answer about the DB structure from one of the other ↓ comments before doing this change. Maybe could not be needed.

lib/LANraragi/Model/Stamp.pm
Difegue: I'm really not concerned by the size of the amount of objects in DB, redis is designed to handle up to 250 million keys. We'll hit user RAM limits way before that, and that will be the same size regardless of how we structure the data.  

> on this comment version, you would need to go through every stamp in the stamps field, and check the page to verify if it belongs to the requested page  

Perhaps we could add the page number to the hash key in that case? I think the reader would likely hold all the stamps in memory at once anyway instead of going back and forth with API calls on every page. (Although doing per-page would work too, i guess it's a matter of seeing what works best)

lib/LANraragi/Controller/Api/Stamp.pm
Difegue: i'd rather go with the pomf-style increment loop first btw (mainly for codebase alignment), we can revisit UUIDs in the future.

tools/openapi.yaml
```diff
@@ -1851,6 +1853,39 @@ paths:
...
+          required: true
+          description: ID of the archive.
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Success response
+          content:
+            application/json:
+              schema:
```
Difegue, line 1910: Would be nice to have a component ref in there if possible

Difegue (changes requested):
Sorry for the delay 😶‍🌫️
The changes to the previous comments look OK to me, I guess we just need the DB structure revamp and the architecture doc updates?

lib/LANraragi/Controller/Api/Stamp.pm
EfronC: I already created a separate PR for UUID integration #1533 and there I mentioned that I don't think Stamps is a good place to implement them, although given your other comment there might be a chance to use it, but I think is better to implement UUID on that separate PR.

lib/LANraragi/Model/Stamp.pm
EfronC: While I start making the methods to have stamps as separated hash tables, one interesting idea: What if instead of having stamps being a list of IDs inside every archive table, we make a separate cache to link stamps with their archives(Like a LRR_STAMPCACHE)? Not saying that this is better or anything, but maybe this could save us the need to add these stamps lists into the backup file, which would make the file bigger and bigger if we start appending every new feature to the Archive object. Just a random thought I had today, not sure if this makes sense on the first place.

image

Meanwhile I will start working it as a list of IDs inside every archive.

EfronC: Added the list of IDs to Archive in the meantime for the current version. Still will let you resolve this comment if the current approach is good enough.

lib/LANraragi/Controller/Api/Stamp.pm
EfronC: Since UUID will be worked on a separate issue, and I added an increment-loop id for stamps, I guess that's all we need for this comment.

psilabs-dev:
A couple questions on the feature spec, I know the code probably answers this but it's better/easier to just have these as comments for future reference in case the code changes

(Also, bc a few integration tests are failing, but they could just be AI doing AI things)

- what is the status code(s) for getting/updating/deleting a stamp that doesn't exist?
- if an archive is deleted are the stamps all deleted or kept?
- what happens when a stamp is added to an archive that doesn't exist, or an out of bounds index?
- any sanitization for position, or is this a client-side concern?
- are stamps contents unlimited in size? not that they *need* to be limited ig
- stamps don't intersect with any feature, such as category, tankoubon, or table of contents?

EfronC:
Most of the design decisions I've made on Stamps are following how we do things on other parts of the platform, so I'm not gonna say that this is the right way to make this API, if you think that something can be done differently for the purpose of the tests, I'm open to do the changes. Now answering your questions:

> * what is the status code(s) for getting/updating/deleting a stamp that doesn't exist?

These 3 actions are served through the render_api_response method, so in case of an error, returns a 400. In the code, I check that the Stamp id exist in the DB, and returns an error if doesn't.

> * if an archive is deleted are the stamps all deleted or kept?

Currently they are kept, I haven't touched the cleaning part, because I wanted to be sure of the structure of the DB first. Originally, with the `STAMPS_<archive>` format it would have been possible to even have a cron job clean it, but since stamps are now bound to the archive, I think I'll have to add one more cleaning step to the archive delete method.

> * what happens when a stamp is added to an archive that doesn't exist, or an out of bounds index?

We check that the archive exists beforehand, so the first case is not possible, it throws an error, the second case is possible, since the page number is just a value, but the stamp will not be accessible later without hardcoding the number in the request. Although, I guess that I need to check on the Frontend what happens when stampedpages EP returns an out of index page at filtering the pages in the Overlay.

> * any sanitization for position, or is this a client-side concern?

Client-side concern, position is just a normalized 0-100 number(Both Width and height), so at backend-side there is nothing to transform.

> * are stamps contents unlimited in size? not that they _need_ to be limited ig

It's just a string, so the content is theoretically only limited by Redis natural limits(And I think URL size, since we send the value as a query param), we can set a limit, but that would depend on what people are using it for. I mean, I don't see why someone will leave a marker larger than a tweet, unless you're using it for a professional environment.

> * stamps don't intersect with any feature, such as category, tankoubon, or table of contents?

No, they are their own set of tables, and its own attribute, they do not share anything else than the archive_id.

lib/LANraragi/Model/Stamp.pm
Difegue: The cache/indexes are ephemeral by nature, so I wouldn't put data in those that can't be obtained in any other fashion. The archive ID <-> stamp link is essential, so to me they need to belong in the archive object.  
(although the stamps themselves do not, all the archive essentially keeps is a list of IDs)

I'm pretty sure we do want stamps to be included in backups, right? It's user-provided data and will likely be used in a variety of ways.

Difegue: (I'm resolving this but feel free to unresolve or write a separate comment if you have additional thoughts!)

EfronC:
Added a little cleanup code to remove the stamps when the archive gets deleted. I guess we are only missing to add the stamps to the backup file.

psilabs-dev:
No request for change. Just nailing things down so I'm not testing or expecting the wrong things.

> position is just a normalized 0-100 number

An integer or a float? I'm considering pages that exceed 1080p, 100 integer is a bit coarse.

> We check that the archive exists beforehand, so the first case is not possible

Is the archive check/stamp write exclusive? If not, it is still possible (toctou), though infrequent and carries risk. (Though tbh if it is confirmed, I wouldn't bother fixing it...)

> mean, I don't see why someone will leave a marker larger than a tweet

If I were ambitious: I might use stamps for an OCR/translation data storage layer which can be rendered frontend to override original JP display on the fly.

This may be stored as a JSON stamp carrying raw/translated data, which is interpreted by client-side customizations. So idk, it might not always be a tweet's length

---

Also agree, stamps are dependent on the archive, if an archive goes then there's no point in keeping the stamp around.

EfronC:
Ok, then I should be a little more specific with my answers:

> > position is just a normalized 0-100 number
> 
> An integer or a float? I'm considering pages that exceed 1080p, 100 integer is a bit coarse.

It is a float, calculated with the formula (x/image_width)*100, which is the same for the height but with y coordinate, and I think I'm not fixing the number of decimals. The idea here is that we calculate the proportional distance to top left of the image(Plus a minor fix due to HTML things), this way the stamp will be placed at the right place whether you size up or down the image.

> > We check that the archive exists beforehand, so the first case is not possible
> 
> Is the archive check/stamp write exclusive? If not, it is still possible (toctou), though infrequent and carries risk. (Though tbh if it is confirmed, I wouldn't bother fixing it...)

No, we do not set a lock for the archive itself, so it's theoretically possible to do the check, just after that the archive gets deleted, and then cause an error because after creating the stamp, we have to add it the stamps attribute of the first. I'm not sure how possible is this scenario is without a machine doing it.

> > mean, I don't see why someone will leave a marker larger than a tweet
> 
> If I were ambitious: I might use stamps for an OCR/translation data storage layer which can be rendered frontend to override original JP display on the fly.
> 
> This may be stored as a JSON stamp carrying raw/translated data, which is interpreted by client-side customizations. So idk, it might not always be a tweet's length

Yeah, that's why I didn't set a limit, my comment was more about that, for a casual reader, the bigger use-case would be to save an authors tweet that references a page, but otherwise, only people doing professional things with the platform will make use of the unlimited space.

EfronC:
I guess with this commit all topics related to stamps for the backend should be completed.

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,320 @@
...
+    }
+    $redis->hset( $archive_id, "stamps", $stamps );
+
+    $redis->quit();
+
+    return ( $key, $err );
+}
+
+# update_stamp(id, content, position)
+#   Removes the stamp from the page.
```
Difegue:
```suggestion
#   Update the stamp's content and position.
```

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,320 @@
...
+    my @allowed_keys = ( 'content', 'position' );
+    my %stamp = $redis->hgetall( $stamp_id );
+    ( $_ = redis_decode($_) ) for ( $stamp{content}, $stamp{position} );
+    %stamp = filter_hash_by_keys( \@allowed_keys, %stamp );
+    $stamp{id} = $stamp_id;
+
+    return %stamp;
+}
+
+# Converts an array of stamp registers to an array ob objects
```
Difegue:
```suggestion
# Converts an array of stamp registers to an array of objects
```

tools/openapi.yaml
```diff
@@ -3783,5 +4028,31 @@ components:
...
+      required:
+        - position
+        - content
+      properties:
+        id:
+          type: string
+          description: ID of the stamp
+        position:
+          type: string
+          description: Position of the stamp in the page in pixel coordinates
```
Difegue: Since the comments above mention 0-100, I guess it's not actually pixel coordinates anymore?  

At this point I guess it actually entirely depends on the frontend, since there's no server-side bounds checking when updating position values.

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,132 @@
...
+sub add_stamp {
+
+    my $self    = shift->openapi->valid_input or return;
+    my $id      = $self->stash('id');
+    my $index    = $self->stash('index');
+    my $content = $self->req->param('content') || "";
+    my $position = $self->req->param('position') || "";
+
+    unless ( defined $index ) {
+        return render_api_response( $self, "add_stamp", "Archive page." );
```
Difegue:
```suggestion
        return render_api_response( $self, "add_stamp", "No specified page for the stamp to attach to." );
```

lib/LANraragi/Model/Archive.pm
```diff
@@ -393,6 +393,22 @@ sub delete_archive ($id) {
...
+    if ( $redis->hexists( $id, "stamps" )) {
+        eval { @stamps = @{ decode_json($stamps) } };
+        if ($@) {
+            die;
+        }
+        foreach my $stamp ( @stamps ) {
+            $redis->del($stamp);
+        }
+    } else {
+        # Stamps attribute was not setted, do nothing.
```
Difegue:
```suggestion
        # Stamps attribute was not set, do nothing.
```

lib/LANraragi/Controller/Api/Stamp.pm
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    my ( $created_id, $err ) = LANraragi::Model::Stamp::add_stamp( $id, $index, $content, $position );
```
Difegue: I'm fully aware this is the consequences of my own actions but -- Given adding/removing stamps will now update the archive hash because of the stamp table, do we now need a lock on `archive-write` here?  

Might be the same for update/delete_stamp now too.

lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,320 @@
...
+
+    my %stamp = convert_stamp_to_object( $redis, $stamp_id );
+
+    $redis->quit();
+
+    return ( \%stamp, $err );
+}
+
+# get_stamps_by_page(id, page)
+#   Gets the list of pages that have at least one stamp.
```
Difegue:
```suggestion
#   Gets the list of stamps that are associated with this Archive ID and page.
```

Difegue (changes requested):
Mostly wording/doc nits left, I'm happy with this overall.  

Please update architecture.md with the finalized DB structure for stamps as well pls 🙏

tools/openapi.yaml
EfronC: I just clarified that is a normalized coordinate.

lib/LANraragi/Controller/Api/Stamp.pm
EfronC: This took a little more work than expected, so you should verify the way I implemented it is the right one for the DELETE. The UPDATE doesn't need it, because we do not modify the archive in that one.

Difegue (approved):
:shipit: 👍   

Nice job!

PR #1495, opened 2026-03-28
Bump alpine, replace Redis with Valkey, drop leftover wsl code and cleanup Dockerfile

Guerra24:
- Bump alpine to 3.23.
- Replace Redis 7 with Valkey 9.
- Remove legacy WSL code.
- Cleanup and reorganize Dockerfile based on how probable are files to change.
- Cleanup alpine dependencies.
- Fix code that generates warnings with Perl 5.42.
- Use Crypt-DES patch in Docker builds.
- Drop AVX libffi patch (needs testing on an affected system).
- Bump all actions to get rid of the Node v20 warning.


Difegue (approved):
rip wsl and avx patch you will be missed (but not that much)

PR #1505, opened 2026-03-31
Migrate nHentai plugin to API

### LINKED ISSUE #1529: Nhentai has modified its cookie rules and no longer provides a csrftoken, which in turn causes the nHentai CF Bypass extension to produce decoding errors for the scraped information
Now no matter how I try to fill in new cookie content, the extension returns the following error:

 Error: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 at /usr/local/share/perl5/site_perl/Mojo/[JSON.pm](https://json.pm/) line 38.

Guerra24:
Basic migration but ideally it should be improved with the use of a user token/api key. The api is public, at least initially, so it should work as-is for the time being.

~~I didn't really check it that much so I don't know what benefit we get from using user tokens and implementing the whole auth workflow over asking users to generate an api keys so that is something we can discuss later.~~


Guerra24:
Added the api key auth plugin replacing the cf bypass one.

> Please set a descriptive User-Agent header: AppName/version (contact or project URL). This helps us identify traffic and reach out if needed.

Also set the user-agent as requested by them.

Technically the API key is still not required as per their notes:

> Auth: Public (optional User Token or API Key for personalization)

PR #1509, opened 2026-04-03
Check if download body-size matches Content-Length header

rtldg:
If a connection drops during a download (e.g. because of a proxy or fucky server) then we can have a truncated body without a good way to check if there was an error.
[Mojo::Response::is_error](https://docs.mojolicious.org/Mojo/Message/Response#is_error) wasn't useful nor did [Mojo::Message::error](https://docs.mojolicious.org/Mojo/Message#error) seem to be anything, so here's a check for the `Content-Length` (if it exists) which can help sometimes.

### DIFF
--- a/lib/LANraragi/Model/Upload.pm
+++ b/lib/LANraragi/Model/Upload.pm
@@ -215,6 +215,12 @@ sub download_url ( $url, $ua ) {
         die( "No valid Content-Disposition header received after 5 attempts, aborting. (Last result: " . $tx->result->body . ")" );
     }
 
+    my $content_length = $tx->result->headers->content_length;
+    my $body_size = $tx->result->body_size;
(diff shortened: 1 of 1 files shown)


rtldg:
Realized the `Content-Length` header wouldn't equal the decompressed body's size if the response body is sent with compression/accept-encoding/gzip.
Luckily, it seems like mojo [updates the `Content-Length` header after decompression](https://github.com/mojolicious/mojo/blob/19fc4f19a0d83204a458ae4a19d192b7eaf4ba81/lib/Mojo/Content.pm#L200-L201).
In general though, download servers aren't going to be serving zips/rars/etc with gzip compression, so I guess that is a non-issue anyway.

Difegue:
Looks about right. Thanks!

Issue #1146, opened 2024-12-31
[Feature request] Disable/hide metadata plugin


ResRipper:
**Please describe your suggestion, and the problem it'd solve.**

Add a toggle to disable/hide a plugin so it doesn't appear in the "Import tags from plugin" list.

Currently there are 21 options shown in the list, and I think normal users will only use 4 or 5 of them on a regular basis. As there will be more plugins to come, it is important to allow the user to hide plugins, resulting in a much cleaner interface.

Difegue:
Long-term the idea is more to stop bundling all plugins with the core server and to rely on an external repo instead; It's currently not very efficient to have to wait for a new server release to get a bunch of plugin fixes.  

Only issue is that requires a bunch of infrastructure work so it hasn't been done yet :^)

ResRipper:
Each plugin is a single file I assume (not familiar with Perl), so the easiest way might be to use a GitHub repo to store plugins and download directly from it; you could even put a description file for all plugins in the repo, so no need to bundle plugin info with the server.

IceBreeze:
Another way could be to add a fixed first option to the "select" menu, with a generic text like "Parse URL...", that when called checks if exists a plugin that can work with the specified URL.

The plugins should expose a method that must be called passing an URL and returns true or false.
Also it might be useful to add a new plugin category ("scrapers"?) to easily exclude these plugins from the "select" menu.

Difegue:
i gave some thought to a plugin repo multiple times in the past, but the issue comes with handling updates and making sure the plugins remain downloaded even for setups like docker users who trash the entire container on updates 🤔

the plugin repo would need an index file generated by CI or smth that lists the hosted plugins, their versions and download URLs; Then we could check that against the on-server plugins, which would need to be stored in a directory outside of the server/container filesystem..

krakerz:
> i gave some thought to a plugin repo multiple times in the past, but the issue comes with handling updates and making sure the plugins remain downloaded even for setups like docker users who trash the entire container on updates 🤔
> 
> the plugin repo would need an index file generated by CI or smth that lists the hosted plugins, their versions and download URLs; Then we could check that against the on-server plugins, which would need to be stored in a directory outside of the server/container filesystem..

i don;t know how perl behaviour (i think it was perl)
but can't it be done like disable / hide it from frontend only?
so the update handling is still the same as before
or in perl, it need to be deleted / removed from the plugins directory for proper disable / hide ?

[...]

PR #1511, opened 2026-04-06
Registry and Plugin (Backend)

psilabs-dev:
- Issue: https://github.com/Difegue/LANraragi/issues/1146
- Plugins demo registry repo: https://github.com/psilabs-dev/lrr-plugins-demo

Under testing, lots of code may be subject to change. There are quite a number of avenues where code might break. I don't recommend reviewing this code atm.

This backend-only PR ~~removes all (non-local) plugins from LRR~~, and implements the registry system and plugin management.

- Multiple registry support
- Management features, such as installing/uninstalling/upgrading plugins, plugin conflict resolution, management APIs.
- Support for installing/uninstalling managed plugins
- Redis is now responsible for plugin package presence management (previously INC's). Which makes uninstall/install easier.
- ~~Configuration-related features, such as metadata plugin priority, enablement, and dropdown hiding. Add metadata plugin priority support.~~

Obligatory security warning, as downloading untrusted arbitrary files and allowing code execution from the internet requires extra care which was not a consideration when plugins were the responsibility of LRR. The boundary for this PR is that LRR will not audit or review files/registries/licenses for the user, the user must trust registries and audit them personally/take responsibility for any bugs/vulnerabilities within plugins.

Some examples of ways to be compromised, which LRR cannot protect against: registry URL typos/typo squatting, DNS poison/certificate updates, hijacked registries/supply chain attacks

(tbh, with this I'm also just wondering whether plugin installations should require a restart altogether)

(PS: maybe we should support uninstalling sideloaded plugins?)


psilabs-dev:
Updated this PR so it doesn't touch existing plugins, that's probably a separate concern to decide what to add/remove from LRR.

psilabs-dev:
An update on duplicate plugins. There are many ways duplicates can happen with the entire plugin system, and the existing dev logic doesn't really handle it besides choosing a survivor and moving on. Now that a subset of plugins have crud, a duplicate *does* happen, then we might also run into myriad of other unforseen issues.

At the moment, I don't think this PR will guarantee graceful duplicate handling, or handling of issues caused by duplicate plugins.

LRR will perform plugin discovery on startup and report duplicates at warning level, and this PR will ensure that LRR plugin APIs do not *introduce* duplicates into the system. But past startup, LRR assumes a clean plugin state, and may not handle cases where a duplicate existed or is introduced by the user via other means.

(unless loc budget/complexity isn't a concern)

psilabs-dev:
Removed channels in favor of semver, all registry plugin versions will be enforced at "x.x.x" style, and LRR will resolve installations by max semver version. Users can opt to install specific versions.

Added stack-based rollback on install failure. If LRR fails to upgrade a plugin, for example, it's going to attempt to restore the previous plugin (which is an existing stance iirc). This is a one-time rollback; if the rollback itself fails, then we're not going to "rollback the rollback", instead return 500 error.

Probably the (remaining) big item is where to place sideloaded plugins, I'll think about that after the local/git registry system is stable.

Commit history is mainly for tracking what's happened. Once the pr matures (after I do another pass), I'll do a rebase and squash.

psilabs-dev:
The controller layer and OpenAPI contract are fixed, with exception of maybe some cosmetic changes but I won't be touching controller files anymore (unless something's really wrong). So this branch is free to api test.

Next up will be improving model/utils-side of things. A lot of the code is validation and exception handling work, not exactly business logic, but I'll have to see what exactly needs changing instead of just moving things around.

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                properties:
+                  operation:
+                    type: string
+                    enum:
+                      - get_default_registry
+                  success:
+                    type: integer
+                    enum:
+                      - 1
+                  registry_id:
```
Guerra24: Ideally this should be simply `id`, just like the response for registry creation.

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                properties:
+                  operation:
+                    type: string
+                    enum:
+                      - remove_default_registry
+                  success:
+                    type: integer
+                    enum:
+                      - 1
+                  registry_id:
```
Guerra24: ditto

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                properties:
+                  operation:
+                    type: string
+                    enum:
+                      - update_default_registry
+                  success:
+                    type: integer
+                    enum:
+                      - 1
+                  registry_id:
```
Guerra24: ditto

tools/Documentation/api-documentation/plugin-api.md
```diff
@@ -5,13 +5,25 @@ description: APIs to list and execute Plugins.
 # Plugin API
 
 {% openapi-operation spec="lanraragi-api" path="/plugins/{type}" method="get" %}
-[OpenAPI lanraragi-api](https://raw.githubusercontent.com/Difegue/LANraragi/refs/heads/dev/tools/openapi.yaml)
+[OpenAPI lanraragi-api](https://raw.githubusercontent.com/psilabs-dev/LANraragi/refs/heads/dev-registry/backend/tools/openapi.yaml)
```
Guerra24: Url got changed

tools/Documentation/api-documentation/registry-api.md
```diff
@@ -0,0 +1,41 @@
+---
+description: APIs to manage Plugin Registries.
+---
+
+# Registry API
+
+{% openapi-operation spec="lanraragi-api" path="/registries" method="get" %}
+[OpenAPI lanraragi-api](https://raw.githubusercontent.com/psilabs-dev/LANraragi/refs/heads/dev-registry/backend/tools/openapi.yaml)
```
Guerra24: Same url

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                    type: integer
+                    enum:
+                      - 1
+                  name:
+                    type: string
+                  namespace:
+                    type: string
+                  version:
+                    type: string
+                  installed_registry:
```
Guerra24: I don't think we need to prefix this with `installed_`

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                    type: string
+                    description: Registry ID the plugin was installed from.
+                  installed_sha256:
```
Guerra24: ditto

tools/openapi.yaml
```diff
@@ -3872,6 +4665,25 @@ components:
...
           nullable: true
           description: Namespace of the login plugin this plugin depends on
+        hidden:
+          type: boolean
+          description: Whether the plugin is hidden from UI dropdowns
+        priority:
+          type: integer
+          minimum: 0
+          description: Execution priority for metadata plugins. Lower values run first.
+        installed_registry:
```
Guerra24: ditto

tools/openapi.yaml
```diff
@@ -3872,6 +4665,25 @@ components:
...
+          type: string
+          nullable: true
+          description: Registry ID the plugin was installed from, or null if not registry-managed
+        installed_version:
```
Guerra24: ditto

tools/openapi.yaml
```diff
@@ -3872,6 +4665,25 @@ components:
...
+          type: string
+          nullable: true
+          description: Installed version recorded in Redis, or null if not registry-managed
+        installed_sha256:
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+        }
+        return;
+    };
+
+    # Backup existing artifact so a same-path upgrade has bytes to restore on rollback.
+    my $backup_path;
+    if ( -e $installpath ) {
+        $backup_path = "$installpath.lrr-rollback";
+        unlink_path($backup_path) if -e $backup_path;
+        unless ( rename_path( $installpath, $backup_path ) ) {
```
Guerra24: `rename_path` and friends should only be used with archives, was there a reason for it here?

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+            my $err = "$!";
+            $logger->error("Cannot back up existing artifact at $installpath: $err");
+            return ( 500, undef, "Cannot back up existing artifact for transactional install: $err" );
+        }
+        push @undo, [
+            "restore prior artifact from $backup_path",
+            sub { rename_path( $backup_path, $installpath ) ? undef : "$!" },
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+            }
+        }
+        return;
+    };
+
+    # Backup existing artifact so a same-path upgrade has bytes to restore on rollback.
+    my $backup_path;
+    if ( -e $installpath ) {
+        $backup_path = "$installpath.lrr-rollback";
+        unlink_path($backup_path) if -e $backup_path;
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+        if ( my @resp = $do_rollback->("Cannot write plugin file: $err") ) {
+            return @resp;
+        }
+        return ( 500, undef, "Cannot write plugin file during installation: $err" );
+    }
+    push @undo, [
+        "unlink artifact at $installpath",
+        sub {
+            return unless -e $installpath;
+            unlink_path($installpath) ? undef : "$!";
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+        delete $INC{$install_relpath}; # clear out undef resulting from require failure
+        if ( my @resp = $do_rollback->("Plugin '$namespace' failed to load: $require_error") ) {
+            return @resp;
+        }
+        return ( 422, undef, "Plugin '$namespace' failed to load: $require_error" );
+    }
+
+    @undo = ();
+    if ( $backup_path && -e $backup_path ) {
+        unlink_path($backup_path) or $logger->warn("Could not remove rollback backup at $backup_path: $!");
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+    }
+
+    # If the upgrade landed at a new path (type-change between published versions, in violation
+    # of spec invariance), remove the old artifact so scan_plugins doesn't rediscover it.
+    if ( defined $currentpath && $currentpath ne $installpath && -e $currentpath ) {
+        unlink_path($currentpath) or $logger->warn("Could not remove stale plugin file at $currentpath: $!");
```
Guerra24: ditto

lib/LANraragi/Model/Plugins.pm
```diff
@@ -314,4 +320,549 @@ sub has_old_style_params (%params) {
...
+
+    # Delete the plugin file (only if it's actually inside LRR lib)
+    if ( -e $installpath ) {
+        my $canonpath = abs_path($installpath);
+        my $plugindir = abs_path( getcwd() . "/lib/LANraragi/Plugin" );
+        unless ( $canonpath && $plugindir && index( $canonpath, "$plugindir/" ) == 0 ) {
+            return ( 403, undef, "Can't delete plugin outside Plugin/ directory: $installpath" );
+        }
+
+        unlink_path($canonpath) or do {
```
Guerra24: ditto

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                    enum:
+                      - get_registry
+                  success:
+                    type: integer
+                    enum:
+                      - 1
+                  error:
+                    type: string
+                    nullable: true
+                  id:
```
Guerra24: id is duplicated in the metadata

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                  format: uri
+                  pattern: "^https://"
+                  description: Git repository HTTPS URL. Required when type is 'git'.
+                ref:
+                  type: string
+                  description: Git ref (branch, tag, or commit). Defaults to 'main'.
+                path:
+                  type: string
+                  minLength: 1
+                  description: Local filesystem path. Required when type is 'local'.
```
Guerra24: Wouldn't it be better to have both `type` and `provider` merged and reuse url for both instead of explicit types? I don't see a reason for two fields that do more or less the same thing.

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+              type: object
+              required:
+                - name
+                - type
+              properties:
+                name:
+                  type: string
+                  minLength: 1
+                  description: User-friendly alias for this registry.
+                type:
```
Guerra24: Throws 500 error if the type is not exact aka lowercase.

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                    - gitea
+                  description: Git hosting provider. Required when type is 'git'.
+                url:
+                  type: string
+                  format: uri
+                  pattern: "^https://"
+                  description: Git repository HTTPS URL. Required when type is 'git'.
+                ref:
+                  type: string
+                  description: Git ref (branch, tag, or commit). Defaults to 'main'.
```
Guerra24: Empty strings will be accepted, is this intended?

tools/openapi.yaml
```diff
@@ -554,6 +556,797 @@ paths:
...
+                  pattern: "^[a-z0-9_-]+$"
+                  description: Namespace of the plugin to install. Registry namespaces are lowercase.
+                registry:
+                  type: string
+                  description: Registry ID to install from.
+                  minLength: 14
+                  maxLength: 14
+                version:
+                  type: string
+                  minLength: 1
```
Guerra24, line 1193: Is this working? Empty string got accepted.

lib/LANraragi/Utils/Registry.pm
```diff
@@ -0,0 +1,302 @@
...
+# $match_fn: coderef($filepath, $fh) -> bool; return true if filepath conflicts.
+sub _find_conflict {
+    my ( $skip_path, $match_fn ) = @_;
+
+    my $plugin_dir = getcwd() . "/lib/LANraragi/Plugin";
+    my $conflict;
+
+    return unless -d $plugin_dir;
+
+    find_path(
```
Guerra24: Same as the other path methods

lib/LANraragi/Model/Registry.pm
```diff
@@ -0,0 +1,462 @@
...
+    # Determine if source fields are changing
+    my $indexcleared = 0;
+    foreach my $field (@SOURCE_FIELDS) {
+        next unless exists $updated_registry{$field};
+        if ( !defined $current_registry{$field} || $current_registry{$field} ne $updated_registry{$field} ) {
+            $logger->info("Source field '$field' changed on '$registry_id'; will clear cached index.");
+            $indexcleared = 1;
+            last;
+        }
+    }
```
Guerra24: path with null value always triggers a clear cache

lib/LANraragi/Model/Registry.pm
```diff
@@ -0,0 +1,462 @@
...
+    end
+    while idx + 1 <= #ARGV do
+        redis.call("HSET", KEYS[1], ARGV[idx], ARGV[idx + 1])
+        idx = idx + 2
+    end
+    if KEYS[2] ~= "" then
+        redis.call("DEL", KEYS[2])
+    end
+    return 1
+LUA
```
Guerra24: I guess this updates only the fields that changed? A comment would be nice

Guerra24 (changes requested):
I don't see a way to list available plugins by registry, is it part of another pr or I'm just blind?

I have a feeling empty strings will cause issues in the api.

Code looks fine maybe a bit overly verbose at least compared to the rest.

I will do a second pass when I get to test it on Windows.

tools/Documentation/api-documentation/plugin-api.md
psilabs-dev: will be reverted. was trying to debug the gitbook problem

psilabs-dev:
> I don't see a way to list available plugins by registry, is it part of another pr or I'm just blind?

No this isn't added

tools/openapi.yaml
psilabs-dev: Pretty sure this is added for a reason, will discuss this later.

tools/openapi.yaml
psilabs-dev: Can't reproduce. Was this with validation enabled or disabled?

psilabs-dev:
One note: any request/response validation that can (and should) be done by OpenAPI won't be re-done by LRR. So this PR assumes that OpenAPI validation is enabled for any testing.

tools/openapi.yaml
Guerra24: Enabled. Req and error.

```json
{"name":"Demo","type":"Git","provider":"github","url":"https://github.com/psilabs-dev/lrr-plugins-demo","ref":"main","path":null}
```
```
Can't use an undefined value as an ARRAY reference at /home/koyomi/lanraragi/script/../lib/LANraragi/Model/Registry.pm line 62.
```

tools/openapi.yaml
psilabs-dev: updated

lib/LANraragi/Model/Plugins.pm
psilabs-dev: not really, moved back

tools/openapi.yaml
psilabs-dev: gone

tools/openapi.yaml
psilabs-dev: Resolving this as an OpenAPI validation bug.

tools/openapi.yaml
psilabs-dev: Could you double-check the validation issues? minLength should've blocked it.

tools/openapi.yaml
psilabs-dev: can you confirm minLength is applied here? I've added it but don't see it on github.

psilabs-dev:
Removed the metadata plugin features as I think we already have enough to go over as is

lib/LANraragi/Model/Registry.pm
psilabs-dev: Yeah it's not very readable, the model layer will need a rework.

tools/openapi.yaml
psilabs-dev: I don't think they should be merged. Type is a transport, provider is a git-specific templating rule (which is also optional if we pass "local" type).

If we merge them, how do we resolve between different providers, esp. if you have self-hosted gitlab or gitea domains, LRR won't know what to resolve to, unless you're planning on something like "gitlab://https://mygitlab.com/org/repo"

psilabs-dev:
basic cdn support is added

tools/openapi.yaml
Guerra24: I was thinking more in the sense of having a Provider interface that then providers implement where they do their thing and return the final composed url/path, then just map them to their type (local, gitlab, github, cdn, etc.) instead of having 2 layers of ifs and explicit checks for each one.

tools/openapi.yaml
Guerra24: Pretty sure is the validation bug.

tools/openapi.yaml
Guerra24: Validation bug.

tools/openapi.yaml
psilabs-dev: On second look you're right; removed

tools/openapi.yaml
psilabs-dev: Works for me. merged

lib/LANraragi/Model/Registry.pm
psilabs-dev: this logic is removed; cache clears on any update

lib/LANraragi/Model/Registry.pm
psilabs-dev: Added a note on `my @fields_to_set`

psilabs-dev:
Model layer (and above) are now reviewable (assuming the tests pass). Next up will be improvements to the Utils-layer.

Guerra24 (commented):
Looks fine to me, tested it on Windows and it works as expected. Model layer looks cleaner and is easier to follow.

psilabs-dev:
The Utils layer (Plugin/PluginState/Registry) should be stable now. Going to start dogfooding on this with a gitlab/codeberg account + self hosted gitea.

Unit and integration tests are still AI written, maybe we'll go through another round of mutation testing before I open it up for review.

psilabs-dev:
squashed, backup at https://github.com/psilabs-dev/LANraragi/pull/new/dev-registry/backend-backup

psilabs-dev:
Checked that registry on self-hosted gitea can be read by LRR (after some cert/dns shenanigans).

psilabs-dev:
Removed inter-worker plugin synchronization logic and added an `/api/server/status` endpoint which carries whether the server should restart.

Also moved plugin loadability out of process, so it doesn't contaminate the process INC cache via require. Should work on both windows via Karen's `perl` guarantees, and in unix via the Config approach.

lib/LANraragi/Model/Plugins.pm
```diff
@@ -20,9 +24,22 @@ use LANraragi::Utils::Generic  qw(exec_with_lock_pure);
...
+    find_namespace_conflict
+    validate_registry_artifact_path
+    resolve_max_version
+    MANAGED_TYPE_DIRS
+);
+use LANraragi::Model::Registry;
+use LANraragi::Model::Server   qw(set_restart_pending);
+
+# Max plugin file size for slurp (files will/should never reach this size anyways but stops OOM)
+use constant MAX_PLUGIN_FILE_SIZE => 100 * 1024 * 1024;      # 100 MB
```
psilabs-dev: I think this should be 1mb, like what monster plugin is more than that

lib/LANraragi/Utils/Registry.pm
```diff
@@ -0,0 +1,321 @@
...
+my @REQUIRED_VERSION_FIELDS = qw(name author description artifact sha256 published_at);
+
+# Resolve a git URL to a raw file URL for a given provider
+# (Should) support github, gitlab, and gitea/codeberg providers but who knows
+sub resolve_git_raw_url {
+    my ( $provider, $url, $ref, $path ) = @_;
+
+    my $logger = get_logger( "Registry", "lanraragi" );
+
+    # TODO: this just needs to be tested more (maybe with a Gitlab + Gitea repo)
```
psilabs-dev: only tested on gitea, I'm not able to test on gitlab

psilabs-dev (commented):
dobfood round 2

tools/openapi.yaml
```diff
@@ -385,7 +387,30 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ServerInfo'
-  
+
+  /server/status:
```
Difegue: I think this could just be part of /api/info like the other server preferences.

tools/Documentation/api-documentation/plugin-api.md
```diff
@@ -15,3 +15,11 @@ description: APIs to list and execute Plugins.
 {% openapi-operation spec="lanraragi-api" path="/plugins/queue" method="post" %}
 [OpenAPI lanraragi-api](https://raw.githubusercontent.com/Difegue/LANraragi/refs/heads/dev/tools/openapi.yaml)
 {% endopenapi-operation %}
+
+{% openapi-operation spec="lanraragi-api" path="/plugins/install" method="post" %}
+[OpenAPI lanraragi-api](https://raw.githubusercontent.com/Difegue/LANraragi/refs/heads/dev/tools/openapi.yaml)
+{% endopenapi-operation %}
+
+{% openapi-operation spec="lanraragi-api" path="/plugins/installed/{plugin_namespace}" method="delete" %}
```
Difegue, line 23: I think the endpoint would be more sensible if we dropped the /installed  


```suggestion
{% openapi-operation spec="lanraragi-api" path="/plugins/{plugin_namespace}" method="delete" %}
```

tools/Documentation/SUMMARY.md
```diff
@@ -51,6 +51,7 @@
 * [Tankoubon API](api-documentation/tankoubon-api.md)
 * [Stamps API](api-documentation/stamp-api.md)
 * [Plugin API](api-documentation/plugin-api.md)
+* [Registry API](api-documentation/registry-api.md)
```
Difegue, line 55: I think we need a doc page for Registries, can't really only have the API here.  

I suppose docs can wait until the frontend makes it in, but I think we should write up the concepts and the types of supported registries already.

lib/LANraragi/Controller/Plugins.pm
```diff
@@ -141,10 +141,6 @@ sub save_config {
 
             } elsif ( ref( $pluginfo->{parameters} ) eq 'HASH' ) {
 
-                # TODO: remove this line (and the ARRAY check above)
```
Difegue, line 144: I get that this line is being removed because $namerds now also stores the registry information for the plugin, but we do still want to deprecate array parameters at some point... Maybe with the move to registries actually.

tools/Documentation/SUMMARY.md
Difegue: (also an archi doc update for the new redis structures)

lib/LANraragi/Controller/Api/Plugins.pm
```diff
@@ -0,0 +1,130 @@
...
+# Install a managed plugin.
+sub install_plugin {
+    my $self        = shift->openapi->valid_input or return;
+    my $body        = $self->req->json;
+    my $namespace   = $body->{namespace};
+    my $registry_id = $body->{registry};
+    my $version     = $body->{version};
+    my $force       = $body->{force} // 0;
+
+    # TODO: maybe consider extending TTL for this sub to 60s if it's not enough.
```
Difegue: Possible to throw the actual install portion to a minion job tbh, I suppose that's a decision we should make now as it'd change the API signature to return a jobId

lib/LANraragi/Model/Plugins.pm
Difegue: yeah I don't see a world where we'd accomodate 100MB plugins; The only cases I can think of are plugins that bundle a tag database or a ML model, but plugins that rely on that should just get users to indicate the path to the DB/model in the plugin's config.

lib/LANraragi/Utils/Plugins.pm
```diff
@@ -166,18 +256,78 @@ sub get_plugin_parameters {
...
+}
+
+# Unregister a registered plugin from the database.
+# Should be called during removal of a plugin or when the plugin
+# could no longer be found.
+sub unregister_plugin {
+    my ( $redis, $namespace ) = @_;
+
+    my $namerds = "LRR_PLUGIN_" . uc($namespace);
+    $redis->hdel(
```
Difegue, line 282: If we deprecate array params, we could change this to just delete the namerds key outright, right?  

Might be worth adding a TODO here if I'm not mistaken.

Difegue (changes requested):
Just a few misc things, I think we can undraft this otherwise.  

I think the `needs_restart` bits should be coordinated with https://github.com/Difegue/LANraragi/pull/1673 since it feels like there's some overlap there. (At the very least, I suppose we should provide an API to restart the server)

lib/LANraragi/Utils/Registry.pm
Difegue: Might be worth disabling gitlab for now if we can't test it, I don't really see repos being over there anyway tbh (+ their CEO is a moron and unlike github they don't have the critical mass to justify us supporting it)

tools/openapi.yaml
psilabs-dev: merged

psilabs-dev:
> I think the needs_restart bits should be coordinated with https://github.com/Difegue/LANraragi/pull/1673 since it feels like there's some overlap there. (At the very least, I suppose we should provide an API to restart the server)

this probably depends on merge order (but weakly). This PR only signals whether LRR needs to restart, but hitagi actually provides (or should provide) the restart APIs which the client can act on.

Following servers like jenkins, LRR should not force a restart after installations, and instead advise the user to manually trigger one. It's only if this statement were reversed that we'll have two merge order cases to consider.

lib/LANraragi/Controller/Api/Plugins.pm
psilabs-dev: done with https://github.com/Difegue/LANraragi/pull/1511/changes/c51b24270a53510cab4a34cfd8e8e8626e198fe0. I'd note that mojo has its own job locking with guard, but it talks past `exec_with_lock` so I don't recommend it

tools/Documentation/api-documentation/plugin-api.md
psilabs-dev: DELETE "/plugins/{plugin_namespace}" and GET "/plugins/{type}" are indistinguishable param-wise (wrt openapi), we can have either one or the other as the methods live under the same URL

Alternatively, we could lift the managed plugins to the "/plugins/installed" URL, e.g. for installing plugins

lib/LANraragi/Utils/Plugins.pm
psilabs-dev: If we just delete the namerds key we would also nuke previous configurations/settings (like enabled for metadata). Which I'm personally fine with (users just need to reconfigure stuff and you're kinda asking for it by uninstalling), but this could be a UX decision.

lib/LANraragi/Controller/Plugins.pm
psilabs-dev: I'd say that's something we can do.

Also this block was removed bc it clears all plugin data on config save (including prov), but that might be unclear, so I just reverted to commenting the whole thing out.

tools/Documentation/SUMMARY.md
psilabs-dev: Added some, though not sure how the docs should be presented with the UI missing 🤔 also idk how technical the docs should be...

psilabs-dev:
Moved default registry to ougi

remaining CI failure is addressed by https://github.com/Difegue/LANraragi/pull/1677

Difegue (approved):
There are a few potential nits left but I'll handle these when making the default registry repo and UI, it's best to get the change in at this point.  

sorry for the delay i've been at war

PR #1516, opened 2026-04-08
Center image vertically in fullscreen mode

### LINKED ISSUE #1360: Center reader/images vertically
**Please describe your suggestion, and the problem it'd solve.**
When viewing on a portrait screen (mobile devices, vertical monitors, split screen etc.), the reader is aligned to the top of the screen regardless of the "fit display to" setting. This is especially noticable with landscape images, as well as in full screen mode for all images, where there will be a large empty section below the image. Not only does it look odd in my opinion, it also makes navigating through landscape images cumbersome on mobile devices since everything is at the top of the screen rather than the center.

I suggest either making the reader vertically centered by default, or adding a setting that would achieve this.

pneumaticity:
Fixes #1360 partially (only for fullscreen mode, the other case is harder).

### DIFF
--- a/public/css/lrr.css
+++ b/public/css/lrr.css
@@ -795,6 +795,10 @@ div.fullscreen img {
   justify-content: center;
 }
 
+.fullscreen #display {
+    height: 100%;
+}
+
 .fullscreen-infinite {
   overflow-y: scroll;
 }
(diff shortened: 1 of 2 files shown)


Difegue:
Straightforward enough, yep. Thanks for this!

PR #1517, opened 2026-04-08
Stamps Frontend

### LINKED ISSUE #1355: [Feature Request] O-Counter / Nut-Button

EfronC:
Just moved the Frontend code from #1493 to here. Will leave it as a draft until the other one merges.


lib/LANraragi/Model/Stamp.pm
```diff
@@ -0,0 +1,254 @@
...
+        $position = $c_content[0]
+    }
+
+    if ( defined $content ) {
+        $content = remove_separator($content, "|");
+    } else {
+        $content = $c_content[1]
+    }
+
+    if ( $redis->exists($faves_id) ) {
```
psilabs-dev: Exist check comes too late. It should come before `$redis->hget($faves_id => $key);` and gate updates against nonexistent stamps.

See https://github.com/Difegue/LANraragi/blob/9d6fda1c9f4fcbe2eb3055b6c23cd7441a9a4d5c/lib/LANraragi/Model/Tankoubon.pm#L239

EfronC: This review should probably be moved to #1493 since this PR should be exclusively for the Frontend files.

psilabs-dev: Yep, just wondered why it was in draft and realized myb

EfronC:
For the pagenumber that we use for the id `STAMPS_<pagenumber>_<ts>`, Should it match Reader.currentPage, or the number shown here(Which is Reader.currentPage+1)?
image

public/js/reader.js
```diff
@@ -1354,3 +1700,8 @@ Reader.handlePaginator = function () {
 Reader.getFilename = function(index) {
     return new URLSearchParams(Reader.pages[index].split("?")[1]).get("path");
 }
+
+window.addEventListener("resize", () => {
+    // Reload the markers everytime the image size changes
+    Reader.renderMarkers();
```
Difegue, line 1753: This doesn't seem to account for changing the Reader fit options (container/width/height)  

Switching to fullscreen also seems to break stamp rendering, but I'm not sure if it makes sense to show stamps in FS to begin with.

locales/template/en.po
```diff
@@ -1153,6 +1153,18 @@ msgstr "FullScreen"
 
 msgid "Toggle Bookmark"
 msgstr "Toggle Bookmark"
+
+msgid "Toggle Stamps"
+msgstr "Toggle Stamps"
+
+msgid "Delete Archive"
```
Difegue: I'm aware there's currently (bad) precedent for it in the po files, but "Delete Archive" already exists as a key so there's no need to add it once more.

public/css/lrr.css
```diff
@@ -304,6 +304,38 @@ p#nb {
...
   cursor: pointer;
+  z-index: 22;
+}
+
+.marker {
+  position: absolute;
+  width: 24px;
+  height: 24px;
+
+  background-image: url('../favicon.ico');
```
Difegue, line 329: This is nice for a V1. One improvement I think we could toss in later would be to make the stamp icon configurable in Settings (probably as a base64 png or something)

public/js/reader.js
```diff
@@ -719,6 +799,222 @@ Reader.toggleHelp = function () {
     // all toggable panes need to return false to avoid scrolling to top
 };
 
+Reader.addStamp = function () {
```
Difegue, line 826: This should probably be tied to a keyboard shortcut as well with a matching line in the Help popup.  

Also, I can't help but think this needs some kind of extra UI florish to indicate the user that they're in marker mode.  
Maybe just changing the mouse pointer to `crosshair` would be enough?  

(I also like `cell` because it does the funny excel cursor but I don't know how that behaves in other OSes so...)

public/js/reader.js
```diff
@@ -719,6 +799,222 @@ Reader.toggleHelp = function () {
...
+            } else {
+                Reader.pageNaviState = true;
+                Reader.renderMarkers();
+            }
+        });
+    });
+    
+
+    // Delete
+    marker.addEventListener("contextmenu", (e) => {
```
Difegue: I take it this means right-clicking a stamp deletes it? It doesn't seem to work on my end :(

public/js/reader.js
```diff
@@ -719,6 +799,222 @@ Reader.toggleHelp = function () {
...
+
+    // Draw markers
+    Reader.markers.forEach((markerData, index) => {
+        Reader.createMarkerElement(markerData, index);
+    });
+}
+
+Reader.toggleStamps = function () {
+    // Show or hide the markers
+    Reader.markersVisible = !Reader.markersVisible;
```
Difegue: This should probably be saved to localStorage and restored on page reload

templates/reader.html.tt2
```diff
@@ -112,7 +114,9 @@
                     <h2>[% c.lh("Admin Options") %]</h2>
 
                     <input id="edit-archive" class='stdbtn' type='button' value="[% c.lh("Edit Archive Metadata") %]"/>
-                    <input id="delete-archive" class='stdbtn' type='button' value="[% c.lh("Delete Archive") %]"/>
+                    <input id="delete-archive" class='stdbtn' type='button' value="[% c.lh("Delete Archive") %]"/><br>
+                    <input id="set-stamp" class='stdbtn' type='button' value="[% c.lh("Set stamp") %]"/>
+                    <input id="filter-stamped" class='stdbtn' type='button' value="[% c.lh("Filter stamped pages") %]"/>
```
Difegue: This doesn't really belong in Admin Options, right? I'm assuming this would work without being authenticated.  

(Although tbh I'm not quite sure the function is useful.. I'd tend to cut it in favor of just adding an icon to the thumbnail for a page if it contains stamps)

templates/reader.html.tt2
```diff
@@ -302,6 +306,7 @@
 <div class="absolute-options absolute-left">
     <a class="fa fa-cog fa-2x" id="toggle-settings-overlay" href="#" title="[% c.lh('Reader Settings') %]"></a>
     <a class="fa fa-question-circle fa-2x" id="toggle-help" href="#" title="[% c.lh('Help') %]"></a>
+    <a class="fa fa-eye-slash fa-2x" id="toggle-stamps" href="#" title="[% c.lh('Toggle stamps') %]"></a>
```
Difegue: This might belong in the Reader Options subpanel instead? I'm not sure whether people would want stamp toggle often enough to warrant having it front-and-center.  

(Mostly because I'm concerned about cramming too many buttons in the Reader... It feels slightly over the line already)

Difegue (commented):
Tested this a little bit so this is mostly usability concerns atm. It's working pretty well already though 👍

locales/template/en.po
Difegue: One more thing -- looks like there's a capitalization mistake in the JS so it tries to fetch "Toggle stamps" instead of "Toggle Stamps".  

Since Stamps are a data object they should be capitalized everytime imo

public/js/reader.js
EfronC: I mean, the way you identify the marker mode is because the backgroud gets darkened by the overlay, with only  the image lighted, like a focus mode, not sure if it didn't trigger on your device. I think I tried to change the mouse pointer to crosshair, but I think something else was overwritting it.

public/js/reader.js
EfronC: Yes, it does. That's weird, maybe contextmenu works differently depending on the browser? I only tested on Brave.

public/js/reader.js
EfronC: Wouldn't this cause that, if you load the markers in one browser, go to another and edit one, the change will not be reflected on the first one on page change since it would be loading from localstorage instead of the DB?

templates/reader.html.tt2
EfronC: Not sure if I missunderstood the comment, but the filter for the thumbnails is because of UX, since that way you can only get the pages "Marked"(Which was the main use case when the feature got started), the alternative of having all the thumbs, and only recognizing the marked ones by scanning with your eyes the whole list of thumbnails to find the tiny marks... It makes my eyes tired just from thinking about it.

And yes, the filter button does not belong to admin options, I just put the button there not to break the UI more than I did. Have a recommendation where to add it?

public/css/lrr.css
EfronC: Base64... For some reason that makes me think of code injection issues hahaha. Yes the idea is that eventually this will allow to have custom marker icons, I just didn't want to add arbitrary ones myself.

locales/template/en.po
EfronC: Why is there a "Delete Archive" here? I don't remember adding this one.

public/js/reader.js
EfronC: Added support for the height and width modes, for fullscreen I just added a line that turns off the stamps.

public/js/reader.js
EfronC: Added a change in the cursor to use cell when on Marker mode.

public/js/reader.js
EfronC: Added an actual context menu, and put edit and delete in there. If this doesn't work, then is a problem of the browser.

public/js/reader.js
EfronC: Actually I didn't read correctly your comment, I moved it to LocalStorage.

templates/reader.html.tt2
EfronC: What do you think of this? Is just a simple example of tabs I found out there, and I'm testing on my local, maybe could help to deal with these extra buttons problem, second tab is not dependant of admin login. It would be more efficient to use jquery tabs, but I've no idea how to add that.

image
image

EfronC: One more image of how it would look like after playing a little with the CSS, ~~if you tell me this could work, then I'll upload this change~~ Added it anyway, since I think could be useful in the future. 

image

public/js/reader.js
```diff
@@ -525,6 +528,9 @@ Reader.initializeSettings = function () {
 
     Reader.containerWidth = localStorage.containerWidth;
     if (Reader.containerWidth) { $("#container-width-input").val(Reader.containerWidth); }
+    
+    Reader.markersVisible = JSON.parse(localStorage.markersVisible);
```
Difegue: This fails if `localStorage.markersVisible` is undefined

public/js/reader.js
```diff
@@ -1460,6 +1480,9 @@ Reader.toggleFullScreen = function () {
 };
 
 Reader.handleFullScreen = function (enableFullscreen = false) {
+    if (Reader.markersVisible && enableFullscreen) {
```
Difegue: This doesn't restore marker view when coming out of fullscreen unless you refresh the page

public/js/reader.js
```diff
@@ -1250,11 +1597,71 @@ Reader.updateArchiveOverlay = function (forceUpdate = false) {
...
+        Reader.overlayFiltered = false;
+        Reader.updateArchiveOverlay(true);
+    } else {
+        Server.callAPI(`/api/archives/${Reader.id}/stamps/`, "GET", null, I18N.ServerInfoError, 
+            (data) => {
+                $("#extract-spinner").hide();
+                let pages = data.result.sort();
+
+                // For each link in the pages array, craft a div and jam it in the overlay.
+                let htmlBlob = "";
```
Difegue: I don't mind keeping the filter in, but as it stands this breaks the ToC (and soon to be Tank) chapter splitting, since the filter remakes the entire thumbnail section.  

It might be easier to keep track of stamped pages ahead of time and just add a class to the thumbnail divs if they contain stamps -- Then this function only needs to hide every thumbnail that doesn't have that specific class.

templates/reader.html.tt2
```diff
@@ -279,6 +304,11 @@
     <input id="auto-next-page-apply" class="favtag-btn config-btn" type="button" style="display:inline;" value="[% c.lh('Apply') %]">
 </div>
 
+<div id="toggle-stamps-visibility">
+    <h2 class="config-panel"> [% c.lh('Toggle Stamps') %]</h2>
+    <input id="toggle-stamps" class="favtag-btn config-btn" type="checkbox" value="[% c.lh('Disabled') %]">
```
Difegue: Display is broken for me here:  
Image  

I think you just need the `fa` class:  

```suggestion
    <input id="toggle-stamps" class="fa" type="checkbox" value="[% c.lh('Disabled') %]">
```  

It looks a bit weird having a config-style toggle switch where all the other settings are Enabled/Disabled config-btn sets, but I can live with that for a V1. (I'll probably add some description text here too tbh)

templates/reader.html.tt2
Difegue: The tabs are nice CSS-wise, but I don't think it's worth adding a separate layer of UI for what would realistically be one button. (The filter)  

Set/Add Stamp needs to be in Admin Options since it requires auth.  You could argue we could move the entire Admin Options block to that tab, but I'm not sure it's necessary atm...  

I think the filter button could just live next to the thumbnails like so.  

Image

(I realized mocking this up that the current overlay-section display is broken if you don't have a ToC present, so I'll fix that in dev in a little bit)

public/js/reader.js
```diff
@@ -134,6 +140,96 @@ Reader.initializeAll = function () {
...
+        if (e.key === "Escape" && Reader.markerMode) {
+            $("#overlay-page").hide();
+            Reader.markerMode = false;
+            Reader.renderMarkers();
+            Reader.toggleArchiveOverlay();
+            Reader.pageNaviState = true;
+            $(".reader-image").css("cursor", "");
+        }
+    });
+    $(document).on("click.set-stamp", "#set-stamp", Reader.addStamp);
```
Difegue: I would also bind this to a new keyboard shortcut on "S".  
(With matching text in the help popup)

Difegue (changes requested):

public/js/reader.js
EfronC: TBH, I'm not sure why I didn't do it like this on the first place, it's better than rewritting the whole segment. Basically I just added a data-stamped to the ones that are, and hide by it.

public/js/reader.js
EfronC: Actually, I removed the button from the overlay and moved set stamp to this keybinding, feels more natural, and gives you the option to highlight the markers with the combo S -> Esc.

Difegue (approved):
Good to go! Thanks for all the work thats been put in this.

PR #1518, opened 2026-04-09
Only push a new history entry if it is different from the current URL

pneumaticity:
This means duplicate history stack entries will not be created when reloading, which makes going back easier and avoids blanking out forward history unless required.

To explain the issue more thoroughly: suppose you navigate from the search page to an archive, use 'back' to get back to the search page, and perform some action that triggers this code (like updating a rating from the search page, or just refreshing). Before this change, that would result in a pushState call that got rid of the 'forward' history entries, so you couldn't navigate forward to the archive again.

Even if there isn't forward history to lose, repeated reloads end up creating more and more (identical) entries on the history stack. Keep hitting f5 and watch the back stack grow (or use browser session restore, that's how I noticed this).

Please note that you need to have nonempty query params to trigger these cases.

### DIFF
--- a/public/js/index_datatables.js
+++ b/public/js/index_datatables.js
@@ -281,8 +281,17 @@ IndexTable.drawCallback = function () {
             IndexTable.isComingFromPopstate = false;
         } else {
             let params = IndexTable.buildURLParameters();
-            if (params === "?") params = "/";
-            window.history.pushState(null, null, params);
+            // don't push duplicate state entries, because that would wipe out forward history and
+            // require multiple 'back' presses to go back)
+            if (params === "?") {
+                // special case for empty search params: window.location.search is "" if there are
+                // no search params, even if window.location ends with '?'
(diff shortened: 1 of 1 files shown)


psilabs-dev:
This does address the issue (in the case of nonempty params) but I'm seeing the buggy (browser stack) behavior on chrome when params are empty.

A further filtering i.e. (`} else if (params !== "?" && params !== window.location.search) {`) would work... but at that point, maybe the problem is with buildURLParameters?

pneumaticity:
This should fix it. I had this working but stupidly forgot to test after 'simplifying' the code in an invalid way. Sorry about that.

pneumaticity:
I don't think the [CI failure](https://github.com/Difegue/LANraragi/actions/runs/24204452575/job/70656460843?pr=1518) is related to this change. I don't see how this change could cause it, and I also see something similar in #1520.

Difegue (approved):

PR #1519, opened 2026-04-09
Add go-to-page feature.

pneumaticity:
When you know exactly what page you want, and especially for larger archives, this can be faster than pulling up the thumbnails with Q.

This has been working well for me for a while. A caveat: it does use prompt() rather than any fancy DOM-based UI, but I tested a bit on firefox and safari and on both of these prompt() seems pretty OK (doesn't seem to disable back/forward or closing the tab, but does prevent page interaction).


Difegue:
Works for me. We can update this in time to use `LRR.showPopUp()` but it doesn't feel necessary for a first iteration. Thanks!

PR #1520, opened 2026-04-09
Fix tests and clear up ci warnings

Guerra24:


Guerra24:
Turns out perlmagick is not enough to get the suit working so add it back as a dependency.

Also make sure the tests run using the current built image instead of the nightly build of the previous commit, this is why the failure was missed in the alpine pr.

PR #1522, opened 2026-04-12
Use `__FILE__` instead of `FindBin::Bin`

psilabs-dev:
For some reason, depending on whether you run prove with unit tests or perl, find::bin gives a different directory per entry script, which resulted in spamming the repo with log files during unit tests.

Sometimes it leaks to `tests/LANraragi/log/lanraragi.log`, because we run from `tests/LANraragi/Utils/Logging.t`.

According to https://perldoc.perl.org/FindBin:

"$Bin or $Dir
Path to the bin directory from where *script* was invoked"

Not sure why we use this; wouldn't `__FILE__` be more stable anyways?

### DIFF
--- a/lib/LANraragi/Utils/Logging.pm
+++ b/lib/LANraragi/Utils/Logging.pm
@@ -7,7 +7,8 @@ use utf8;
 use feature 'say';
 use feature 'isa';
 use POSIX;
-use FindBin;
+use File::Basename;
+use Cwd 'abs_path';
 use Time::HiRes;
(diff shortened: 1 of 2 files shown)


Difegue:
The idea with FindBin vs `_FILE_` is that the launch script is basically always at the same position, whereas `__FILE__` is subject to needing to retool the path in case the code gets moved to a different file/package.  

I do admit it's unlikely for logging/temp to really move at this point, so it's probably fine.

PR #1526, opened 2026-04-15
Parse top-level GalleryDL metadata into tags

aspasia-patron:
This commit maps top-level gallery-dl json metadata into LANraragi namespaced tags for nhentai specifically, but i'm sure i saw other sources storing artist, language at the top level.

artist, 
group,
parody,
character/characters,
language/lang,
type.

Testcase included.


okaros:
Oh nice, you did the extra legwork to handle the top-level tag collisions that were giving me headaches back when I originally wrote this. Thank you!

Only suggestion I have is to add some additional test cases to cover:
*  The lang->language fallback (the current test case has both entries, so the fallback doesn't get exercised).
*  Same with merging the characters and character fields at the top-level.

But that's really minor stuff.

@Difegue Looks good to me. I poked and prodded it in my dev instance and it works cleanly with all the examples I could think to throw at it.  Only edge case I could see is if someone did something weird with their gallery-dl config: If you were deliberately injecting a different author/group/etc...  in the tags and relying on the plugin to ignore the original gallery-dl top-level metadata field this would cause *both* versions to be included. I think that's a pretty tolerable side-effect, though, and I don't see any way to work around it without setting up a different flavor of the same problem.

aspasia-patron:
Thanks for the review. I will add more test cases later today. I have a couple of other questions/suggestions related to Lanraragi:

- I initially had to look at the code due this plugin not parsing the metadata.json:  I had to adjust my json naming and create a conversion script. I think it will be better to allow arbitrary json to be parsed if it exists in the expected place.

- I was deliberately storing galleries uncompressed and had to redo my archives solely for Lanraragi. Storing the gallery as a directory with individual files allows me to find the duplicates and compare media easily. I'd love to understand why many (Mihon, LANraragi came to my mind) cataloging software insist on storing media in a certain way. 

Regarding gallery-dl metadata: I was thinking of taking a long shot approach and verifying support for every manga-related extractor from gallery-dl, not just nhentai. Gallery-dl already has tests i can compare against.

For me personally it'd be easier to implement this plugin in another language, but as far as my current assessment goes, no reference plugin/binding to other language exists for Lanraragi. I don't think it's sustainable to maintain your own metadata scrapers and parsers like in this case in Perl. Projects like gallery-dl can do a better job for scraping, and allowing people to populate the DB with shim scripts will give more flexibility long term.

okaros:
> Projects like gallery-dl can do a better job for scraping, and allowing people to populate the DB with shim scripts will give more flexibility long term.

The LRR API is pretty solid on that front and might be something to consider looking at if what you really need is  just going straight from gallery-dl into LRR. That has the bonus of not needing to be perl, too. gallery-dl supports executing arbitrary scripts/commands during/after its various steps, so it wouldn't be too bad to wire up from its side of things.

aspasia-patron:
>The LRR API is pretty solid on that front and might be something to consider looking at if what you really need is just going straight from gallery-dl into LRR

Thanks for the hint. I will likely implement this plugin on my own in another language and will figure out a way for how to integrate it into lrr.

Had to generate synthetic data with an LLM due to setup issues on my end for the latest commit.

aspasia-patron:
Search.pm expects json with script_load, evalsha, but mocks.pl doesn't have those methods. Is this correct? My code didn't modify any of it, so it looks like a preexisting issue.

Difegue:
script_load is unmocked yeah but that's expected atm, the tests only verify the fallback ordering.  

The test failure was in the language tag parsing part, I went ahead and fixed it.  
The code changes look good to me otherwise, thanks!

PR #1537, opened 2026-04-30
Add support for uncompressed folders as content

qip:
Treat leaf directories containing image files as first-class content items alongside compressed archives. A folder qualifies as content when it has images and no subdirectories that themselves contain images.

Changes:
- Utils/Generic: add is_content_folder() leaf-directory predicate, returns true for leaf directories containing images with no image-containing subdirectories
- Utils/Archive: add directory branches
- Utils/Database: directory-aware compute_id as SHA-1 of concat first 16KB per file in sorted order
- Shinobu: discover content folders during scan, watch for image file events to trigger folder detection, handle folder deletion
- Model/Archive: serve directory pages directly from disk
- Controller/Api/Archive: for now, block download for folder-based content, could try client-side download all then zip, or just server-side compression, leave for future developement
- tests/folder_support.t: I've no idea how to write tests with perl, purely ChatGPT's take

### DIFF
--- a/lib/LANraragi/Controller/Api/Archive.pm
+++ b/lib/LANraragi/Controller/Api/Archive.pm
@@ -104,6 +104,11 @@ sub serve_file {
 
     my $file = get_archive_path( $redis, $id );
     $redis->quit();
+
+    if ( -d $file ) {
+        return render_api_response( $self, "serve_file", "Cannot download a folder-based archive as a single file." );
+    }
+
     $self->render_file( filepath => compat_path($file), filename => basename($file) );
 }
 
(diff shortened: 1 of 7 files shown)


qip:
I know there's similar requests marked as won't fix and it has been stated as designer choice, but well, just in case it could be helpful.

BTW although I've make sure it's working, you can expect issues, I'm no good at perl at all.

Difegue:
Thanks for this; As you've inferred I don't plan on supporting sparse folders in core at the moment as it can cause multiple issues with Windows and/or symbolic links.

The patches look good at a glance though so if they work for you, nice! I expect it'll be slightly annoying to maintain given the amount of spots you need to modify though.

PR #1538, opened 2026-05-01
nHentai support for meta.json

Guerra24:
meta.json doesn't match the API so I'm going to leave this as a draft until that gets resolved... hopefully. Also makes the api key a requirement, this is a prerequisite for *the other thing*.


Guerra24:
Tried contacting them about this but got nothing back. Contact email is unresponsive. Closing this for time being.

PR #1547, opened 2026-05-08
Remove duplicate function

siliconfeces:
This function is overwritten by another implementation directly after.


Difegue:
uhhhh for my defense I was just moving blocks of code around without reading the function names 

i sure hope someone gets fired for that blunder

PR #1548, opened 2026-05-09
Use proper JS modules

siliconfeces:
This PR turns a bunch of JS files into JS modules.

Primarily it's a big, annoying PR to review. However, it also makes chunky JS files a bit easier to split up, makes it a touch easier to extract common code to library modules, etc!


siliconfeces:
Not sure why the tests fail on windows, but it seems to be consistent. Stamps frontend will be merged before this either way so keeping this in draft for now!

siliconfeces:
There's some more stuff that can be spruced up (`marked` for instance) but I'll leave that for another PR

Difegue (approved):
Looks OK and I appreciate this cutting down on the lines of JS imports in the templates.  

Hopefully I won't forget to add `export`s to functions in the future and wonder why shit aint working

Issue #1556, opened 2026-05-14
Logger crashes with "Failed to acquire shared log lock" on NFS-backed log volumes


zakame:
**LRR Version and OS**
LANraragi v0.9.71 (post-#1406, on `dev`). Running in Kubernetes (k3s) on Linux, Docker image deployment. Log volume is an NFS-backed PersistentVolume mounted without a working lock manager (default `nolock` / no `local_lock=flock`).

**Bug Details**
After #1406 added flock-based locking to log rotation, the logger now dies on NFS-backed log volumes when `flock()` returns `EBADF` or `ENOLCK` — both are documented possibilities on NFS without a working lock manager (see flock(2) and `perldoc -f flock`). The fatal call sites are in `lib/LANraragi/Utils/RotatingLog.pm`:

- `append()` — `flock($lockfh, LOCK_SH) or die "Failed to acquire shared log lock: $!"`
- `new()` and `maybe_rotate()` — similar `die` on lock acquisition

Before #1406, log appends were best-effort and tolerated filesystems where `flock()` is unreliable. After #1406, any `info()`/`warn()` call from the app can propagate the die up through the caller, taking out the request or worker.

This affects anyone running LRR with logs on:
- NFS without `local_lock=flock` (or without nfs-lockd / rpc.statd reachable)
- Other shared filesystems where `flock()` returns `EBADF`/`ENOLCK`

**Matching Logs**
```
Failed to acquire shared log lock: Bad file descriptor at lib/LANraragi/Utils/RotatingLog.pm line 93.
```

**Suggested fix direction**
1. Classify `EBADF`/`ENOLCK` from `flock()` as "this filesystem does not support flock," log one warning, and fall back to the pre-#1406 best-effort path (no locking, no rotation). Other failures (`EWOULDBLOCK` from `LOCK_NB` contention) should still propagate normally.
2. Add an `LRR_LOG_LOCK_DIRECTORY` env var so operators who must keep logs on the network volume can relocate just the lockfile to a local filesystem (e.g. `/tmp`) and keep rotation working.
3. Document the two operator escape hatches:
   - `LRR_LOG_LOCK_DIRECTORY=/tmp` (or any local FS), or
   - mount the log volume with NFS `local_lock=flock`.

Happy to send a PR — I've already got the fix and tests working on my k3s cluster. Wanted to file the issue first so the discussion has a home.

psilabs-dev:
Hi, Thanks for the report!

[...]

PR #1559, opened 2026-05-16
RotatingLog probe check and lock directory configuration

psilabs-dev:
Issue: https://github.com/Difegue/LANraragi/issues/1556

Add a flock probe check at the start of `get_logger`. If probe fails, fallback to mojo log. Probe check is cached per-process.

Also add suggested `LRR_LOG_LOCK_DIRECTORY` configuration.

### DIFF
--- a/lib/LANraragi/Utils/Logging.pm
+++ b/lib/LANraragi/Utils/Logging.pm
@@ -89,7 +89,7 @@ sub get_logger {
                         path    => $logpath,
                         level   => 'info',
                         logfile => $logfile,
-                        tempdir => $tempdir
+                        lockdir => $tempdir
                     );
                 };
                 $retry_error = $@;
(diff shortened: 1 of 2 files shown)


Difegue (approved):

zakame:
@psilabs-dev @Difegue sorry for the late reply - this PR doesn't fix my issue still:

```
Attempting to create directory /home/koyomi/perl5
Can't load application from file "/home/koyomi/lanraragi/script/lanraragi": Failed to acquire shared log lock: Bad file descriptor at /home/koyomi/lanraragi/script/../lib/LANraragi/Utils/RotatingLog.pm line 93.
```

I think its not enough to test whether a lock was acquired, but also if the lock is valid (no `EBADF`/`ENOLOCK`.) The tests on my PR check for this and I had that PR/branch deployed OK since the time I posted it.

PR #1563, opened 2026-05-17
Fix OpenAPI validation bypass

psilabs-dev:
Bug found by @Guerra24 , please clarify if needed.

It is possible to bypass OpenAPI validation even when it's enabled with `Content-Type: application/json; charset=utf-8`. The bug is in JSON::Validator, thanks to archey347 in related PR: https://github.com/jhthorsen/json-validator/pull/289 and PR 266.

For the sake of simplicity, multipart-bypass is applied, as applying the 289 fix would introduce another issue as discussed in 266 (which we might also bring in). We may also consider vendoring source + tests depending on how much we want to handle this problem.

Also explicitly added JSON::Validator.


Guerra24 (commented):
I would say we need this fix but I would prefer if it was applied in the same way the Crypt::DES or Image::Magick patches are applied.

Guerra24 (approved):
LGTM!

Difegue (approved):
gucci

Issue #1562, opened 2026-05-16
Thumbnail generation failed on ARMv7 architecture


xperiazu21:
**LRR Version and OS**
v.0.9.71, on Debian 13 (Raspberry Pi OS 32Bit), Docker method.

**Bug Details**
After adding a new archive, automatic thumbnail generation fails. Even using the "Regenerate all thumbnails" option in the config doesn't work. However, this issue does not occur on another ARMv8 machine running via Docker.

I did some testing with a clean Lanraragi environment setup. Thumbnail generation works fine starting from version v.0.9.40, but the issue begins to appear in version v.0.9.60. I suspect this might be related to the introduction of libvips support.

**Matching Logs**
Docker logs

```
[Minion] [info] Starting thumbnail regen job (force = 1)

(process:189): GLib-GObject-CRITICAL **: 16:45:03.974: value "0" of type 'gint' is invalid or out of range for property 'width' of type 'gint'

(process:213): GLib-GObject-CRITICAL **: 16:45:04.762: value "0" of type 'gint' is invalid or out of range for property 'width' of type 'gint'
[Archive] [error] Thumbnail generation failed for archive '/home/koyomi/lanraragi/content/test1.zip' entry '001_cover.jpg' -> '/home/koyomi/lanraragi/thumb/e1/e1cab705ca32bfc265042ccbc10f8455659b5dfa.jpg': Error resizing image for fit: thumbnail_buffer: no property named `'

[Minion] [warn] Error while generating thumbnail: Error resizing image for fit: thumbnail_buffer: no property named `'

```

**Screenshots**
Minion screenshot

<img width="750" height="447" alt="Image" src="https://github.com/user-attachments/assets/646b2d90-c44f-4b59-aab9-a27d77386eb1" />

siliconfeces:
Which raspberry pi model is that?

xperiazu21:
It's Raspberry Pi 2B v1.1

xperiazu21:
I tried your pull request changes. It successfully generates thumbnail files now, but the thumbnails are all completely black.

Guerra24:
@xperiazu21 Hi, I validated it in a QEMU ARMv7 VM and it did fix the issue. Could you try with other archives or deleting and adding it back to see if that triggers thumbnail generation again?

xperiazu21:
[...]

PR #1565, opened 2026-05-19
Allow disabling VIPS support

siliconfeces:
Allow disabling VIPS if it causes issues with some slop tests. Related to #1562 I guess.


Difegue:
I guess we don't need this anymore since the vips issue stems from alpine?

siliconfeces:
elmo

Maybe valuable for troubleshooting, maybe completely pointless.

Difegue:
Let's close it for now; We can always bring it back if further trouble shooting w/ vips is needed.

PR #1566, opened 2026-05-19
Manually pin Minion to ver 11.0

Guerra24:
Had to manually reference Minion 11.0 because cpanm refused to install the correct version.


Difegue (approved):
its never easy

PR #1571, opened 2026-05-23
Mention Linux support in the reader docs

Guerra24:


Difegue (approved):

Difegue:
the avalonia dream is real

PR #1576, opened 2026-05-27
Add eslint to CI

siliconfeces:
Run eslint in CI and fix the errors (not warnings) that made it go boom.

### DIFF
--- a/.github/workflows/push-continuous-integration.yml
+++ b/.github/workflows/push-continuous-integration.yml
@@ -18,6 +18,18 @@ env:
 
 name: "Continuous Integration \U0001F44C\U0001F440"
 jobs:
+  eslint:
+    name: ESLint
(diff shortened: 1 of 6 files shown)


.github/workflows/push-continuous-integration.yml
```diff
@@ -18,6 +18,18 @@ env:
 
 name: "Continuous Integration \U0001F44C\U0001F440"
 jobs:
+  eslint:
+    name: ESLint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-node@v4
```
Guerra24: > The following actions are running on Node.js 20 and may not work as expected: actions/setup-node@v4
```suggestion
    - uses: actions/setup-node@v6
```

PR #1577, opened 2026-05-27
Add htm, use for toasts

siliconfeces:
Be more fancy and use HTM for toast markup.

By itself it's not necessarily very interesting, but it's an example/demonstration that people can now use preact/htm for frontend stuff.


Difegue (approved):
I'm not necessarily convinced given we barely have any preact components, but the syntax does look better for toasts. Let's see if people use this i guess!

PR #1582, opened 2026-05-29
Move version cache buster to path for js

siliconfeces:
I wasn't 100% pleased with how importmap-reliant JS imports were, so I changed approach. It also prevented most IDEs from resolving the imports for autocomplete, so that works better now.

Now the version cache-buster is part of the path and doesn't require importmap-entries for "ordinary" non-vendor javascript files. This allows you to do relative imports (like `./server.js` instead of `mod/server`).

A disadvantage is that importing stuff from HTML files is a touch uglier again as you need to build the path each time rather than just once in the importmap.

### DIFF
--- a/lib/LANraragi/Utils/Routing.pm
+++ b/lib/LANraragi/Utils/Routing.pm
@@ -3,6 +3,7 @@ package LANraragi::Utils::Routing;
 use strict;
 use warnings;
 use utf8;
+use v5.36;
 
 use Config;
 use Encode;
@@ -12,6 +13,8 @@ use Mojolicious::Plugin::Minion::Admin;
 
 use LANraragi::Utils::Login      qw(is_logged_in_api);
 use LANraragi::Utils::OpenAPI    qw(apply_openapi_mojo_overrides);
+
+use Cwd qw(abs_path);
 
 use constant IS_UNIX => ( $Config{osname} ne 'MSWin32' );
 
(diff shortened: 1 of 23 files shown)


lib/LANraragi/Utils/Routing.pm
```diff
@@ -68,6 +68,14 @@ sub apply_routes {
     # Routers used for all loginless routes
     my $public_routes = $self->routes;
 
+    $public_routes->get( '/js/:version/*filepath' => [ version => qr/\d+\.\d+\.\d+/ ] )->to(
```
psilabs-dev, line 74: paths need sanitization 

Image

siliconfeces: Big oof!

siliconfeces: The mojolicious path class is a bit "special" (never got `merge`and `contains` to do anything useful), but I think I've managed to block that stuff now, albeit a bit heavy-handed.

Difegue: The /page API does something similar with the way it handles paths, iirc we just resolve the given path and verify if it's a sub-path of the content folder.  

Feels like we could do a similar thing here instead of explicitly checking for ".." which feels a bit brittle?

Difegue (changes requested):
One question - The current iteration looks sound tho so we could potentially just go with that

lib/LANraragi/Utils/Routing.pm
siliconfeces: Just pushed some stuff just as you were reviewing, but that was mostly tests.

Doesn't the /page api just fetch stuff from cache or from inside archives nowadays?

Difegue: Ah right it uses peek directly now, it used to extract the whole shebang beforehand and that's where the path traversal blocker came in.  

The comment still applies tho but realistically I don't think there'd be another way to get it to break outside to the wider filesystem without `..`... Unless there's some magic linux mojo i'm unaware of

siliconfeces: I want to do some further research, so I'll keep this PR in draft for now.

siliconfeces: Changed the approach to be more solid, uses abs_path and checks against the `public` directory.

Difegue (approved):
lgtm, good thought on symlinks. (altho i think if you symlink to /etc/passwd in yout public folder you're just asking for it at this point, but still)

PR #1583, opened 2026-05-29
Extract index context menu to a separate module

siliconfeces:
Move context menu code for the index page into a separate module. Reduces size of thicc JS and moves some JS out of a template file.


public/js/mod/index_contextmenu.js
```diff
@@ -0,0 +1,261 @@
...
+            return items;
+        },
+    );
+}
+
+// #endregion
+
+export function initialize(catListData) {
+    const catList = catListData || [];
+    pseudoCopyBtn = $("#pseudo-copy-btn")
```
Difegue:
```suggestion
    pseudoCopyBtn = $("#pseudo-copy-btn");
```

PR #1584, opened 2026-05-30
Fix tank progress updates

psilabs-dev:
There is a logic error with tanks where `&& progress > data.progress` is always false. See Model::Tankoubon::get_tankoubon:

```perl
$tank{progress} = int( $tank{progress} || 0 );
```

This means the condition never updates, which means that tank progress is never updated. A couple edits to harden the contract:

1. explicitly handle tank progress status codes
2. pull `id.startsWith` into an indicator
3. define `serverProgress` based on isTank.

The `data.progress` and `data.result.progress` inconsistency is probably going to bite us again though.


Difegue:
i used this PR as a hotbed to redesign the tank API a bit.  

get_tank now has the exact same return syntax as its predecessor. Paging and fulldata are in a dedicated endpoint instead.

PR #1591, opened 2026-06-03
Reader modernization step 1

siliconfeces:
Split up reader and start modernizing by using preact.

Initially this PR was only intended to split the reader into multiple files based on functionality, but has since expanded to also start migrating stuff to preact.

Switching between standard view mode and infinite view does no longer reload the page. Hopefully this have not introduced too many new bugs.

Attempts have been made to keep attribution, but may require some specific `git blame` calls to get that showing up and a squash merge will definitely break that.


public/js/mod/reader_common.js
```diff
@@ -0,0 +1,1243 @@
...
+                .attr("href", artistSearchUrl)
+                .text(artistName);
+            const titleContainer = $("<span></span>")
+                .text(`${state.content.title} by `)
+                .append(link);
+            $("#archive-title").empty().append(titleContainer);
+            $("#archive-title-overlay").empty().append(titleContainer.clone());
+        } else {
+            $("#archive-title").text(state.content.title);
+            $("#archive-title-overlay").text(state.content.title);
```
siliconfeces, line 145: Stuff like this is technically archive overlay-related and should rather end up there, but I'm thinking about handling that in follow-up PRs.

Does that sound reasonable?

Difegue:
Pausing this while #1214 gets thru

public/js/mod/reader_common.js
```diff
@@ -0,0 +1,1605 @@
...
+    effect(() => {
+        clearMarkers();
+        localStorage.infiniteScroll = state.infiniteScroll.value;
+    });
+
+    effect(() => {
+        localStorage.AutoNextPageInterval = +state.AutoNextPageInterval.value || 10;
+    });
+
+    effect(() => {
```
Difegue, line 263: I think you're missing an effect to save ignoreProgress back to localStorage (If I understand how this stuff works properly)

Difegue: Same for containerWidth ?

public/js/mod/reader_stamps.js
```diff
@@ -0,0 +1,390 @@
...
+        // Reload the markers everytime the image size changes
+        renderMarkers();
+    });
+}
+
+/** Process inputs
+ * @param {JQuery.KeyDownEvent<Document, undefined, Document, Document> | JQuery.KeyUpEvent<Document, undefined, Document, Document>} e
+*/
+function handleShortcuts(e) {
+    switch (e.which) {
```
Difegue, line 120: This is missing the  

```
if (e.target.tagName === "INPUT") return;
```  
gate that's in the shared shortcut handler.  

I haven't tested it but you might be able to activate stamps by error by inputting "s" in the container width textbox or something.

Difegue (changes requested):
Sorry for the merge conflicts yet again, but I've given this a pass and I dont think there'd be too many bugs with that one. Just a few Qs from me  

Fine by me to keep trimming down _common in later PRs.

Difegue (approved):
:shipit:

PR #1593, opened 2026-06-05
Convert to docker multi-stage build and drop Magick

### LINKED ISSUE #1562: Thumbnail generation failed on ARMv7 architecture

Guerra24:
- Merges part of the install scripts into the Dockerfile to reduce layers.
- cpanm now installs everything in the home's local lib directory.
- Drops Magick from the docker image.
- Removed unused dependencies.
- Removed the 8-second delay when stopping the container.
- Reverts 427abc9a716805c0ec044e7fb5843fec8fa2e229 because caching is not ideal and causes more misses instead.
- Bumps to Alpine 3.24. Fix #1562


Difegue (approved):
LGTM.  
Should we consider moving module patching to `install.pl` as part of this PR?

Guerra24:
- Partially reverted #1566 and removed the hardcoded cpan mirror which is what broke pinning in the cpanfile.
- Merged again #1340.
- Also cleanup install.pl in preparation for moving patching to it but turns out we might not need it, still kept the improvements.

tools/install.pl
```diff
@@ -163,15 +164,16 @@
...
-        install_package( "Win32API::File", $cpanopt );
+        install_package( "Win32::Process",                     $cpanopt );
+        install_package( "Win32::FileSystemHelper",            $cpanopt,
+            "https://github.com/Guerra24/Win32-FileSystemHelper/archive/308b92c958bb4931dfd704cc5025f93e28ef0c8a.zip" );
+        install_package( "File::ChangeNotify::Watcher::Win32", $cpanopt,
+            "https://github.com/Guerra24/File-ChangeNotify-Watcher-Win32/archive/7cb4e60823569cca8e7652d19b1ba5b5cac00a16.zip" );
+        install_package( "Win32API::File",                     $cpanopt );
     }
 
+    install_package( "Net::IDN::Encode", $cpanopt, "ETHER/Net-IDN-Encode-2.501-TRIAL.tar.gz" );
```
TomaSajt, line 175: I am updating this package inside nixpkgs and discovered this change.
In what way is `Net::IDN::Encode` being used? Does `ETHER`'s `2.501-TRIAL` version have fixes that `CFAERBER`'s `2.500` version doesn't?

TomaSajt: Inside nixpkgs we have the changes from https://github.com/cfaerber/Net-IDN-Encode/pull/11 applied as a patch applied to `Net::IDN::Encode`. Seemingly it's the same change as `ETHER`'s `2.501-TRIAL` version.

So I'll just use our patched version.

Guerra24: `Net::IDN::Encode` is being used indirectly through [json-validator](https://github.com/jhthorsen/json-validator). Originally it was not a hard-dependency but after changes in the validator it became one and the work required to patch and remove it is way higher than simply leaving it as is.

The issue with CFAERBER's version is that it does not compile on Perl 5.38+ and as far as I know it has been completely abandoned with the author being completely unresponsive. ETHER's version is simply 2.500 + a tiny patch that fixes it, the version on CPAN is the one in [this repo](https://github.com/karenetheridge/Net-IDN-Encode).

Guerra24: > So I guess I can just patch out this line.

Yes, that looks fine to me.

TomaSajt: Thanks for the answers!

PR #1590, opened 2026-06-03
Open lockfile with read access to fix flock on NFS

zakame:
Fixes #1556

Narrows the original #1561: just the `+>>` lockfile open mode fix as suggested by @psilabs-dev.

Since #1406, `RotatingLog` dies on `LOCK_SH` in `append()` when the log volume is NFS - the fd was opened write-only (`>>`), which does not satisfy NFS flock requirements for shared locks per [`perldoc flock`](https://perldoc.perl.org/functions/flock).

**Changes:**
- `+>>` open mode on lockfile handles (`lockfh` and `ensure_lock`) so the fd has read intent
- Remove unused `O_CREAT` and `O_RDWR` imports from `Fcntl`

PR #1595, opened 2026-06-08
Address the probe part of the 1556 problem

psilabs-dev:
Issue: https://github.com/Difegue/LANraragi/issues/1556

The probe should check that "+>>" works as intended. Of course, this assumes that "+>>" is applied to all instantiating cases of RotatingLog (which part of https://github.com/Difegue/LANraragi/pull/1590 covers).


Difegue:
i think it's the first time i'm merging two 2-line PRs from different people in a row that affect different parts of the same file to fix the same bug

PR #1596, opened 2026-06-08
Keep display on when using auto next page

yoraldo:
Auto next page is a great feature, but I got tired of tapping on my phone when it was about to sleep. This uses [WakeLock](https://developer.mozilla.org/en-US/docs/Web/API/WakeLock) to keep the device screen on while auto next page is being used. It is supported by latest devices and browser versions, but the only caveat is that it requires an HTTPS connection.

Tested on a couple of browsers on Android and Windows.

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -1977,12 +1977,6 @@ async function requestWakeLock() {
         console.error("Error acquiring wake lock:", err);
     }
 }
-// Browser releases the wake lock when the tab is hidden so we need to reacquire it
-document.addEventListener("visibilitychange", async () => {
-    if (autoNextPage && wakeLock !== null && document.visibilityState === "visible") {
-        await requestWakeLock();
-    }
-});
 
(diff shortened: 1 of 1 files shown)


public/js/reader.js
```diff
@@ -1952,3 +1957,30 @@ jQuery(() => {
         }
     });
 });
+
+async function requestWakeLock() {
+    if (wakeLock !== null) {
+        return;
+    }
+    if (!("wakeLock" in navigator)) {
+        console.error("Wake Lock API is not available.");
```
Difegue:
```suggestion
        console.error("Wake Lock API is not available. You're likely running in an outdated browser or without HTTPS.");
```

Difegue:
Nice, works for me.  
A bit unfortunate that this degrades outside of https, but unlike with clipboard interaction it doesn't feel important enough to try and figure out a workaround so 🤷

PR #1597, opened 2026-06-08
Fix tank-related bugs

psilabs-dev:
Two bugs.

Bug 1: Tank currently breaks: in safari if you open a tank you get:

```
Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'pages[index].split')
```

because "/full" doesn't necessarily return fulldata...

Bug 2: after fixing bug 1, if you open a tank, the thumbnail API will 400, due to OpenAPI validation. Hence the fix at templates/reader.html.tt2

### DIFF
--- a/lib/LANraragi/Controller/Api/Tankoubon.pm
+++ b/lib/LANraragi/Controller/Api/Tankoubon.pm
@@ -44,7 +44,7 @@ sub get_tankoubon_full {
     my $tank_id = $self->stash('id');
     my $req     = $self->req;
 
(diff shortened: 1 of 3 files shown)


Difegue:
Hmm, I thought I'd solved 1 before pushing the tank API split, but apparently not.. Thanks for catching that. 🙇

Difegue:
(I'm not convinced defaulting to -1 for /full is a good idea, but it is what the code does so it doesn't matter for this PR)

PR #1598, opened 2026-06-10
Remember last selected metadata plugin.

pneumaticity:
Saves the most recently selected metadata plugin to localStorage and selects it by default.

This saves a couple clicks for me on average; ymmv. The current behavior is for the first plugin in the list to be selected.

I've seen some work is ongoing for a plugin registry, but can't tell if there are any changes to the edit page planned that would obsolete this.


Difegue:
I don't believe plugin registry will impact this, no. Thanks for the conttribution, I think that's helpful 👍

PR #1600, opened 2026-06-12
Fix archive replacement when uploading different content with the same filename

Reiyy:
When uploading an archive through the `api/archives/upload` endpoint with the same filename but different file content, metadata provided in the request (such as tags) is not applied correctly. The archive ultimately continues to display the metadata from the previously uploaded archive.

If the exact same file is uploaded again (same filename and same content), the behavior is correct.

Cause:
In `lib/LANraragi/Model/Upload.pm`:

```perl
# If we are replacing an existing one, just remove the old one first.
if ($replace_dupe) {
    $logger->debug("Delete archive $id before replacing it.");
    LANraragi::Model::Archive::delete_archive($id);
}
```

The `$id` passed to `delete_archive()` is the ID computed for the newly uploaded archive, rather than the ID of the existing archive currently associated with that filename.
When the uploaded content is identical, both the old and new archives generate the same ID, so the issue does not become visible.

This fix only addresses that specific scenario. All other existing replacement behavior remains unchanged.

### DIFF
--- a/lib/LANraragi/Model/Upload.pm
+++ b/lib/LANraragi/Model/Upload.pm
@@ -83,8 +83,21 @@ sub handle_incoming_file ( $tempfile, $catid, $tags, $title, $summary ) {
 
     # If we are replacing an existing one, just remove the old one first.
     if ($replace_dupe) {
-        $logger->debug("Delete archive $id before replacing it.");
-        LANraragi::Model::Archive::delete_archive($id);
+        my $redis_config = LANraragi::Model::Config->get_redis_config;
+        my $encoded_output_file = IS_UNIX ? encode_utf8( $output_file ) : $output_file;
+        my $old_id = $redis_config->hget( "LRR_FILEMAP", $encoded_output_file );
+        
+        # Same filename, different content: old_id found via filemap
+        if ( $old_id && $old_id ne $id ) {
+            $logger->debug("Delete archive $old_id before replacing it.");
+            LANraragi::Model::Archive::delete_archive($old_id);
+            $redis_config->hset( "LRR_FILEMAP", $encoded_output_file, $id );
+        } else {
+            $logger->debug("Delete archive $id before replacing it.");
+            LANraragi::Model::Archive::delete_archive($id);
+        }
+
+        $redis_config->quit();
     }
 
     # Add the file to the database ourselves so Shinobu doesn't do it


Difegue:
Greetings,  

In this scenario I believe the server _almost_ functions as expected:  
- Uploading a file with a different ID will not delete the metadata of the previous ID, as they're different files (dupe detection does nothing)
- If the filename of the new file is the same, it will overwrite the actual physical file (this is where it's fucked)
- The database itself will have both IDs cleanly separated, although the filemap will retain 2 entries for the same filename with different IDs, which is likely what causes the server to get confused.  

I think the solution here could be much simpler -- We should just add an extra `-e $output_file` check after dupe detection has ran and append a "(1)" to the filename if it already exists. (Or a 2, 3, etc..)

Reiyy:
Thanks for the clarification.

If you use automatic renaming, does this mean the files will "coexist" instead of being "replaced"?

My interpretation came from a few places in the current implementation which all seem to treat same-filename uploads as part of the replacement workflow.

The setting description currently says:

> If enabled, LANraragi will overwrite old archives when a newer one (with the same name) is uploaded through the Web Uploader or the Download System. This will delete metadata for old files when they're replaced!

There is also the comment in Upload.pm:

```perl
# If we are replacing an existing one, just remove the old one first.
```

And the upload API currently handles both content duplicates and filename collisions through the same `replace_dupe` setting:

```perl
# Stop here if file is a dupe and replacement is turned off.
if ( ( -e $output_file || $isdupe ) && !$replace_dupe ) {

    # Trash temporary file
    unlink_path $tempfile;

    # The file already exists
    my $suffix = " Enable replace duplicated archive in config to replace old ones.";
    my $msg =
      $isdupe
      ? "This file already exists in the Library." . $suffix
      : "A file with the same name is present in the Library." . $suffix;

    return ( 409, $id, $filename, $msg );
}
```

From my perspective, the current implementation appears to treat both content duplicates and filename collisions as cases handled by the same replacement feature.

If the intended behavior is instead to keep both archives and automatically rename the new file, that certainly works too. I just wanted to point out that it seems different from how the current setting description, comments, and upload flow describe the feature.

In that case, it might also be worth updating those descriptions and comments so the intended behavior is clearer to users and contributors.

One additional use case that came to mind is archive updates.

For example, if a user uploads a newer version of an existing archive (same filename, different content) and expects it to replace the previous version, automatically renaming the new file would result in both versions remaining in the library.

That may be desirable for some users, but it feels slightly different from the current wording of "overwrite old archives" and "replace old ones", which seem to imply that the older archive is removed as part of the workflow.

Difegue:
Yeah, the core issue here is that one code path (the one that warns you if replace_dupe is false) handles both ID and filename collisions, but the other one only handles ID collisions.

Looking at it again, replace_dupe would be a bit of a useless option if it only replaced exact ID matches... It'd only be convenient for cases where you have two different files with the same ID, which is more of a bug than anything else atm.  

To that end, it probably indeed makes more sense to continue assuming replace_dupe also handles filename collisions, and clean up the existing behavior so it also removes the prior outdated filemap entry.

Difegue:
I'll look at the code again in detail later but it does look like your solution is correct; We might be able to simplify it a bit though.

Difegue:
I ended up reimplementing this myself, but the logic is basically the same. (sans the encoding stuff which I think would break on specific filesystems, and with a bit more comments)  

Thanks for raising this 👍

PR #1610, opened 2026-06-17
Fix RCE due to badly escaped args

siliconfeces:
Partially/mostly AI sloppa, but seems to work!


Difegue:
I'm inclined to do the swap to vips instead of this if it proves to be better.

also fuck sloppa

PR #1611, opened 2026-06-17
Replace ghostscript with libvips

siliconfeces:
This fixes the RCE and is hopefully more awesome in general. It really needs more testing across different platforms and stuff though.


Guerra24:
https://github.com/Difegue/LANraragi/blob/7baaee586dfbcffc3dc9068795485f0b561a060f/tools/build/windows/utf8-support.ps1#L7
https://github.com/Difegue/LANraragi/blob/7baaee586dfbcffc3dc9068795485f0b561a060f/tools/build/windows/create-dist.sh#L12
https://github.com/Difegue/LANraragi/blob/7baaee586dfbcffc3dc9068795485f0b561a060f/tools/build/windows/gs.exe.manifest#L1-L22

siliconfeces:
Tested and works™ on x86-64 and an RPI5 on linux. There's room for efficiency improvements by not doing the rube goldberg thing, but that would require some decent amount of rework and, like, PDF is probably not the most used feature.

tools/install.pl
```diff
@@ -115,12 +115,6 @@
     say("OK!");
 }
 
-#Check for GhostScript
```
Difegue, line 118: The script should probably be edited to check for vips at this point, we're not really in the "optional" phase anymore. 

Maybe check for vips and then fallback to the existing magick check if it fails?

Difegue (changes requested):
Just one nit, I think the change is sound otherwise. 
I haven't really tested this much at this stage, probably need a bit of time to verify if vips parses pdf files as well as GS does.

tools/install.pl
siliconfeces: Sure, if it's okay that the check isn't perfect (warn rather than `die`).

A "proper" check would require that FFI::CheckLib::find_lib is available, and the cpanfile dependencies are installed later.

Difegue: Yeah, that's perfectly fine; The intent is more to warn users that they need vips for pdf if they want to make use of that.

siliconfeces: Done, except on windows where it applies a positive mindset and assumes it's installed.

Difegue:
Sorry for the delay, wanted to do some testing on the change -- PDF parsing seems to be at the same level of quality.  
This PR does generate much larger images from PDFs than the previous version of the server did in some instances. (I've seen x10 page sizes from 500KB to 5MB on some files)  

I think this is just a matter of tweaking the DPI setting in vips. We could also leave it like that for now and let people use the downscaler settings if they want smaller files. I'll merge this and we'll see what people think 👍

PR #1612, opened 2026-06-19
Stamp hotfixes

### LINKED ISSUE #1586: Stamps get wrong position in fullscreen
**LRR Version and OS**
Docker, current nightly

**Bug Details**
Positioning of stamps get all whack in fullscreen.

**Screenshots**
These stamps are supposed to be on the image.

Image

### LINKED ISSUE #1587: Stamps silently fail to add when logged out
**LRR Version and OS**
Docker, current nightly

**Bug Details**
Attempting to add stamps while logged out shows the "Stamp added!" success toast, but the network request fails with a `401 Unauthorized` causing the stamp to not actually be added.

Instead, adding stamps should be disabled while logged out and show an error message if the request fails.

### LINKED ISSUE #1588: Stamps get misplaced in fit to height and short windows
**LRR Version and OS**
Docker, latest nightly

**Bug Details**
If "Fit display to" is set to "Height" AND the window is sufficiently short, stamps gets misplaced.

**Screenshots**
Image

EfronC:
This is just a provisional PR, so I don't forget the required changes needed to fix #1586  , #1587 , and #1588 . This could either be merged now, or we could wait until the reader modernization PRs merge, so this gets reworked on the new style. Either way is ok with me.

While working on this I also met with a behavioral error with server.callAPI, where this fail safe for unauthorized calls, because OpenAPI returns an error response that is not caught in said function, hence leading to make an error look like a correct response. I did not do anything in this PR to fix that function, since is a shared function for multiple type of responses, hence it should be looked more deeply. In the meantime, I just added a workaround to my use of said function, although is not the most beautiful solution.

### DIFF
--- a/public/js/reader.js
+++ b/public/js/reader.js
@@ -222,8 +222,18 @@ export function initializeAll(trackProgressLocally, authenticateProgress) {
             markerMode = false;
             if (result.isConfirmed && result.value.trim() !== "") {
                 const { arcId, localPage } = getArchiveForPage(page);
-                Server.callAPI(`/api/archives/${arcId}/stamps/${localPage}?position=${markerData.x},${markerData.y}&content=${result.value}`, "PUT", "Stamp added!", I18N.StampError,
+                Server.callAPI(`/api/archives/${arcId}/stamps/${localPage}?position=${markerData.x},${markerData.y}&content=${result.value}`, "PUT", null, I18N.StampError,
                     (data) => {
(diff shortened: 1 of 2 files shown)


Difegue (approved):
Looks good. I'l rework callAPI on top of this to make it slightly less jank.  
(I very much forgot about the openAPI-specific error codes when I fixed response.ok obscuring server error messages..)

PR #1615, opened 2026-06-23
Replace ClipboardJS with native usage

siliconfeces:
Using the API directly is simple enough.


Difegue:
Doesn't this require https?

siliconfeces:
Fuck, it does unless it's on localhost.

PR #1629, opened 2026-07-01
Improve cache hits in PRs

Guerra24:
Adjust PR caches to read from the main one instead of it being PR specific and don't write anything back for security reasons.


Guerra24:
Ready to merge. Also dfug could you delete all the caches so that we get a clean state after merging this?

PR #1634, opened 2026-07-03
fix: resolve Server is not defined error in plugin page

chu-shen:


Guerra24 (approved):
Thanks! Works as expected.

PR #1639, opened 2026-07-04
Change encoding used in edit page

siliconfeces:
Title and summary were double-encoded in the inputs.


Guerra24 (approved):

lib/LANraragi/Controller/Edit.pm
```diff
@@ -41,9 +41,9 @@ sub index {
             template  => "edit",
             id        => $id,
             name      => $name,
-            arctitle  => xml_escape($title),
```
psilabs-dev, line 44: xml_escape import should be removed

psilabs-dev (approved):
Besides the import, all good. Thanks!

lib/LANraragi/Controller/Edit.pm
Difegue: will do post-merge

PR #1641, opened 2026-07-04
Manga/comic support concept PR

EfronC:
This is a simple PR to present a possible integration of uncompressed, ongoing type of comics. The idea behind this is that there are some cases that are not covered by the current archive and tanks, like ongoing comic and mangas, that might require to add new files every week/2 weeks/month, but you don't want to either add new zips, that later you have to manually add to a Tank to group them, nor add more files to an existent file, just for the first KBs to be modified, and the metadata to be erased. What this feature attempts is to provide a self organizing tool, that is idempotent to adding new files to it, and also doesn't spam the search results if you added a manga with a 1000 chapters.

For the base design, I'm using Tachiyomi/Mihon local source as the inspiration. Here we have a local folder, ignored by Shinobu, where the user has to follow a structure like:

image

Then we would have a new Minion that goes through this structure and creates the files and tanks on the DB. I'm using a new minion for this, because it has a slightly different logic than Shinobu, so better not try to mess the logic there, and since this workflow would only work if the local folder exists, then this second minion would be able to sleep if said folder does not exist on the content dir, otherwise it would work as a File Watcher, that only reacts when a new chapter folder gets added to add new entries to the DB, when you add a new image file, it would only update the pagecount of its parent file.

image

The most problematic part is how to create IDs for folders, that can be regenerated between all devices without anything to hash, which is basically impossible, so I gave up on that idea, and though of something else. This instead creates a UUID for each folder, and then saves it inside inside a LRR.json(Yet to add to the code), which then, if you decide to move your LRR installation from a Windows computer to a Linux one, with a different encoding, then there would not be a problem, because the JSON will remain utf8 encoded, so you can load the ID from there, and assign the already existent object in the DB to the new folder path; this also works for renaming the folder. The exact same logic applies for the chapters and their IDs.

While I mentioned that that these create archives and tanks, I want to clarify that these are not the same Archive and Tankoubon objects, but new hash tables that will share the same structure, for the purpose of being compatibles with some of the fetch endpoints for search and reader, but by itself these should not be compatible with other tools for management, since the minion should be the main path to manage the files and chapters, based on the file structure. In other words, these would be compatible with the structure response of the GETs for Archive and Tanks(But for a few differences, like there being no need to uncompress the file first, since it'd already be), so the already existent interfaces from the Frontend can be reused, but will require their own management endpoints, being the minion the main tool to maintain them(Chapters also won't be possible to be obtained from the search, since they won't be added to either the archive cache, nor be separated from their Tank, only the tank will be added to the search like the original tanks do).

image

This is a simple example PR, nothing works yet, basically this only provides the minion to explore the file structure, and create the chapter objects(Haven't added the manga tanks), but wanted to present it to get opinions about if this is worth to give a shot, because if I continue, it will definitely be a lot of work to do(And probably will separate in multiple PR, with this only providing the new Minion). If this sounds too complex, or doesn't fit with the platform, then it would be better to know now so I don't continue working on it.

PS: This refers to uncompressed folder, but has the potential to be extended with compressed cbz/zip idempotent chapter files in the future... But that is outside this original concept.

### DIFF
--- a/lib/LANraragi.pm
+++ b/lib/LANraragi.pm
@@ -13,7 +13,7 @@ use Config;
 use URI::Escape;
 use Time::HiRes qw(gettimeofday);
 
-use LANraragi::Utils::Generic    qw(start_shinobu start_minion get_version);
+use LANraragi::Utils::Generic    qw(start_tsubasa start_shinobu start_minion get_version);
 use LANraragi::Utils::Logging    qw(get_logger get_logdir);
 use LANraragi::Utils::Plugins    qw(get_plugins);
 use LANraragi::Utils::TempFolder qw(get_temp);
@@ -194,6 +194,12 @@ sub startup {
         shutdown_from_pid( get_temp . "/shinobu.pid" );
     }
     start_shinobu($self);
+
+    # Start File Watcher
+    if ( IS_UNIX ) {
+        shutdown_from_pid( get_temp . "/tsubasa.pid" );
+    }
+    start_tsubasa($self);
 
     # Check if this is a first-time installation.
     first_install_actions();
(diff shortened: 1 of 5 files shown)


CirnoT:
That is a lot of complexity that is already covered by other tools/readers available that cater specifically to manga though. Might be worth creating issue first to see if there's any interest in this at all.

Difegue:
Yeah I'm sorry but this does seem like too much work to maintain in the long term, you're basically making an entirely separate storage/filewatching engine for sparse folders at this point.  

For ongoing series the best thing I can recommend at this point is the .zip workaround that takes advantage of the flaw in the server's ID calculation. I might officialize that if we ever strengthen the IDs by moving to XXHash or something.

PR #1663, opened 2026-07-07
Fix openapi.yaml lint errors and warnings using redocly

gustaavv:


tools/openapi.yaml
```diff
@@ -3,6 +3,9 @@ openapi: 3.1.0
 info:
   version: 0.9.6
```
Difegue:
```suggestion
  version: 0.9.8
```

not related to the PR but might as well bump it

Difegue (approved):

PR #1667, opened 2026-07-07
Add support for .cbw files

### LINKED ISSUE #1228: support .cbw files
.cbw is comic book web files, they don't contain any of the images themselves but is rather just an metadata file with the image urls saved in the text file.
This makes them tiny and is perfect for mobile use and other things.

Most urls out there is static, aslong as they are online they should be loadable.

On the Lanraragi side, you would load in the image url directly instead of the image file in the WebUI.

These files can easily be converted later down the line by the user if the files are still online obviously!

This gives the user the ability to have an entire website "stored" in just a gigabyte or two at most while having all the metadata locally and giving the user all the metadata control a client like Lanraragi or ComicRack might support!

In the end it gives the user more control over online files, being able to process metadata from the sites in bulk.
Most of the sites i have tested this one has static image urls with CDN's which doesnt have much say in the end. 
Don't see it being a huge issue in the future.

https://wiki.mobileread.com/wiki/CBW
Here is a single example file:
[Tachibana-san's Circumstances With a Man.zip](https://github.com/user-attachments/files/19632838/Tachibana-san.s.Circumstances.With.a.Man.zip)

Nothing stops Lanraragi and its download system to save comics in a .cbw format either as an online alternative!

Shizuku-in:
## Summary

Add support for `.cbw` (ComicBookWeb) files.

Closes #1228

## How it works

CBW follows the same pattern as PDF: a non-image source file where `get_filelist` returns synthetic page names and `extract_single_file` produces image bytes. For CBW, "extraction" means an HTTP fetch from the remote CDN, cached in PageCache (LRU). This avoids CORS/mixed-content/hotlink issues that would occur with client-side loading.

- Parses `<Variables>` substitution and `[format:a-b]` range expansion per the CBW spec
- When a page is served, the next 3 pages are prefetched asynchronously for instant cache hits on forward navigation
- Download is denied with a user-facing toast
- `<Info/>` metadata is not yet parsed, which can be a follow-up

## Test

`prove -l tests/` and all tests successful.

Also manually verified:  upload -> thumbnail generation -> reader page serving -> thumbnail serving -> download denial toast

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Open source server for archival of comics/manga, running on Mojolicious + Redis.
 
 ## Features  
 
-* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/pdf supported, barebones support for epub)  
+* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/pdf supported, barebones support for epub and cbw)  
 
 * Read archives directly from your web browser: the server reads from within compressed files using temporary folders.
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Open source server for archival of comics/manga, running on Mojolicious + Redis.
 
 ## Features  
 
-* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/pdf supported, barebones support for epub and cbw)  
+* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/cbw/pdf supported, barebones support for epub)  
 
 * Read archives directly from your web browser: the server reads from within compressed files using temporary folders.
 
--- a/lib/LANraragi/Controller/Api/Archive.pm
+++ b/lib/LANraragi/Controller/Api/Archive.pm
@@ -16,6 +16,7 @@ use LANraragi::Utils::Database qw(get_archive_json set_isnew);
 use LANraragi::Utils::Logging  qw(get_logger);
 use LANraragi::Utils::Redis    qw(redis_encode);
 use LANraragi::Utils::Path     qw(compat_path get_archive_path move_path);
+use LANraragi::Utils::Archive  qw(is_cbw);
 
 use LANraragi::Utils::Login qw(is_logged_in_api);
 
@@ -104,6 +105,18 @@ sub serve_file {
 
     my $file = get_archive_path( $redis, $id );
     $redis->quit();
+
+    if ( is_cbw($file) ) {
+        return $self->render(
+            openapi => {
+                operation => "serve_file",
+                success   => 0,
+                error     => "CBW archives cannot be downloaded. They are XML metadata files that reference remote page images by URL; the actual comic pages are fetched on demand while reading."
+            },
+            status => 415
+        );
+    }
+
     $self->render_file( filepath => compat_path($file), filename => basename($file) );
 }
 
--- a/lib/LANraragi/Controller/Api/Archive.pm
+++ b/lib/LANraragi/Controller/Api/Archive.pm
@@ -16,7 +16,6 @@ use LANraragi::Utils::Database qw(get_archive_json set_isnew);
 use LANraragi::Utils::Logging  qw(get_logger);
 use LANraragi::Utils::Redis    qw(redis_encode);
 use LANraragi::Utils::Path     qw(compat_path get_archive_path move_path);
-use LANraragi::Utils::Archive  qw(is_cbw);
 
 use LANraragi::Utils::Login qw(is_logged_in_api);
 
@@ -105,18 +104,6 @@ sub serve_file {
 
     my $file = get_archive_path( $redis, $id );
     $redis->quit();
-
-    if ( is_cbw($file) ) {
-        return $self->render(
-            openapi => {
-                operation => "serve_file",
-                success   => 0,
-                error     => "CBW archives cannot be downloaded. They are XML metadata files that reference remote page images by URL; the actual comic pages are fetched on demand while reading."
-            },
-            status => 415
-        );
-    }
-
     $self->render_file( filepath => compat_path($file), filename => basename($file) );
 }
 
--- a/lib/LANraragi/Utils/Archive.pm
+++ b/lib/LANraragi/Utils/Archive.pm
@@ -26,13 +26,12 @@ use File::Temp qw(tempdir);
 use POSIX qw(strerror);
 use Mojo::DOM;
 use Mojo::UserAgent;
-use Mojo::IOLoop;
 
 use LANraragi::Utils::TempFolder qw(get_temp);
 use LANraragi::Utils::Logging    qw(get_logger);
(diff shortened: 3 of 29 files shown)


lib/LANraragi/Controller/Api/Archive.pm
```diff
@@ -104,6 +105,18 @@ sub serve_file {
 
     my $file = get_archive_path( $redis, $id );
     $redis->quit();
+
+    if ( is_cbw($file) ) {
+        return $self->render(
```
Difegue: don't think that's needed, users can download the cbw file if they want to

public/js/mod/index_contextmenu.js
```diff
@@ -51,7 +51,16 @@ function handleContextMenu(option, id) {
             LRR.openInNewTab(new LRR.ApiURL(`/reader?id=${id}`));
             break;
         case "download":
-            LRR.openInNewTab(new LRR.ApiURL(`/api/archives/${id}/download`));
+            if (extension === "cbw") {
```
Difegue: Ditto, don't think that's necessary

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
-  qw(is_file_in_archive extract_file_from_archive extract_single_file extract_thumbnail generate_thumbnail get_filelist);
+  qw(is_file_in_archive extract_file_from_archive extract_single_file extract_thumbnail generate_thumbnail get_filelist is_cbw parse_cbw_urls cbw_prefetch);
 
 sub is_pdf {
     my ( $filename, $dirs, $suffix ) = fileparse( $_[0], qr/\.[^.]*/ );
     return ( $suffix eq ".pdf" );
 }
 
+# CBW (ComicBookWeb) files are XML metadata files that reference remote page images by URL.
+# They contain no image data themselves; LR proxies and caches those images on demand.
```
Difegue:
```suggestion
# They contain no image data themselves; LRR proxies and caches those images on demand.
```

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
+    }
+
+    return ($url);
+}
+
+# parse_cbw_urls($file)
+# Reads a CBW file from disk and returns its ordered list of page image URLs.
+sub parse_cbw_urls ($file) {
+
+    # Memoize parsed URL lists by file path so that extract_single_file
```
Difegue:
```suggestion
    # Memorize parsed URL lists by file path so that extract_single_file
```

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
+    my @urls  = parse_cbw_urls($archive);
+
+    # Find the index of the currently requested page
+    my $current_idx;
+    for my $i ( 0 .. $#pages ) {
+        if ( $pages[$i] eq $current_path ) { $current_idx = $i; last; }
+    }
+    return unless defined $current_idx;
+
+    state $pf_ua = Mojo::UserAgent->new;
```
Difegue: Feels like you could just call `fetch_cbw_image` in a loop here and set the results into cache instead of duplicating code.

The tradeoff is that you'll instantiate a new UA for each page, but those are relatively cheap anyway.

Difegue (changes requested):
Pretty good! I mostly have simplification requests and some comment nits.

I thought initially we'd implement this by entirely offloading the remote page fetching to the web reader, but doing the proxying makes this transparent to thirdparty clients and does allow for resizing/index thumbnail to work out of the box.

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
+
+    # Memoize parsed URL lists by file path so that extract_single_file
+    # (called once per page) doesn't re-read/re-parse the XML every time.
+    state %url_cache;
+
+    return @{ $url_cache{$file} //= do {
+        open( my $fh, '<:raw', $file ) or die "Could not open CBW file '$file': $!\n";
+        my $xml = do { local $/; <$fh> };
+        close($fh);
+        [ parse_cbw_xml($xml) ];
```
Guerra24, line 137: This most likely needs to be changed to not break on Windows. I will be testing it locally and adding the diff here as a comment.

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
 use POSIX qw(strerror);
+use Mojo::DOM;
+use Mojo::UserAgent;
+use Mojo::IOLoop;
 
 use LANraragi::Utils::TempFolder qw(get_temp);
 use LANraragi::Utils::Logging    qw(get_logger);
 use LANraragi::Utils::Generic    qw(is_image shasum_str);
 use LANraragi::Utils::Redis      qw(redis_decode redis_encode);
 use LANraragi::Utils::Path       qw(get_archive_path);
```
Guerra24:
```suggestion
use LANraragi::Utils::Path       qw(get_archive_path open_path_or_die);
```

lib/LANraragi/Utils/Archive.pm
```diff
@@ -24,25 +24,219 @@ use Archive::Libarchive::Extract;
...
+# parse_cbw_urls($file)
+# Reads a CBW file from disk and returns its ordered list of page image URLs.
+sub parse_cbw_urls ($file) {
+
+    # Memoize parsed URL lists by file path so that extract_single_file
+    # (called once per page) doesn't re-read/re-parse the XML every time.
+    state %url_cache;
+
+    return @{ $url_cache{$file} //= do {
+        open( my $fh, '<:raw', $file ) or die "Could not open CBW file '$file': $!\n";
```
Guerra24:
```suggestion
        open_path_or_die( my $fh, '<:raw', $file );
```

Guerra24 (changes requested):
These changes should be all that's needed.

Shizuku-in:
Updates:

- c76310a - remove cbw download denial; fix some comment typos
- 2515490 - simplify `cbw_prefetch` to reuse `fetch_cbw_image`; switch `open` to `open_path_or_die`

Guerra24 (approved):

README.md
```diff
@@ -39,7 +39,7 @@ Open source server for archival of comics/manga, running on Mojolicious + Redis.
 
 ## Features  
 
-* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/pdf supported, barebones support for epub)  
+* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/pdf supported, barebones support for epub and cbw)
```
Difegue:
```suggestion
* Stores your comics in archive format. (zip/rar/targz/lzma/7z/xz/cbz/cbr/cbw/pdf supported, barebones support for epub)  
```

Difegue (approved):

PR #1669, opened 2026-07-08
carousel wait when search is not initialized

psilabs-dev:
carousel errors out through toast when 204 (still initializing) is received. Should wait until backend is ready


psilabs-dev:
`LRR_TANKGROUPED` doesn't get computed in an empty instance, so 204 happens regardless.

PR #1672, opened 2026-07-10
Add .lrrignore support for filtering archives

sitiyou:
This PR add `.lrrignore` support with plexignore based syntax, triggered during Shinobu scanning and when new files are added.

See https://github.com/Difegue/LANraragi/discussions/1120

### syntax

Based on the [Plex `.plexignore`](https://support.plex.tv/articles/201381883-special-keyword-file-folder-exclusion/) spec, extended with some useful features from gitignore:

| Syntax | Description |
|--------|-------------|
| `*` | glob wildcard, matches any character except `/` |
| `#` | comment (line start only; `\#` escapes to literal `#`) |
| `!` | negation, overrides parent directory rules (`\!` escapes to literal `!`) |
| Leading `/` | anchors the pattern to the `.lrrignore` file's directory |
| Trailing `/` | matches all contents of the directory |
| `\*` | literal `*` |
| `\` | for any other `\X`, the backslash is silently dropped (`\X` → `X`) |

Rules are resolved relative to the `.lrrignore` file's own directory, with nested rules overriding parent ones.

### interface

`Ignore.pm` exports four methods:

- `is_ignored($file, $rules)`: walks parent directories via `dirname` and matches against compiled regex rules
- `build_ignore_rules($content_dir)`: scans the directory tree for `.lrrignore` files and builds a rule set
- `initialize()`: called at server startup, builds rules and stores them in Redis as `LRR_IGNORE_RULES`
- `load_ignore_rules()`: loads the frozen rule set from Redis with `state $cached`

### windows compatibility

The project's existing cross-platform filesystem APIs don't quite cover some of the path operations this feature needs. I'm not familiar with Windows, so I can't say for sure it'll work there. The tests pass under MSYS2. And that is all I can guaranteen.

### unit test

Added tests covering the various ignore syntax rules.

### DIFF
--- a/lib/LANraragi/Utils/Ignore.pm
+++ b/lib/LANraragi/Utils/Ignore.pm
@@ -1,0 +1,191 @@
+package LANraragi::Utils::Ignore;
+
+use strict;
+use warnings;
+use utf8;
+use feature qw(signatures state);
+no warnings 'experimental::signatures';
+
+use File::Spec;
+use File::Basename qw(dirname basename);
+use Config;
+
+use LANraragi::Utils::Path    qw(open_path find_path create_path);
+use LANraragi::Utils::String  qw(trim);
+use Storable                  qw(nfreeze thaw);
+
+use Exporter 'import';
+our @EXPORT_OK = qw(
+    is_ignored
+    build_ignore_rules
+    initialize
+    load_ignore_rules
+);
+
+use constant IS_UNIX => ( $Config{osname} ne 'MSWin32' );
+
+# Called at server startup to scan .lrrignore files and store rules in Redis.
+sub initialize {
+    my $redis = LANraragi::Model::Config->get_redis_config;
+    my $rules = build_ignore_rules( LANraragi::Model::Config->get_userdir );
+    $redis->set( "LRR_IGNORE_RULES", nfreeze($rules) );
+    $redis->quit();
+}
+
+# Load cached ignore rules from Redis.
+# Rules are immutable at runtime; they only change when the server restarts.
+sub load_ignore_rules {
+    state $cached = do {
+        my $redis = LANraragi::Model::Config->get_redis_config;
+        my $data  = $redis->get("LRR_IGNORE_RULES");
+        $redis->quit();
+        return undef unless $data;
+
+        # compile regexp
+        my $rules = thaw($data);
+        for my $patterns (values %{ $rules->{entries} }) {
+            for my $p (@$patterns) {
+                $p->{regex} = qr{$p->{regex}};
+            }
+        }
+        return $rules;
+    };
+    return $cached;
+}
+
+# Scan the content directory for .lrrignore files and build rule entries.
+sub build_ignore_rules ($content_dir) {
+    $content_dir =~ s{\\}{/}g unless IS_UNIX;
+    my $rules = {
+        content_dir => $content_dir,
+        entries     => {},
+    };
+
+    my @lrrignore_dirs;
+    find_path(
+        sub {
+            my $f = create_path($_);
+            return if -d $f;
+            return unless basename($f) eq '.lrrignore';
+            push @lrrignore_dirs, dirname($f);
+        },
+        $content_dir
+    );
+
+    for my $dir (@lrrignore_dirs) {
+        load_ignore_file( $dir, $rules->{entries} );
+    }
+
+    return $rules;
+}
+
+# Rules data structure:
+# $rules = {
+#     content_dir => $path,                      # Root content directory for scanning
+#     entries     => {                           # Hash map: dir -> patterns
+#         $path => [                             # Parsed patterns from a single .lrrignore file
+#             {
+#                 pattern  => "*.tmp",           # Original pattern string
+#                 regex    => qr/.../i,          # Compiled regex for matching
+#                 negation => 0|1,               # 1 = ! re-include, 0 = ignore
+#             },
+#             ...
+#         ],
+#         ...
+#     },
+# }
+#
+sub is_ignored ($file, $rules) {
+    return 0 unless $rules;
+
+    my $entries = $rules->{entries};
+    return 0 unless $entries && keys %$entries;
+
+    return 0 unless is_subpath($file, $rules->{content_dir});
+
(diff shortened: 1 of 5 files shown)


sitiyou:
I forgot to handle escaped special characters in ignore rules.
Maybe I need to rethink the .lrrignore syntax design, since implementing full gitignore compatibility syntax is turning out to be more complex than I expected.
It also means characters like `[]` would require escaping to work properly, but they are extremely common in manga titles.

Maybe I should only implement a subset of gitignore instead.

sitiyou:
@Difegue this PR is ready for review. The syntax has been simplified and I've tested some common use cases in docker.

lib/Shinobu.pm
```diff
@@ -133,11 +135,20 @@ sub update_filemap {
...
 
+    my $ignore_rules = build_ignore_rules($dirname);
+
     # Get all files in content directory and subdirectories.
     find_path(
         sub {
             $_ = create_path($_);
             return if -d $_;    #Directories are excluded on the spot
+
+            update_ignore_rules( dirname($_), $ignore_rules );
```
Difegue: Re-updating the rules for every single incoming file feels overkill.  

I think we could get by only updating the rules when Shinobu gets started.

lib/Shinobu.pm
```diff
@@ -329,6 +341,13 @@ sub new_file_callback ($name) {
     $logger->debug("New file detected: $name");
     unless ( -d $name ) {
 
+        my $dirname = LANraragi::Model::Config->get_userdir;
+        my $ignore_rules = build_ignore_rules( $dirname, $name );
```
Difegue: ditto

lib/Shinobu.pm
```diff
@@ -357,6 +376,15 @@ sub deleted_file_callback ($name) {
     }
 }
 
+sub cleanup_ignored_archive ($file) {
```
Difegue: I would not do this, it makes it all too easy for people playing with ignore rules to accidentally hose metadata.  

The only cleanup we should do is wipe the "file" field of the redis key so that they can be removed by a clean database run.

Difegue (changes requested):

lib/Shinobu.pm
sitiyou: Loading all `.lrrignore` files at startup would mean walking the filesystem twice: once to gather ignore rules, once to update filemaps.

Since `find_path` visits files in the same directory sequentially, `update_ignore_rules` can be simplified to a single string comparison for most calls:
```perl
return if $rules->{current_dir} && $rules->{current_dir} eq $dst_dir;
```

---

BTW, `File::Find` supports pruning to skip ignored directories, but `Win32::LongPath::Find` lacks this functionality so I did not implement pruning.

And `Win32::LongPath::Find` does not return directories, meaning we can't only update rules when entering a directory.

lib/Shinobu.pm
sitiyou: This can be optimized by caching all ignore rules after the filemap scan. So any modifications made by the user will not take effect until Shinobu is restarted.
And newly added files can simply use the cached rules, no extra filesystem api calls are needed.

lib/Shinobu.pm
```diff
@@ -154,14 +166,22 @@ sub update_filemap {
     my @deletedfiles = grep { !$fshash{$_} } @filemapfiles;
```
psilabs-dev: All the ignored files are treated as deleted? What happens to metadata during a cleanup?

lib/Shinobu.pm
```diff
@@ -154,14 +166,22 @@ sub update_filemap {
...
 
-    # Delete old files from filemap
+    # Delete old files from filemap and cleanup ignored archives
+    my $redis_arc = LANraragi::Model::Config->get_redis;
     foreach my $deletedfile (@deletedfiles) {
         $logger->debug("Removing $deletedfile from filemap.");
+        if ( -e $deletedfile ) {
+            my $id = $redis->hget( "LRR_FILEMAP", $deletedfile );
+            if ($id) {
+                $redis_arc->hset( $id, "file", "" );
```
psilabs-dev: related (since "file" is false now): https://github.com/Difegue/LANraragi/blob/b94e4805677d4ca75e7d61ca13c4e3e99c4b99c8/lib/LANraragi/Utils/Database.pm#L391

lib/LANraragi/Utils/Ignore.pm
```diff
@@ -0,0 +1,190 @@
+package LANraragi::Utils::Ignore;
```
psilabs-dev, line 1: Missing upload-time filtering

lib/Shinobu.pm
psilabs-dev: Needs a lock

psilabs-dev (changes requested):
Couple questions/issues from me at first look

I think ignore reads are cheap compared to ID compute, shinobu can probably do a read once before every scan. Upload API on the other hand is less forgiving (maybe cache the rules in redis? but ehh)

Otherwise, users should not expect to update rules during LRR uptime, bc the results will be inconsistent per worker (unless you're reading per call).

sitiyou:
OK, let me rethink the design:

- The ignore rules should be global and immutable, to make them consistent across all workers (upload, shinobu, and any other plugins).
- An extra scan of the filesystem to construct ignore rules is actually acceptable, considering the filesystem cache. (30ms vs 10ms, in my test with 700 files and 15 directories on HDD using find . > /dev/null)

So I should construct ignore rules only at startup, and cache them internally in ignore.pm. (but is it ok to make ignore.pm stateful? or simply put them in redis?)

sitiyou:
UPDATE:
- Ignore rules are now built at startup and stored in redis for global access.
- upload.pm now rejects incoming files if they match any ignore rules.

lib/Shinobu.pm
sitiyou: It seems ok in my test. `delete_archive` will not delete the actual file since the `file` field was wiped, only metadata will be deleted.

```perl
# lib/LANraragi/Model/Archive.pm +426
    if ( -e $filename ) {
        my $status = unlink_path($filename);
```

PR #1673, opened 2026-07-10
Add process manager

Guerra24:
Adds a tiny process manager (Hitagi) to handle all the processes that run as part of the server (Mojo, Shinobu, Tsubasa) and exposes a cli-like interface over a Unix socket to allow control of the processes' lifetime. Also takes over and abstracts away all process management that originally existed in the server.

Moves the Minion worker to its own dedicated process like on Windows. Worker.pm -> Tsubasa.pm.

Adds a tiny client on the LRR side that exposes the socket as regular methods.

By default if Hitagi is not running LRR will run with the old process management as if nothing had changed. If it is available then all the process-related code is disabled and instead handled through it.

This is detected by checking if the env variable `HITAGI_SOCK` has a value, if it has one then it assumes the parent process is Hitagi and runs in that mode.

Initial client api:
```
LANraragi::Utils::Hitagi::restart_all(); # Restart all processes
LANraragi::Utils::Hitagi::restart( <process> ); # (Re)starts a process (lanraragi, shinobu, tsubasa)
LANraragi::Utils::Hitagi::stop( <process> ); # Stops a process
LANraragi::Utils::Hitagi::pid( <process> ); # Gets the pid of a process
```

For this initial PR it is only enabled on Docker.

Aside from that it cleans up the remaining S6 jank caused by the stop script.
Provides the restart capabilities (without restarting the whole container) that #1511 needs.
Unblocks slim containers (no valkey and no s6).
Unblocks multi-workers on Windows.


Guerra24:
Ready for review... finally.

PR #1681, opened 2026-07-16
Add es-module-shims for legacy importmap support

onyxlake:
Add [es-module-shims](https://github.com/guybedford/es-module-shims) to provide an import map fallback for legacy browsers on end-of-life devices (e.g., iPad Air 2).


Difegue (approved):
Happy to get this in but I can't guarantee how long we'll keep support for older browsers working.  

This is a light enough lift, but if we end up having to add three thousand polyfills it'll probably be too hard to keep the compat going..

PR #1692, opened 2026-08-07
Fix thumbnail and tags not showing on hover in dupe page

Guerra24:


Difegue (approved):
Starting to wonder if this module stuff might not be hard to maintain in the long run

PR #1702, opened 2026-08-22
Wrap icon circle around DT page number

psilabs-dev:
When the DT page is long, it stretches the background circle instead of applying a bevel.

Before:

Screenshot 2026-08-22 at 1 39 48 PM

After:

Screenshot 2026-08-23 at 11 23 40 AM


Difegue (approved):
this has been an issue for years and i've always been 2lazy to fix it 🫡 ty

PR #1706, opened 2026-08-25
Replace Authen::Passphrase with Crypt::Passphrase

### LINKED ISSUE #1683: Unicode in password field causes server error
**LRR Version and OS**
LRR 0.9.81 - Atomica in Docker on Debian.

**Bug Details**
Unicode input in password field in /login causes server error: 
"input must contain only octets at /home/koyomi/perl5/lib/perl5/x86_64-linux-thread-multi/Crypt/Eksblowfish/Bcrypt.pm line 85."

String to copypaste for test: "♀"

It is not a big problem, I don't use unicode in passwords and I would be fine with [a-zA-Z0-9_] validator or similar, I just think server shouldn't explode like this on a slightly weird input.

**Matching Logs**

There is no log of this. By contrast, if you try to log in with incorrect ASCII password, it usually gives error log like this: 
"[2026-07-22 07:32:19] [LANraragi] [warn] Failed login attempt with password 'test' from 192.168.0.110"

**Screenshots**

Image

Guerra24:
Remove Crypt::DES patch and fix #1683.

The `{CRYPT}` thing is only for back-compat, it should probably be removed at some point.

The subtype is hardcoded to `2a` for back-compat but it should be changed or Bcrypt replaced entirely with something else. I have another branch that migrates directly to Argon2 so that is another option.


Difegue (approved):
Nice to finally get rid of the Crypt::DES patches for sure :shipit: Thanks for this/

PR #1719, opened 2026-09-08
Support Up/Down arrow keys for page navigation in the reader

### LINKED ISSUE #1718: Feature request: Support Up/Down arrow keys for page navigation in the reader (presenter clicker support)
Hi! I'd like to suggest adding support for the `Up` / `Down` arrow keys as page-navigation shortcuts in the reader, alongside the existing `Left` / `Right` arrow keys.

### Use case
I read LANraragi from the couch with a wireless presenter clicker — the small handheld remote pictured below:

wireless presenter clicker with up/down keys

These remotes only carry a few buttons: `Up` / `Down` arrows plus a hyperlink / laser button. They send no `Left` / `Right` keys. Because the reader currently only listens to `Left` / `Right` for page turning, the clicker's buttons are ignored, and I have to keep one hand on the mouse — defeating the whole point of going hands-free with a remote.

### Current behavior
- Reader accepts `Left` / `Right` (and a handful of other shortcuts introduced by #455 and #1214) for page navigation.
- `Up` / `Down` do nothing in the reader.

### Desired behavior
Treat `Up` as `Left` (previous page) and `Down` as `Right` (next page) — basically a one-line alias to the existing page-turn handler.

### Why
It's a tiny change (an extra keybinding) that unlocks full hands-free reading with an off-the-shelf wireless presenter clicker, which is much more ergonomic for couch / tabletop reading than juggling a mouse or a full keyboard.

Happy to test or submit a PR if that's helpful.

zangwill:
Closes #1718

I read LANraragi with a wireless presenter clicker — a small handheld remote with only `Up` / `Down` arrows plus a laser / hyperlink button. These remotes send no `Left` / `Right` keys, so the reader currently ignores the clicker's buttons, forcing me to use a mouse just to flip pages — which defeats the purpose of holding a remote in the first place.

This PR aliases `Up` ≡ `Left` and `Down` ≡ `Right` in `handleShortcuts`, including the `Shift` modifier for first / last page jumps.

### DIFF
--- a/public/js/mod/reader_common.js
+++ b/public/js/mod/reader_common.js
@@ -565,6 +565,7 @@ function handleShortcuts(e) {
             spaceScrollProcessInput(e);
             break;
         case 37: // left arrow
+        case 38: // up arrow
             if (e.shiftKey) {
                 changePage("first", true);
             } else {
@@ -572,6 +573,7 @@ function handleShortcuts(e) {
             }
             break;
         case 39: // right arrow
+        case 40: // down arrow
             if (e.shiftKey) {
                 changePage("last", true);
             } else {


Difegue (approved):
I dont mind adding this in spirit, but other contributors have mentioned this might conflict with infinite scrolling mode; Have you tested the behavior with that? 

Leaving as tentatively approved until i have time to verify regardless 😅

